PR #112, opened 2020-06-30
Add more tests

skorfmann:
This needs some work around the ink testing library itself, see here https://github.com/vadimdemedes/ink-testing-library/issues/3

The issue is around `useEffect`. While we could work with timeouts, it would be way nicer to have a proper solution.


skorfmann:
Let's come back to this when approaching #166

PR #199, opened 2020-07-17
Readme: fix link to cdktf/

rstacruz:
Just a quick typo fix :)


skorfmann:
Thanks! 

I believe the failing build is due to secrets not being available for external pull requests. I'll check which options we have there.

PR #215, opened 2020-07-20
Fix interpolation bug in complex computed list of data sources

jooohn:
Interpolation of `ComplexComputedList` for `TerraformDataSource`s should starts with `data.` prefix.

### DIFF
--- a/packages/cdktf/lib/terraform-resource.ts
+++ b/packages/cdktf/lib/terraform-resource.ts
@@ -13,6 +13,8 @@ export interface ITerraformResource {
   count?: number;
   provider?: TerraformProvider;
   lifecycle?: TerraformResourceLifecycle;
+
+  interpolationForAttribute(terraformAttribute: string): string;
 }
 
 export interface TerraformResourceLifecycle {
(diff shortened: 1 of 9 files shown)


packages/cdktf/lib/terraform-resource.ts
```diff
@@ -13,6 +13,8 @@ export interface ITerraformResource {
   count?: number;
   provider?: TerraformProvider;
   lifecycle?: TerraformResourceLifecycle;
+
+  interpolationForAttribute(terraformAttribute: string): string;
```
jooohn, line 17: This line is going to change the visibility of existing `interpolationForAttribute`. I prefer this style to exposing `interpolationPrefix` attribute for example, but let me know if there is any coding policy about this.

skorfmann:
Thanks for your contribution, very appreciated. It looks good at a first glance. Could you provide a pointer to the provider schema and the resource where you encountered this issue?

jooohn:
@skorfmann

I've encountered this issue when tried this data source attribute.

https://www.terraform.io/docs/providers/aws/d/organizations_organization.html#roots

anubhavmishra:
@skorfmann should be good to merge?

skorfmann:
@jooohn thanks again for your contribution. You can use it via `cdktf@next` until there's a proper release.

jooohn:
thank you for approving this!

Issue #33, opened 2020-04-27
Support for Synthesizing Terraform Remote State Backends


skorfmann:
I think that remote state functionality should be supported as well. That would enable the usage in more complex environments (e.g. thinking of centrally managed VPCs)

jsteinich:
I worked on this a bit for [Terrastack](https://github.com/jsteinich/terrastack/tree/backends). Also included typed backend support

jsteinich:
I started some early work on this: https://github.com/hashicorp/terraform-cdk/pull/241

skorfmann:
This will be fully available with https://github.com/hashicorp/terraform-cdk/pull/241

PR #224, opened 2020-07-21
Typed backend config

jsteinich:
This adds the ability to specify [Terraform Settings](https://www.terraform.io/docs/configuration/terraform.html) when creating a `TerraformStack`, including strongly typed backends.

I still need to add some test cases, but I believe it is otherwise complete. I think remote state belongs in a separate request.


jsteinich:
I'm not a big fan of parsing the Go schemas directly, but if the Terraform team is willing to support getting them as JSON in some fashion, then that makes a lot of sense.

Making backends `Constructs` has some nice properties, but there is one point that I wonder about:
I like enforcing with the API that only 1 backend can be set. We could enforce in the synth or just let Terraform handle it though.

skorfmann:
> I'm not a big fan of parsing the Go schemas directly, but if the Terraform team is willing to support getting them as JSON in some fashion, then that makes a lot of sense.

👍 

> Making backends `Constructs` has some nice properties, but there is one point that I wonder about:
> I like enforcing with the API that only 1 backend can be set. We could enforce in the synth or just let Terraform handle it though.

Yes, this could be validated after synth. For now, I think it'd be sufficient to let Terraform do the validation.

anubhavmishra:
> ### Mid / Long term
> I think the backends should follow the same concept as the `TerraformProvider` and would be ideally generated from a schema.

I think long term it makes sense to automatically generate the backends but for now, there is no way to do it.

> Since there are [Go schemas](https://github.com/hashicorp/terraform/tree/ba081aa10a8b8a4a32648ad9f67670a1425dec0b/backend/remote-state) for the backends (e.g. [S3](https://github.com/hashicorp/terraform/blob/ba081aa10a8b8a4a32648ad9f67670a1425dec0b/backend/remote-state/s3/backend.go#L23-L264), I think that the backend classes should be generated from there somehow. This could be similar to what we do for provider Constructs and be a build step in `cdktf`.
> 
> The benefits I see for doing this:
> 
> * No manual work to keep it aligned with Terraform
> * No need to test individual backends, as long as the code generator is tested properly
> * Inline docs would be generated as well
> * Long term maintenance effort significantly less compared to doing this manually
> 
> Perhaps the Terraform team could support this by exposing the Go schemas as JSON schema somehow? /cc @anubhavmishra

We can have a conversation with the Terraform team to see if that is possible.

> ### Short term
> I think it'd be good to rework this PR to match the anticipated generated structre, which means:
> 
> * Have a base `TerraformBackend` class
> * Treat backends as a Construct which get synthesised simlar to `Resources`, `Data` and `Providers`.
> * Drop support for generic `TerraformSettings`
> 
> This would mean, we'd be good to go with the manual implementation and could replace this later on by something which is generated from a JSON schema with little to none breaking changes

Backends don't tend to change as much as we should be good with the manual implementation for now.

> What are your thoughts @jsteinich / @anubhavmishra

anubhavmishra:
I edited the issue that was related to this PR. https://github.com/hashicorp/terraform-cdk/issues/33

skorfmann:
@jsteinich would you be up for refactoring this accordingly?

jsteinich:
I should be able to get the changes done in the next couple of days.
I think remote state makes sense as a separate PR after this one.

skorfmann:
> I should be able to get the changes done in the next couple of days.
> I think remote state makes sense as a separate PR after this one.

Awesome, thanks! Looking forward to it.

jsteinich:
I believe this is ready to go, but there are a few points that might require more discussion.

1. I kept the code pretty minimal. I could see making some changes when/if switching to generated code, but I think having less code is better when manually maintained.
2. I didn't transfer comments from the documentation. Seems like they would be useful longer term, but I'm not sure they add a ton right now.
3. I think examples are going to need some more organization. I put the new examples in a folder structure that made some sense to me, but there are a few different ways this could be done. Can probably wait a bit on it though.

skorfmann:
Cool, looks good at a first glance. Will look into it more thoroughly later today.

packages/cdktf/lib/backends/backend_cos.ts
```diff
@@ -0,0 +1,24 @@
+import { Construct } from "constructs";
+import { TerraformBackend } from '../terraform-backend';
+import { keysToSnakeCase } from "../util";
+
+export class CosBackend extends TerraformBackend {
```
skorfmann, line 5: Is there a particular reason the file is named `backend_cos` and the class `CosBackend`?

packages/cdktf/lib/terraform-backend.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export abstract class TerraformBackend extends TerraformElement {
+    constructor(scope: Construct, id: string, protected readonly name: string) {
+        super(scope, id);
+    }
+
+    // jsii can't handle abstract classes?
+    protected synthesizeAttributes(): { [name: string]: any } {
+        return {}
+    }
```
skorfmann, line 12: I'm wondering if it'd make sense to support escape hatches here as well?

packages/cdktf/lib/backends/backend_cos.ts
jsteinich: I think it was from before I put them in their own folder. I'll rename.

packages/cdktf/lib/terraform-backend.ts
jsteinich: It's still possible to used the stack level escape hatch, but I can try adding support.

skorfmann:
Yeah, I think it would be good for consistency reasons

Issue #217, opened 2020-07-20
Error when trying to get openstack provider


Madko:
Hi,
Can't find any doc about openstack and cdk so I'm using the one for docker and adapting it.
Here is my cdktf.json:
```
{
  "language": "python",
  "app": "pipenv run ./main.py",
  "terraformProviders": ["openstack@~>1.29"],
  "codeMakerOutput": "imports"
}
```
and here is the error I get while running cdktf get:
```
$ cdktf get
⠇ downloading and generating providers...
⠼ downloading and generating providers...
Error: non-zero exit code 1
    at ChildProcess.<anonymous> (/usr/local/lib/node_modules/cdktf-cli/lib/util.js:89:27)
    at Object.onceWrapper (events.js:417:26)
    at ChildProcess.emit (events.js:310:20)
    at maybeClose (internal/child_process.js:1021:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:286:5)
non-zero exit code 1
```

Is that normal ? I only have one new file:
./.terraform/plugins/linux_amd64/terraform-provider-openstack_v1.29.0_x4

Best regards,
Edouard

skorfmann:
This is a bug and relates to #124 

With debugging output this looks like

```
➜  python-aws git:(master) ✗ ../../packages/cdktf-cli/bin/cdktf get
⠦ downloading and generating providers...
⠼ downloading and generating providers...
providers/openstack/data-openstack-networking-port-v2.ts:352:50 - error TS2554: Expected 2 arguments, but got 3.

352     return new StringMap(this, 'dns_assignment', index);
                                                     ~~~~~
providers/openstack/networking-port-v2.ts:391:50 - error TS2554: Expected 2 arguments, but got 3.

391     return new StringMap(this, 'dns_assignment', index);
                                                     ~~~~~
Error: non-zero exit code 1
    at ChildProcess.<anonymous> (/Users/sebastian/projects/cdk/terraform-cdk/packages/cdktf-cli/lib/util.js:89:27)
    at Object.onceWrapper (events.js:422:26)
    at ChildProcess.emit (events.js:315:20)
    at maybeClose (internal/child_process.js:1026:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:286:5)
non-zero exit code 1
```

anubhavmishra:
@Madko are you still seeing this bug? Otherwise, we can close this issue.

PR #229, opened 2020-07-22
Add Lists of Maps

### LINKED ISSUE #124: List of String Maps
The `vault` provider `.gen/providers/vault/transit-secret-backend-key.ts:239:40` (`nomad` provider also affected)

```json
      "keys": {
        "type": [
          "list",
          [
            "map",
            "string"
          ]
        ],
        "description": "List of key versions in the keyring.",
        "computed": true
      },
```

renders invalid to

```typescript
  // keys - computed: true, optional: false, required: true
  public keys(index: string) {
    return new StringMap(this, 'keys', index);
  }
```

skorfmann:
This extends the schema generator to support lists of maps. 

Fixes #124 #217


examples/typescript-vault/main.ts
```diff
@@ -0,0 +1,36 @@
...
+import { App, TerraformStack, TerraformOutput } from 'cdktf';
+import { VaultProvider, TransitSecretBackendKey, Mount } from './.gen/providers/vault'
+
+class MyStack extends TerraformStack {
+  constructor(scope: Construct, name: string) {
+    super(scope, name);
+
+    new VaultProvider(this, 'vault', {
+      address: 'http://0.0.0.0:8200',
+      token: "YOUR_VAULT_TOKEN"
```
anubhavmishra, line 11: Maybe we can use the default vault address and token environment variables. we can support `VAULT_ADDR` and `VAULT_TOKEN` the two environment variables.

skorfmann: Yes, done b59d659

anubhavmishra: Awesome, feel free to merge!

PR #230, opened 2020-07-22
Avoid conflicts with resource attributes called "node"

### LINKED ISSUE #125: Node is a reserved name in cdktf / constructs
In the `consul` provider: `.gen/providers/consul/service.ts:284:14`

```
  // node - computed: false, optional: false, required: true
  private _node: string;
  public get node() {
    return this._node;
  }
  public set node(value: string) {
    this._node = value;
  }
```

Is invalid, since we defined it in `TerraformElement`. 

We could either rename it in our implementation, or rewrite provider attributes to another name.

skorfmann:
The `consul` provider has an attribute called `node`, which conflicts with the function in `TerraformElement`.

This introduces a more specific name `constructNode` to avoid conflicts like this.

Fixes #125

### DIFF
--- a/packages/cdktf/lib/terraform-element.ts
+++ b/packages/cdktf/lib/terraform-element.ts
@@ -14,11 +14,11 @@ export class TerraformElement extends Construct {
   constructor(scope: Construct, id: string) {
     super(scope, id)
 
-    this.node.addMetadata('stacktrace', 'trace')
+    this.constructsNode.addMetadata('stacktrace', 'trace')
     this.stack = TerraformStack.of(this);
   }
 
-  public get node(): Node {
+  public get constructsNode(): Node {
     return Node.of(this)
   }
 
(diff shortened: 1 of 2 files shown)


packages/cdktf/lib/terraform-element.ts
```diff
@@ -14,11 +14,11 @@ export class TerraformElement extends Construct {
   constructor(scope: Construct, id: string) {
     super(scope, id)
 
-    this.node.addMetadata('stacktrace', 'trace')
+    this.constructNode.addMetadata('stacktrace', 'trace')
     this.stack = TerraformStack.of(this);
   }
 
-  public get node(): Node {
+  public get constructNode(): Node {
```
eladb, line 21: Mmm... Interesting... I am wondering if we should take this into account in constructs 10.x?

skorfmann: Potential naming conflicts should be a thought of, yes.

eladb:
@skorfmann We discussed this in our team and unfortunately we will have to reintroduce `construct.node` in `constructs` 10.0 in order not to be break many AWS CDK apps when we remove the compatibility layer we have in 1.x.

So we are "stuck" with `node` (forever).

I would recommend changing your the terraform code generator to simply mangle the name `node` in case it appears in a resource (maybe something like `consolNode` (prefix the type name).

PR #238, opened 2020-07-23
Add python docker example

abhishekamralkar:
A Docker example in Python.


examples/python-docker/README.org
```diff
@@ -0,0 +1,87 @@
+#+title: Python-Docker
```
anubhavmishra, line 1: I would use the [markdown](https://en.wikipedia.org/wiki/Markdown) format for the README.

anubhavmishra: Also, please follow the guide shown [here](https://github.com/hashicorp/terraform-cdk/tree/master/examples/python-aws#python-aws) and update the README with the instructions accordingly.

examples/python-docker/Pipfile
```diff
@@ -0,0 +1,11 @@
...
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[requires]
+python_version = "3.7"
+
+[packages]
+constructs="3.0.4"
+cdktf="0.0.12"
```
anubhavmishra, line 11: I would use the latest version of cdktf package.

examples/python-docker/README.org
abhishekamralkar: README updated as per the guide and formatting changed to markdown.

examples/python-docker/Pipfile
abhishekamralkar: 0.0.12 is the latest version for cdktf as per the Pypi.

skorfmann: @anubhavmishra  meant `cdktf@next`, however we [don't release](https://pypi.org/project/cdktf/#history) pre releases for Python at the moment. I'm not even sure, if there's something comparable to `cdktf@next` in pypi.

However, I think for the examples we should stick to the monorepo references [as here](https://github.com/hashicorp/terraform-cdk/blob/b8aeded4cb58a7f367195d5ed8a4632588f43eca/examples/python-aws/Pipfile)

```
[[source]]
name = "pypi"
url = "https://pypi.org/simple"
verify_ssl = true

[requires]
python_version = "3.7"

[packages]
cdktf = {path = "./../../dist/python/cdktf-0.0.0-py3-none-any.whl"}
```

abhishekamralkar: @skorfmann Pipfile updated to use the monorepo.

skorfmann: Cool, thanks 👍

skorfmann:
Good from my point of view, @anubhavmishra ?

PR #241, opened 2020-07-24
Remote state

jsteinich:
Meant to be merged after #224 

An approach to remote state where each backend has an explicit type.

Before making backends constructs, I had a version which just had one remote state type which took a backend instance as a parameter. If we'd rather go that route I could implement either by making a user enter the backend name or by creating a wrapper class around the backend props that had the name.

I went with the current approach since it has a better matching feel to backends in my opinion, but I'd be happy to implement differently.

I imagine cross stack references will be built using remote state. It seems feasible to generate these, but we can always revisit when implementing. In the short term, it feels valuable to get to feature parity with raw Terraform.


skorfmann:
Cool, I like the direction of this 👍 

> I imagine cross stack references will be built using remote state. 

Haven't really thought about it, but it seems likely.

>  It seems feasible to generate these, but we can always revisit when implementing.

Thought the same when looking at this, that would be pretty cool. But agree, that's nothing for now.

> In the short term, it feels valuable to get to feature parity with raw Terraform.

👍

jsteinich:
I think this is ready to go. Once #224 is merged in I'll rebase and it should be quite a bit easier to see the changes.

@skorfmann tests failed because a snapshot has some of my local paths in it. I can update the snapshot to match what it expects, but I'm wondering if there is a better way to handle that. Or maybe I've just messed something up in my dev environment.

skorfmann:
> I think this is ready to go. Once #224 is merged in I'll rebase and it should be quite a bit easier to see the changes.

Just merged that one.

skorfmann:
> @skorfmann tests failed because a snapshot has some of my local paths in it. I can update the snapshot to match what it expects, but I'm wondering if there is a better way to handle that. Or maybe I've just messed something up in my dev environment.

Hm, the `creation_stack` looks wrong. I think I saw that over the weekend in one of my examples as well. That's likely a bug related to Tokens.

packages/cdktf/lib/tokens/private/stack-trace.ts
```diff
@@ -13,5 +13,5 @@ export function captureStackTrace(below?: Function): string[] {
   if (!object.stack) {
     return [];
   }
-  return object.stack.split('\n').slice(1).map(s => s.replace(/^\s*at\s+/, ''));
+  return object.stack.split('\n').slice(1).map(s => s.replace(/^\s*at\s+/, '').replace(/\(.*packages/, '(packages').replace(/\(.*node_modules/, '(node_modules'));
```
skorfmann, line 16: What's this about?

jsteinich: It was an attempt to work around inconsistent traces, but really it is an incomplete hack that is at best unrelated and at worst obscures where errors occur. I'll take it out.

jsteinich:
@skorfmann Is there any way to re-run the build action?

skorfmann:
> @skorfmann Is there any way to re-run the build action?

Just triggered the build again. What I sometimes do is an empty commit as trigger `git commit --allow-empty -m 'trigger build'` :)

packages/cdktf/lib/tokens/private/stack-trace.ts
skorfmann: As mentioned here, I think that's a bug https://github.com/hashicorp/terraform-cdk/pull/241#issuecomment-664402253 - Let me check

docs/working-with-cdk-for-terraform/remote-state.md
```diff
@@ -0,0 +1,28 @@
+# Terraform Remote State
+
+[Terraform remote state](https://www.terraform.io/docs/providers/terraform/d/remote_state.html) retrieves state data from a [Terraform backend](https://www.terraform.io/docs/backends/index.html). This allows you to use the root-level outputs of one or more Terraform configurations as input data for another configuration.
+
+Typesciprt example usage:
+```typescript
+.....
+import { TerraformCloudRemoteState } from 'cdktf';
```
anubhavmishra, line 8: I would rename this to `TerraformCloudRemoteBackend` as that is essentially what you are configuring when you specify the following:

```
{
   organization: 'hashicorp',
   workspaces: {
       name: 'vpc-prod'
    }
}
```

examples/typescript/backends/s3/main.ts
```diff
@@ -1,5 +1,6 @@
 import { Construct } from 'constructs';
-import { App, TerraformStack, S3Backend } from 'cdktf';
+import { App, TerraformStack, S3Backend, S3RemoteState } from 'cdktf';
```
anubhavmishra, line 2: Terraform only supports one backend. Could you elaborate what is the difference between a `S3Backend` and `S3RemoteState`?

anubhavmishra:
In Terraform, a remote backend does both stores the state and perform remote operations in Terraform Cloud.

https://www.terraform.io/docs/backends/types/remote.html

There are standard backends that allow for storing the state in AWS S3, GCS bucket, etc. I am not sure if we want to introduce a "Remote State" concept into CDK for Terraform. Ideally, we should work with the "Remote Backend" vs "Standard Backend" concept.

jsteinich:
Remote State refers to https://www.terraform.io/docs/providers/terraform/d/remote_state.html. The `backend` arguments is the same name that is used when configuring a [backend](https://www.terraform.io/docs/backends/types/index.html) and the `config` argument takes the same parameters as the [backend](https://www.terraform.io/docs/backends/types/index.html).

anubhavmishra:
> Remote State refers to https://www.terraform.io/docs/providers/terraform/d/remote_state.html. The `backend` arguments is the same name that is used when configuring a [backend](https://www.terraform.io/docs/backends/types/index.html) and the `config` argument takes the same parameters as the [backend](https://www.terraform.io/docs/backends/types/index.html).

Yes, I think the key thing to note here is that [remote_state](https://www.terraform.io/docs/providers/terraform/d/remote_state.html) is a data resource which is different from a Terraform backend. We have [data source](https://github.com/hashicorp/terraform-cdk/blob/master/docs/working-with-cdk-for-terraform/data-sources.md) support in CDK for Terraform already. I would like us to use the same convention for remote state data source. For example, use `DataTerraformRemoteState` as the name with the following configuration:

```hcl
backend = "remote"

  config = {
    organization = "hashicorp"
    workspaces = {
      name = "vpc-prod"
    }
  }
```

Similarly, an AWS S3 remote state data source can be defined as follows:

```hcl
  backend = "s3"

  config = {
    bucket = "terraform-state-prod"
    key    = "network/terraform.tfstate"
    region = "us-east-1"
  }
```

skorfmann:
> Yes, I think the key thing to note here is that [remote_state](https://www.terraform.io/docs/providers/terraform/d/remote_state.html) is a data resource which is different from a Terraform backend. We have [data source](https://github.com/hashicorp/terraform-cdk/blob/master/docs/working-with-cdk-for-terraform/data-sources.md) support in CDK for Terraform already. I would like us to use the same convention for remote state data source. For example, use `DataTerraformRemoteState` as the name with the following configuration:

Was wondering if this could be generated somehow, but looks like this moved into core according to this https://registry.terraform.io/providers/hashicorp/terraform/latest So probably no easy accessible schema.

skorfmann:
e

> Yes, I think the key thing to note here is that [remote_state](https://www.terraform.io/docs/providers/terraform/d/remote_state.html) is a data resource which is different from a Terraform backend. We have [data source](https://github.com/hashicorp/terraform-cdk/blob/master/docs/working-with-cdk-for-terraform/data-sources.md) support in CDK for Terraform already. I would like us to use the same convention for remote state data source. For example, use `DataTerraformRemoteState` as the name with the following configuration:

And yes, prefixing with `Data` would make sense.

README.md
```diff
@@ -51,6 +51,7 @@ Choose a language:
 * Using the CDK for Terraform [escape hatch](./docs/working-with-cdk-for-terraform/escape-hatch.md).
 * Using the CDK for Terraform [tokens](./docs/working-with-cdk-for-terraform/tokens.md).
 * Using Terraform [data sources](./docs/working-with-cdk-for-terraform/data-sources.md).
+* Using Terraform [remote state](./docs/working-with-cdk-for-terraform/remote-state.md).
```
skorfmann, line 54: Since `remote state` is effectively a data source, let's move that to a section in the `data source` docs.

docs/working-with-cdk-for-terraform/remote-state.md
skorfmann: @anubhavmishra and I just talked about it, there was a bit confusion around the naming here. Will elaborate in a comment on the PR itself.

skorfmann:
As @anubhavmishra pointed out, `remote state` is just another form of a data source. Hence, it should be treated as such. 

For the class names, we thought about the following schema:

- `DataTerraformRemoteState` -> Terraform Cloud / Enterprise since `DataTerraformRemoteStateRemote` would look a bit weird.
- `DataTerraformRemoteStateS3`
- `DataTerraformRemoteStateConsult`

README.md
anubhavmishra: We can call it "Terraform Remote State".

jsteinich:
I'll go ahead and update the naming and docs.

I think it makes sense to keep them in the same files as the corresponding backend since there is pretty tight coupling between the two.

While it is data source, it doesn't have all the same properties as a regular data source, so I'm very hesitant to make it actually extend from `TerraformDataSource`.

skorfmann:
> While it is data source, it doesn't have all the same properties as a regular data source, so I'm very hesitant to make it actually extend from `TerraformDataSource`.

I wasn't necessarily expecting to extend from `TerraformDataSource`. However, since you brought it up - could you elaborate on the differences?

jsteinich:
- Remote state doesn't have the meta-arguments (dependsOn, Count, etc).
- It also isn't attached to a provider.
- The interpolated format is slightly different.

skorfmann:
> * Remote state doesn't have the meta-arguments (dependsOn, Count, etc).
> * It also isn't attached to a provider.
> * The interpolated format is slightly different.

Cool, thanks 👍  Let's not extend from `TerraformDataSource` then, but still stick to the suggested naming schema above.

jsteinich:
I've done the renaming and doc reorg. I think that covers it, but let me know of any other changes.

packages/cdktf/test/remote-state.test.ts
```diff
@@ -0,0 +1,341 @@
...
+        workspace: "my_workspace",
+        defaults: {
+            someProp: 'some_value'
+        }
+    });
+
+    expect(Testing.synth(stack)).toMatchSnapshot();
+})
+
+// This test is currently too sensitive to the environment it is ran in
```
skorfmann, line 289: Let me quickly check what the issue is with https://github.com/hashicorp/terraform-cdk/issues/256

packages/cdktf/lib/terraform-remote-state.ts
```diff
@@ -0,0 +1,64 @@
+import { Construct } from "constructs";
+import { TerraformElement } from "./terraform-element";
+import { Token } from "./tokens";
+import { deepMerge } from "./util";
+
+export interface TerraformRemoteStateConfig {
+    readonly workspace?: string;
+    readonly defaults?: { [key: string]: any };
+}
```
skorfmann, line 9: Does this apply to all remote state data sources?

jsteinich: Yeah, they do. Looks like I missed that in the renaming pass, though I'm not sure what the best name would be.

Some other approaches:

- Change to be closer to how `TerraformMetaArguments` and `TerraformResourceConfig` are setup and include as a base interface
- Make them optional parameters on the constructor
- Just have them be properties

Thoughts:

- Current approach is a bit awkward to use.
- The meta route is nice except that the config is shared with backends.
- Not all potential languages support optional parameters, so that isn't great.
- Having to create a local variable to set is inconvenient.

packages/cdktf/test/remote-state.test.ts
skorfmann: Well, let's go with `getString` for now. For some reason resolving the `Intrinsic` value doesn't work as expected. However, we don't need this here anyway.

packages/cdktf/lib/terraform-remote-state.ts
skorfmann: > * Change to be closer to how `TerraformMetaArguments` and `TerraformResourceConfig` are setup and include as a base interface

That's what I thought as well 👍 

> * The meta route is nice except that the config is shared with backends.

Couldn't the remote state props extend from `TerraformRemoteStateConfig` and `S3BackendProps`?

skorfmann:
Thanks for the updates, I'll look into this later today.

PR #185, opened 2020-07-16
Add Kubernetes TypeScript example

Redeux:
A Kubernetes example in TypeScript based on the config from the [HashiCorp Kubernetes Learn guide](https://learn.hashicorp.com/terraform/kubernetes/deploy-nginx-kubernetes)

PR #242, opened 2020-07-24
Add Kubernetes Python example

aareet:
Inspired by #185, a Kubernetes example in Python based on the config from the [HashiCorp Kubernetes Learn guide](https://learn.hashicorp.com/terraform/kubernetes/deploy-nginx-kubernetes).

### DIFF
--- a/examples/python-kubernetes/Pipfile
+++ b/examples/python-kubernetes/Pipfile
@@ -1,0 +1,10 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[requires]
+python_version = "3.7"
+
+[packages]
+cdktf = "~=0.0.12"
--- a/examples/python-kubernetes/Pipfile
+++ b/examples/python-kubernetes/Pipfile
@@ -7,4 +7,4 @@ verify_ssl = true
 python_version = "3.7"
 
 [packages]
-cdktf = "~=0.0.12"
+cdktf = {path = "./../../dist/python/cdktf-0.0.0-py3-none-any.whl"}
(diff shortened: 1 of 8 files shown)


examples/python-kubernetes/README.md
```diff
@@ -0,0 +1,43 @@
...
+The above command will create a folder called `cdktf.out` that contains all Terraform JSON configuration that was generated.
+
+Run Terraform commands
+
+```bash
+cd cdktf.out
+terraform init
+terraform plan
+terraform apply
+```
```
anubhavmishra, line 43: I would also add a section that uses `cdktf deploy` command to deploy the example.

aareet: Thanks! Added a cdktf deploy section and squashed my commits

examples/python-kubernetes/Pipfile
```diff
@@ -0,0 +1,10 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[requires]
+python_version = "3.7"
+
+[packages]
+cdktf = "~=0.0.12"
```
skorfmann, line 10: Please use the references [as here](https://github.com/hashicorp/terraform-cdk/blob/b8aeded4cb58a7f367195d5ed8a4632588f43eca/examples/python-aws/Pipfile)

```
[[source]]
name = "pypi"
url = "https://pypi.org/simple"
verify_ssl = true

[requires]
python_version = "3.7"

[packages]
cdktf = {path = "./../../dist/python/cdktf-0.0.0-py3-none-any.whl"}
```

aareet: Made the change 🍰

PR #254, opened 2020-07-27
Typescript Kubernetes Examples

skorfmann:
Recreating https://github.com/hashicorp/terraform-cdk/pull/185 by @Redeux where the branch got lost when the repo was published


skorfmann:
Should be good to go when build is green. The readme will be changed as batch later on

PR #255, opened 2020-07-27
Reorganize Examples into Language Folders

skorfmann:
As suggested by @jsteinich and discussed with @anubhavmishra this reorganizes the examples into language folders and fixed some of the examples along the way. Also added and changed the entries in the Readme.


jsteinich:
It might make sense to have another layer for some of the major cloud providers.
For example typescript/aws, typescript/gcs. Then the current folders of those names would actually get renamed to something like "simple" and be in typescript/aws/simple.
Would probably make sense to move the backend examples into those folders.

I'm not sure how many examples to plan for...

skorfmann:
> It might make sense to have another layer for some of the major cloud providers.

I'd keep it simple for now. When this list grows, I think this would mandate a separate repo. Something like this: https://github.com/aws-samples/aws-cdk-examples

Issue #201, opened 2020-07-17
Run builds for Windows as Well


skorfmann:
Since windows behaves slightly different, tests should be run for windows as well.

skorfmann:
Github actions support windows in general - see https://docs.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners 

Not sure what the best way would be to run windows tests. It'd be nice to run the integration tests somehow. Alternatively, we could come up with a dedicated test scenario for windows. Thoughts?

jsteinich:
Looks like the Windows runners have Docker, so an integration testing using the docker example might be a start.
Longer term it would probably be good have more thorough testing of the cli.

skorfmann:
> Looks like the Windows runners have Docker, so an integration testing using the docker example might be a start.

In the existing integration we've just a `null` resource deployment. That's good enough for the cli integration for now.

What I meant: I haven't touched in Windows in ages and will have to read up on how to basic things like installing node, npm. But perhaps the Github runner supports this already, don't know.

skorfmann:
These are the integration tests I was talking about: https://github.com/hashicorp/terraform-cdk/tree/master/test - We could try to add a dedicated windows integration build based on this https://github.com/hashicorp/terraform-cdk/tree/master/test/test-typescript-deploy

jsteinich:
There's a pretty extensive list of things installed: https://github.com/actions/virtual-environments/blob/master/images/win/Windows2019-Readme.md

I'm on Windows most of the time so I can do some local testing.

Do you have a preference between platform specific scripts vs writing something in JS/TS/Python that would be cross platform?

skorfmann:
> There's a pretty extensive list of things installed: https://github.com/actions/virtual-environments/blob/master/images/win/Windows2019-Readme.md

That looks sufficient, yes 👍 

> I'm on Windows most of the time so I can do some local testing.

That would be great, wanna take a stab at it?

> Do you have a preference between platform specific scripts vs writing something in JS/TS/Python that would be cross platform?

In the best case this would be delegated to JS. I messed this up already once :) see #205

PR #257, opened 2020-07-28
Windows integration tests

jsteinich:
Aiming to fix https://github.com/hashicorp/terraform-cdk/issues/201

Planning to convert most test scripts to JS and then run on a GitHub Windows runner.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,7 +1,5 @@
 name: Build
-on:
-  pull_request:
-  push:
+on: [pull_request]
 
 jobs:
   build:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -69,6 +69,7 @@ jobs:
           terraform_wrapper: false
       - name: Download dist
         uses: actions/download-artifact@v2
+        with:
           name: dist
       - name: integration tests
         run: yarn integration-windows
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,7 +38,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        terraform: ["0.12.29", 0.13.0-rc1]
+        terraform: ["0.12.29", "0.13.0"]
     container:
       image: hashicorp/jsii-terraform
     needs: build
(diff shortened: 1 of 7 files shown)


skorfmann:
I think I might have misunderstood your comment [here](https://github.com/hashicorp/terraform-cdk/issues/201#issuecomment-663525555), sorry!

> Do you have a preference between platform specific scripts vs writing something in JS/TS/Python that would be cross platform?

For the integration test, a script similar [to this](https://github.com/hashicorp/terraform-cdk/blob/8a5d26bf8b84cf19967f4bdb773fc405eca30a05/test/test-typescript-deploy/test.sh) would be good.

jsteinich:
Turns out this is currently blocked on https://github.com/hashicorp/terraform/issues/25068 as the apply fails with no useful information other than error code of 1.

jsteinich:
> Turns out this is currently blocked on [hashicorp/terraform#25068](https://github.com/hashicorp/terraform/issues/25068) as the apply fails with no useful information other than error code of 1.

I put in a workaround for this since the fix is only in the 0.13 beta version.

@skorfmann Would you like me to convert the other tests to also run on Windows? The Windows tests run in parallel to the linux ones, so it should take about the same wall clock time to run. I do think the python test is likely to fail at this point, so I'd probably skip that one.

If just this one test is good, then there's just a small amount of cleanup I'd like to do before merging.

One other note is that I'm only running the Windows tests against the 0.12.X version currently. It shouldn't be too difficult to add in 0.13 if desired.

.github/workflows/build.yml
```diff
@@ -28,8 +28,50 @@ jobs:
...
+          path: dist
       - name: integration tests
         run: yarn integration
         env:
           TERRAFORM_CLOUD_TOKEN: ${{ secrets.TERRAFORM_CLOUD_TOKEN }}
           TERRAFORM_BINARY_NAME: "terraform${{ matrix.terraform }}"
+
+  windows_integration:
+    runs-on: windows-latest
+    needs: build
```
skorfmann, line 61: Can we make this a matrix build as well?

skorfmann:
Pretty cool - thanks again 👍

PR #262, opened 2020-07-29
Python: setup imports dir as a package

cmclaughlin:
Running `pylint main.py` warns:

```
main.py:4:0: E0611: No name 'aws' in module 'imports' (no-name-in-module)
```

This change gets rid of that warning.

### DIFF
--- a/packages/cdktf-cli/lib/get/base.ts
+++ b/packages/cdktf-cli/lib/get/base.ts
@@ -71,6 +71,8 @@ export abstract class GetBase {
   private async harvestPython(targetdir: string, targetName: string) {
     const target = path.join(targetdir, targetName);
     await fs.move(`dist/python/src/${targetName}`, target, { overwrite: true });
+    // Make the codeMakerOutput dir a Python package for IDEs and linting
+    fs.writeFileSync(path.join(targetdir, '__init__.py'), '', 'utf-8');
   }
 
   protected abstract typesPath(name: string): string;
--- a/packages/cdktf-cli/templates/python/{{}}.gitignore
+++ b/packages/cdktf-cli/templates/python/{{}}.gitignore
@@ -1,5 +1,6 @@
 dist/
-imports/
+imports/*
+!imports/__init__.py
 .terraform
 cdktf.out
 terraform.tfstate*


skorfmann:
Just having an empty `__init__.py` file will fix this? 

> But I'm not sure how to reference the `codeMakerOutput` directory name from
> there.

It'd be either set in `cdktf.json` or [default](https://github.com/hashicorp/terraform-cdk/blob/d785af4bf2dad8f1a30b34b8e751ece967c7680d/packages/cdktf-cli/lib/config.ts#L6-L9) to `.gen` if unset. But yes, it'd be a bit disconnected. The info could be passed down to the template, though  - see [here](https://github.com/hashicorp/terraform-cdk/blob/d785af4bf2dad8f1a30b34b8e751ece967c7680d/packages/cdktf-cli/bin/cmds/init.ts#L80). You'd have to load the config similar [to here](https://github.com/hashicorp/terraform-cdk/blob/d785af4bf2dad8f1a30b34b8e751ece967c7680d/packages/cdktf-cli/bin/cmds/get.ts#L9). What do you think @cmclaughlin ?

cmclaughlin:
@skorfmann thanks for the feedback!

Yep, adding an empty `__init__.py` file to a directory makes it a Python module. And that gets rid of the pylint warning, which I think is worth doing.

I was able to pass in `codeMakerOutput` to the template, but as you noted it defaults to .get at that point. So I refactored and simply read it in from `cdktf.json` which at that point has the right value.

Please take another look and let me know what you think.

skorfmann:
> I was able to pass in `codeMakerOutput` to the template, but as you noted it defaults to .get at that point. So I refactored and simply read it in from `cdktf.json` which at that point has the right value.

Hm, with [readConfigSync](https://github.com/hashicorp/terraform-cdk/blob/d785af4bf2dad8f1a30b34b8e751ece967c7680d/packages/cdktf-cli/lib/config.ts#L23-L27) it should only be the default if not set in `cdktf.json`

cmclaughlin:
Guessing `cdktf.json` doesn't exist yet... I'll take another look.

cmclaughlin:
I double checked... with this change `codeMakerOutput` is `.gen`, not 'imports'.

```diff
diff --git a/packages/cdktf-cli/bin/cmds/init.ts b/packages/cdktf-cli/bin/cmds/init.ts
index c858054..9f80d01 100644
--- a/packages/cdktf-cli/bin/cmds/init.ts
+++ b/packages/cdktf-cli/bin/cmds/init.ts
@@ -8,6 +8,8 @@ import * as terraformCloudClient from './helper/terraform-cloud-client';
 import * as chalk from 'chalk';
 import { terraformCheck } from './terraform-check';
 import { displayVersionMessage } from './version-check'
+import { readConfigSync } from '../../lib/config';
+
 
 const chalkColour = new chalk.Instance();
 
@@ -75,9 +77,10 @@ This means that your Terraform state file will be stored locally on disk in a fi
     }
 
     const deps: any = await determineDeps(argv.cdktfVersion, argv.dist);
+    const config = readConfigSync();
 
     await sscaff(templateInfo.Path, '.', {
-      ...deps, ...projectInfo
+      ...deps, ...projectInfo, ...config
     });
   }
 }
diff --git a/packages/cdktf-cli/templates/python/.hooks.sscaff.js b/packages/cdktf-cli/templates/python/.hooks.sscaff.js
index d4af60b..ef9d801 100644
--- a/packages/cdktf-cli/templates/python/.hooks.sscaff.js
+++ b/packages/cdktf-cli/templates/python/.hooks.sscaff.js
@@ -22,6 +22,8 @@ exports.post = options => {
     terraformCloudConfig(options.$base, options.OrganizationName, options.WorkspaceName)
   }
 
+    console.log(options);
+
```
 
Which outputs:

```javascript
{
  npm_cdktf: 'cdktf@^0.0.1',
  npm_cdktf_cli: 'cdktf-cli@^0.0.1',
  pypi_cdktf: 'cdktf~=0.0.1',
  Name: 'xxx',
  Description: 'xxx',
  OrganizationName: '',
  WorkspaceName: '',
  output: 'cdktf.out',
  codeMakerOutput: '.gen',
  '$base': 'turd5'
}
```

skorfmann:
Oh yeah, that makes sense when thinking about it. This will be loaded in the CLI before the template is applied. Sorry for the confusion.

skorfmann:
So, after having thought about this again: I think this should be part of the [code making process](https://github.com/hashicorp/terraform-cdk/blob/d785af4bf2dad8f1a30b34b8e751ece967c7680d/packages/cdktf-cli/lib/get/base.ts#L63), and not be retrofitted in the template. It'll be much simpler as well, since we have all the context we need at this point.

Sorry again for the initial confusion about this. Could you adapt it accordingly?

cmclaughlin:
Thanks for the feedback! I think we're all set... I agree it's much cleaner now.

Issue #265, opened 2020-07-29
Build watching for changes


cmclaughlin:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

I'm not very familiar with TypeScript, Yarn, etc, so I'm not sure about this... but just getting started making some changes it would be nice to "compile" one file at a time and not the entire project. `yarn build` takes a while and drains my laptop battery 😄 

I see there's a watch option in`packages/cdktf-cli/package.json` -- could we have something like that in the root of the repo? In other words, could we run `yarn watch` in the root of the repo and have it watch/compile changes to `packages/cdktf*`?

Also, expanding the docs on environment setup would good too... I'm not sure if this is the right way to get started, but I edited `packages/cdktf/package.json` and specified a version there (.e.g. 0.0.1). And then used `--cdktf-version=0.0.1` with my copy of `cdktf` - without that I was getting this error:

    cannot use version 0.0.0, use --cdktf-version, --dist or CDKTF_DIST to install from a "dist" directory

skorfmann:
Thanks for raising this issue @cmclaughlin - I just created #266 to improve the guide and added a `yarn watch` command. Would be great if you could provide feedback on this.

skorfmann:
Fixed in https://github.com/hashicorp/terraform-cdk/pull/266

PR #266, opened 2020-07-30
Improve contribution guide and global watch

skorfmann:
As pointed out in https://github.com/hashicorp/terraform-cdk/issues/265 the current contribution guide is very basic. This PR improves this guide and adds a `yarn watch` command to build `cdktf` and `cdktf-cli` packages:

Screenshot 2020-07-30 at 11 19 35

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -74,13 +74,15 @@ If you'd want this permanently, you can add this line to your profile settings (
 
 ### Create link
 
+Let's link `cdktf` and `cdktf-cli`, run the following the repository root folder:
+
 ```shell
-$ cd packages/cdktf-cli
-$ yarn link
-$ cd -
+$ yarn link-packages
 $ cdktf --version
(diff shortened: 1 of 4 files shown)


cmclaughlin:
I just went through the setup instructions... it's much clearer now!

One idea... could these two `yarn link` commands be combined into one block? e.g. 

```
$ cd packages/cdktf-cli
$ yarn link
$ cd ../cdktf
$ yarn link
$ cd ../..
$ cdktf --version
0.0.0
```

The way it is now, the tutorial has the reader `cd ~/my-local-cdktf-example` then assume they're back in the root of this repo (to run `cd packages/cdktf`). So I think it's more clear to group those two `yarn link` commands together.

However, I think the "Build & Package" section might need to go before the "Create link" section for this to work.

skorfmann:
Cool, thanks for the feedback. Adapted it a bit to make the linking easier.

Issue #268, opened 2020-07-30
cdktf init pipenv fails


sonfrau:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### cdktf & Language Versions

I'm on Ubuntu 20.04

Terraform version : v0.12.29
Node version : v17.7.0
Yarn version : 1.22.4
Pipenv : version 2020.6.2
Python 3.8.2

### Affected Resource(s)

I was not able to run your simple python example. I didn't have got any problem with the Typescript example.

Any clue would be pretty thankful for avoiding this behaviour.

### Expected Behavior

New stack/template with no errors.

### Actual Behavior
```
 ~/python-cdktf  $ cdktf init --template="python" --local
Note: By supplying '--local' option you have chosen local storage mode for storing the state of your stack.
This means that your Terraform state file will be stored locally on disk in a file 'terraform.tfstate' in the root of your project.

We will now set up the project. Please enter the details for your project.
If you want to exit, press ^C.

Project Name: (default: 'python-cdktf')
Project Description: (default: 'A simple getting started project for cdktf.')
Warning: Your Pipfile requires python_version 3.7, but you are using 3.8.2 (/home/miquel/.local/share/v/p/bin/python).
  $ pipenv --rm and rebuilding the virtual environment may resolve the issue.
  $ pipenv check will surely fail.
Pipfile.lock not found, creating…
Locking [dev-packages] dependencies…
Locking [packages] dependencies…
Updated Pipfile.lock (a65489)!
Installing dependencies from Pipfile.lock (a65489)…
  🐍   ▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉ 0/0 — 00:00:00
To activate this project's virtualenv, run pipenv shell.
Alternatively, run a command inside the virtualenv with pipenv run.
Warning: Your Pipfile requires python_version 3.7, but you are using 3.8.2 (/home/miquel/.local/share/v/p/bin/python).
  $ pipenv --rm and rebuilding the virtual environment may resolve the issue.
[...]

Issue #267, opened 2020-07-30
cdktf init " 'which' is not recognized as an internal or external command, operable program or batch file."


LucaMali:
When launching the "cdktf init" command to prepare the project folder with python language i get this error "'which' is not recognized as an internal or external command, operable program or batch file. Unable to find "pipenv". Install from https://pipenv.kennethreitz.org" 

I'm on Windows 10 and even if pipenv is installed I get the "Unable to find 'pipenv' error" 

Terraform version : v0.12.29
Node version : v12.18.2
Yarn version : 1.22.4
Pipenv : version 2020.6.2
Python 3.7.8

![Capture](https://user-images.githubusercontent.com/46742509/88909873-f3102580-d25b-11ea-90e9-e9bd00cc2137.PNG)

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

skorfmann:
Hey @LucaMali - thanks for creating this issue. 

This looks like a system specific issue to me. @jsteinich since you have more experience running `cdktf` on windows, do you have an idea what could cause this?

jsteinich:
I can confirm the issue. It is caused by https://github.com/hashicorp/terraform-cdk/blob/d785af4bf2dad8f1a30b34b8e751ece967c7680d/packages/cdktf-cli/templates/python/.hooks.sscaff.js#L9

I'll try to work on a fix tonight.

LucaMali:
Ok thank you.
I'm gonna wait for updates.

skorfmann:
Fixed by #270

PR #270, opened 2020-07-31
Windows Python fixes

### LINKED ISSUE #194: Windows Javascript error when running `cdktf init --template="python"`
### Community Note

### Expected Behavior

No error.
### Actual Behavior

Experienced this error when using the CLI to init my Python project.
### Steps to Reproduce

Run `cdftf init --template="python"`
### References

image

* #0000

jsteinich:
Aiming to fix #194 , #267, #268 and anything else that comes up while getting python working on Windows.

### DIFF
--- a/packages/cdktf-cli/templates/python/Pipfile
+++ b/packages/cdktf-cli/templates/python/Pipfile
@@ -2,6 +2,3 @@
 name = "pypi"
 url = "https://pypi.org/simple"
 verify_ssl = true
-
-[requires]
-python_version = "3.7"
--- a/packages/cdktf-cli/templates/python/Pipfile
+++ b/packages/cdktf-cli/templates/python/Pipfile
@@ -2,3 +2,6 @@
 name = "pypi"
 url = "https://pypi.org/simple"
 verify_ssl = true
+
+[requires]
+python_version = "3"
--- a/packages/cdktf-cli/templates/python/Pipfile
+++ b/packages/cdktf-cli/templates/python/Pipfile
@@ -4,4 +4,4 @@ url = "https://pypi.org/simple"
 verify_ssl = true
 
 [requires]
-python_version = "3"
+python_version = "3.7"
--- a/packages/cdktf-cli/templates/python/Pipfile
+++ b/packages/cdktf-cli/templates/python/Pipfile
@@ -4,4 +4,4 @@ url = "https://pypi.org/simple"
 verify_ssl = true
 
 [requires]
-python_version = "3.7"
+python_version = "3"
--- a/packages/cdktf-cli/templates/python/cdktf.json
+++ b/packages/cdktf-cli/templates/python/cdktf.json
@@ -1,6 +1,6 @@
 {
   "language": "python",
-  "app": "pipenv run ./main.py",
+  "app": "pipenv run main.py",
   "terraformProviders": ["aws@~> 2.0"],
   "codeMakerOutput": "imports"
 }
--- a/packages/cdktf-cli/templates/python/cdktf.json
+++ b/packages/cdktf-cli/templates/python/cdktf.json
@@ -1,6 +1,6 @@
 {
   "language": "python",
-  "app": "pipenv run main.py",
+  "app": "pipenv run python main.py",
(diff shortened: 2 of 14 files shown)


jsteinich:
I'm making progress but I think I'm getting blocked by something related to #262.
Following the python getting started, I'm getting `ImportError: cannot import name 'Instance' from 'imports.aws' (unknown location)`

packages/cdktf-cli/templates/python/Pipfile
```diff
@@ -4,4 +4,4 @@ url = "https://pypi.org/simple"
 verify_ssl = true
 
 [requires]
-python_version = "3.7"
\ No newline at end of file
+python_version = "3"
```
skorfmann, line 7: I don't think that'll work - see https://github.com/awslabs/cdk8s/blob/5759d0d927268bb4e9cb9ba01d13053b44834fed/docs/getting-started/python.md#prerequisites

packages/cdktf-cli/lib/get/base.ts
```diff
@@ -48,7 +48,7 @@ export abstract class GetBase {
         });
 
         const pacmak = require.resolve('jsii-pacmak/bin/jsii-pacmak');
-        await shell(pacmak, [ '--target', options.targetLanguage, '--code-only' ]);
+        await shell('node', [ pacmak, '--target', options.targetLanguage, '--code-only' ]);
```
skorfmann, line 51: According to [this comment](https://github.com/awslabs/cdk8s/pull/183), it might be worthwhile implementing https://github.com/aws/jsii-srcmak rather than fixing this here.

packages/cdktf-cli/templates/python/cdktf.json
```diff
@@ -1,6 +1,6 @@
 {
   "language": "python",
-  "app": "pipenv run ./main.py",
+  "app": "pipenv run python main.py",
```
skorfmann, line 3: That's cross platform compatible?

packages/cdktf-cli/templates/python/Pipfile
jsteinich: We might want to remove entirely then or switch away from pipenv since something like ">=3.7" isn't supported.

packages/cdktf-cli/lib/get/base.ts
jsteinich: I saw you mentioned that elsewhere. I'm not really familiar with what that entails, but I can take a look.

skorfmann: see #272

packages/cdktf-cli/templates/python/cdktf.json
jsteinich: It worked in the two platforms I tested on.
On *nix systems the shebang line allows for switching the interpreter. On Windows that doesn't work, so explicit specification is generally necessary.

packages/cdktf-cli/templates/python/Pipfile
skorfmann: Hm, when looking at the [jsii build config](https://github.com/aws/jsii/blob/6ea8daa8a022c5bda5fd1854d49292e65caac2be/.github/workflows/main.yml) it looks like at least 3.6 - 3.8 should work. And according to [this](https://www.python.org/downloads/) 3.5 won't be supported soon. 

Screenshot 2020-07-31 at 16 15 14

@eladb are there docs which specify Python version support for `jsii`?

skorfmann:
@jsteinich Do you think this might be easier now since #276 got merged?

jsteinich:
> @jsteinich Do you think this might be easier now since #276 got merged?

Doesn't really change much since I had worked around some of the things it fixed, but I was waiting on it.
I should hopefully get this working in the next few days.

packages/cdktf-cli/templates/python/Pipfile
skorfmann: Well, let's just go with the major version then. Seems to work fine.

skorfmann (approved):

Issue #272, opened 2020-07-31
Use jsii-srcmak


skorfmann:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

As suggested by @eladb we should transition to https://github.com/aws/jsii-srcmak for the code generator process in the cli. This will likely help with #270 and paves the way for #197 

It's used in [cdk8s](https://github.com/awslabs/cdk8s/blob/f0ef3b467969a90efcbfc663353565c8639bb5bd/packages/cdk8s-cli/lib/import/base.ts) already. It can be implemented in a similar way.

### References

- #270 
- #197

jsteinich:
Can probably use https://github.com/awslabs/cdk8s/pull/205 to help guide the process.

skorfmann:
Implemented in #276

Issue #98, opened 2020-06-09
Prebuilt Provider Packages


skorfmann:
When thinking about distributing L2 or L3 constructs, we'll need provider packages. Without, it's quite hard to define the dependency on used providers (e.g. Null Provider). 

![CleanShot 2020-06-07 at 15 27 38](https://user-images.githubusercontent.com/136789/84129593-75d9e880-aa42-11ea-974f-ccfc4fef18ae.png)

## Current State 

In order to do this at the moment by the example of `aws-ecr-asset`:

- Create / Publish `aws-ecr-asset` package
- Add a note in the Readme / post install to add the `null` provider to `cdktf.json` 
- When there's a breaking change in one of the upstream packages, it'll break and the user will have dig through weird error messages

## Desired State

We have prebuilt provider packages, with potentially a few predefined L2 / L3 constructs for each of these providers, which allows leaving the dependency management to proper tooling (npm et al). This would also get us closer to a workflow where there's essentially no need for a local Terrafor CLI (if the Terraform Cloud workflow is chosen).

The version of distributed packages shouldn't be tied to the original provider, but always track latest. The version would be mandated by changes to `cdktf` the `<provider>-constructs` and possibly breaking changes by the actual underlying Terraform Provider / Engine.

- The provider package gets auto-generated via `cdktf get`
- It tracks `latest` of the provider
- Users which require a specific provider version for some reason, will have to fallback to generate it manually by `cdktf get`

### Proposed Repo Structure 

To allow individiual versioning of providers, I think it makes sense to use something like the following repo structure

#### hashicorp/terraform-cdk

- packages/cdktf
- packages/cdktf-cli

#### hashicorp/terraform-cdk-aws

- packages/@cdktf/aws # metapackage to pull in cdktf / aws provider / aws constructs
- packages/@cdktf/aws-provider
- packages/@cdktf/aws-constructs

#### hashicorp/terraform-cdk-azure

- packages/@cdktf/azure # metapackage to pull in cdktf / azure provider / azure constructs
- packages/@cdktf/azure-provider
- packages/@cdktf/azure-constructs

## Conclusion

Personally, I think that L2 / L3 constructs will be the main selling point for the Terraform CDK. Therefore I'd consider this quite important. 

Thoughts on this one?

jsteinich:
[...]

PR #276, opened 2020-08-02
Migrate to jsii-srcmak

jsteinich:
Address #272 

Python integration test is currently failing, but all other tests are passing.


jsteinich:
It's also extremely slow, so hopefully that's related to the failure

jsteinich:
I believe the error I'm encountering is the same as https://github.com/awslabs/cdk8s/issues/273 since both the aws provider and the vpc module are getting set to the `generated` namespace.

jsteinich:
@eladb Before I start profiling things, do you have any idea why jsii-srcmak would be so much slower than jsii + jsii-pacmak in a simple wrapper script?

eladb:
> @eladb Before I start profiling things, do you have any idea why jsii-srcmak would be so much slower than jsii + jsii-pacmak in a simple wrapper script?

No idea. I haven't noticed this myself but now I am curious. Let me know what you find out!

skorfmann:
> It's also extremely slow, so hopefully that's related to the failure

Yes, for Python it  feels pretty slow indeed.

> I believe the error I'm encountering is the same as [awslabs/cdk8s#273](https://github.com/awslabs/cdk8s/issues/273) since both the aws provider and the vpc module are getting set to the `generated` namespace.

Looks like there might [be a fix](https://github.com/aws/jsii-srcmak/pull/22) on master. But the build [failed](https://github.com/aws/jsii-srcmak/runs/939271677) for some reason and therefore no package was published. Perhaps we can give it a shot straight from github?

eladb:
> Looks like there might [be a fix](https://github.com/aws/jsii-srcmak/pull/22) on master. But the build [failed](https://github.com/aws/jsii-srcmak/runs/939271677) for some reason and therefore no package was published. Perhaps we can give it a shot straight from github?

I am on it

eladb:
> > Looks like there might [be a fix](https://github.com/aws/jsii-srcmak/pull/22) on master. But the build [failed](https://github.com/aws/jsii-srcmak/runs/939271677) for some reason and therefore no package was published. Perhaps we can give it a shot straight from github?
> 
> I am on it

Released

skorfmann:
> > It's also extremely slow, so hopefully that's related to the failure
> 
> Yes, for Python it feels pretty slow indeed.

Well, without measuring it - the current version feels slow as well. :)

skorfmann:
> Released

Thanks!

packages/cdktf-cli/test/get/util.ts
```diff
@@ -1,31 +1,29 @@
...
-      await jsiiCompile(workdir, {
-        stdout: true,
-        name: name,
-        main: name,
-        providerPath: './providers/aws/index'
-      });
+      const manifest = JSON.parse(await fs.readFile(jsiiPath, 'utf-8'));
+
+      // patch cdktf version in manifest because it's not stable
+      manifest.dependencies.cdktf = '999.999.999';
```
skorfmann, line 24: Could you elaborate on this number?

jsteinich: It could be anything. That value is just what cdk8s uses.

The idea is that the dependency version doesn't provide any meaningful information to the snapshot and so should be effectively ignored.

jsteinich:
In my testing, the python integration test went from 4.5 minutes to 27 minutes with this change. I don't think we should merge until that time is cut down.

I haven't definitively tracked down the cause yet. My current hunch is that this line: https://github.com/aws/jsii-srcmak/blob/b3ca41e005af735adef79422033d179192875dfc/src/srcmak.ts#L20 is causing a lot of unnecessary file io.

skorfmann:
> In my testing, the python integration test went from 4.5 minutes to 27 minutes with this change. I don't think we should merge until that time is cut down.

Ok, that's a huge difference. I'll try this on my machine again. Which kind of setup do you have? Linux, Mac, Windows?

jsteinich:
> Ok, that's a huge difference. I'll try this on my machine again. Which kind of setup do you have? Linux, Mac, Windows?

That's running the same docker container used by the build workflow using docker desktop wsl2 backend on a wsl2 filesystem from Windows.

The github runner does also show a decent increase in time as well: going from ~7 minutes for all integration tests to ~23 minutes.

eladb:
> I haven't definitively tracked down the cause yet. My current hunch is that this line: https://github.com/aws/jsii-srcmak/blob/b3ca41e005af735adef79422033d179192875dfc/src/srcmak.ts#L20 is causing a lot of unnecessary file io.

Sounds like this is something we should be able to fix. Can you guys raise an issue towards jsii-srcmak and we will tackle in short order?

skorfmann:
> The github runner does also show a decent increase in time as well: going from ~7 minutes for all integration tests to ~23 minutes.

I'm experementing with #98 and a full [jsii build](https://github.com/skorfmann/cdktf-provider-aws/runs/948707228?check_suite_focus=true) takes 9 minutes. That's with `cdktf@latest` and `"jsii": "^1.9.0",`. So, it's probably not `jsii-srcmak` which got slower, but `jsii` itself?

Screenshot 2020-08-05 at 12 10 29

Also, we probably should upgrade and align all `jsii` dependencies.

jsteinich:
> > I haven't definitively tracked down the cause yet. My current hunch is that this line: https://github.com/aws/jsii-srcmak/blob/b3ca41e005af735adef79422033d179192875dfc/src/srcmak.ts#L20 is causing a lot of unnecessary file io.
> 
> Sounds like this is something we should be able to fix. Can you guys raise an issue towards jsii-srcmak and we will tackle in short order?

While there is some unnecessary work being done (already being called from a temporary directory), this doesn't appear to be the issue. I'm still working to track down the main cause, but once I have more conclusive information I'll make sure to file anything.

eladb:
Copying @RomainMuller

RomainMuller:
I suggest you give [`0x`](https://www.npmjs.com/package/0x) a shot to try to figure out where time is spent.

skorfmann:
@jsteinich thanks for creating the [issue](https://github.com/aws/jsii/issues/1856) in the `jsii` repo

skorfmann:
is my assumption correct, that this is mostly a CI issue around integration tests? I did some local testing again, and the times for `cdktf get` for the `aws` example were ranging between one minutes (Mac OSX) and roughly two minutes (Docker `hashicorp/jsii-terraform`), for both `master` and this branch.

If so, I'd probably ignore the performance regression for now. 

If it's an actual user facing issue, we should submit a PR to fix https://github.com/aws/jsii/issues/1856

Thoughts?

jsteinich:
> mostly a CI issue around integration tests

The other place where it will show up is the first time someone sets up a python project, or every project init if they use a different virtual environment per project.

skorfmann:
> > mostly a CI issue around integration tests
> 
> The other place where it will show up is the first time someone sets up a python project, or every project init if they use a different virtual environment per project.

So, a `cdktf init --language python` would cause it?.

jsteinich:
> So, a `cdktf init --language python` would cause it?.

If it's the first time running on a system / virtual env, then yes. Subsequent runs use cached data and are just subject to the standard slowness.

skorfmann:
> Subsequent runs use cached data and are just subject to the standard slowness.

Do you know how to clear that cache to reproduce easily?

jsteinich:
> Do you know how to clear that cache to reproduce easily?

Easiest repro I know is creating a new pipenv.
https://github.com/psf/black/issues/248#issuecomment-391456129 might work for clearing.

skorfmann:
> Easiest repro I know is creating a new pipenv.

Hm, I believe I did that, and in a Docker container I'd get a new one anyway, I guess.

skorfmann:
@jsteinich what do you think, push for a fix in jsii or just merge it? For the CI case, we could just have a `black` stub: 

```
cat /usr/local/bin/black
#! /bin/bash

exit 0
```

😆

jsteinich:
> what do you think, push for a fix in jsii or just merge it?

Doesn't look like it would be too difficult to create a PR to make `black` optional. I don't know how long it would take to get changes approved and released though. We'd also have to get it added to jsii-srcmak as well, but that seems more likely to go quickly.

I'm not sure the urgency of better Windows support, Java support, and anything else that this is holding up though. If any of those are wanted for the next release, I think it would make sense to get this merged in, but then hold on releasing until performance was improved.

eladb:
We can short circuit `black` in srcmak in the meantime (just include a stub and add it to the beginning of the `PATH` when we execute Pac-Pak). 

Once merge it will be immediately released.

skorfmann:
> I'm not sure the urgency of better Windows support, Java support, and anything else that this is holding up though. 

It would be good to get this merged, so we don't have too many parallel tracks here. 

> We can short circuit `black` in srcmak in the meantime (just include a stub and add it to the beginning of the `PATH` when we execute Pac-Pak).

Also a good idea 👍

eladb:
> > I'm not sure the urgency of better Windows support, Java support, and anything else that this is holding up though. 
> 
> 
> 
> It would be good to get this merged, so we don't have too many parallel tracks here. 
> 
> 
> 
> > We can short circuit `black` in srcmak in the meantime (just include a stub and add it to the beginning of the `PATH` when we execute Pac-Pak).
> 
> 
> 
> Also a good idea 👍 
> 
> 
> 
> 
> 
> 

Would you guys like to pick this up?

jsteinich:
> Would you guys like to pick this up?

I'd feel comfortable stubbing `black` within cdktf, but I haven't managed to get a custom version of `jsii-srcmak` to work when calling from cdktf yet.

eladb:
On it

eladb:
Okay, coming up: https://github.com/aws/jsii/pull/1864 (by @RomainMuller)

jsteinich:
@eladb Any idea when the next jsii release will be?

skorfmann:
> @eladb Any idea when the next jsii release will be?

Just go released a few hours ago https://github.com/aws/jsii/releases/tag/v1.11.0

skorfmann:
This is likely relevant: https://github.com/hashicorp/terraform-cdk/pull/328

jsteinich:
> > @eladb Any idea when the next jsii release will be?
> 
> Just got released a few hours ago https://github.com/aws/jsii/releases/tag/v1.11.0

Looks like jsii-srcmak doesn't pass tests when upgrading to jsii 1.11.0. Once that's fixed and released (let me know if I can help), then I'll get this updated.

eladb:
> > > @eladb Any idea when the next jsii release will be?
> > 
> > 
> > Just got released a few hours ago https://github.com/aws/jsii/releases/tag/v1.11.0
> 
> Looks like jsii-srcmak doesn't pass tests when upgrading to jsii 1.11.0. Once that's fixed and released (let me know if I can help), then I'll get this updated.

Released jsii-srcmak with jsii 1.11.0

skorfmann:
Awesome work @jsteinich and thanks for the quick support @eladb @RomainMuller

anubhavmishra:
@skorfmann can we close #272 ?

Issue #249, opened 2020-07-26
Add support for Terraform Variables


skorfmann:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

It would be good to support Terraform variables as first class construct. This will be helpful for hybrid approaches and for creating a Terraform module from constructs.

vmarchaud:
I'm not sure i understand correctly, but does this issue means that we can't use terraform variable right now using CDK ?

skorfmann:
> I'm not sure i understand correctly, but does this issue means that we can't use terraform variable right now using CDK ?

There's no explicit construct to support variables at the moment. Variables still can be supported via [escape hatches](https://cdk.tf/escape-hatch). See [this](https://dev.to/skorfmann/a-terraform-cdk-construct-which-doubles-as-native-terraform-module-447k/#input-variables) for a full example.

Here an excerpt:

```ts
stack.addOverride('variable', {
  tags: {
    description: "Tags for the instance",
    type: "map(string)"
  },
  instance_type: {
    description: "Instance type",
    type: "string"
  }
})
```

vmarchaud:
@skorfmann Thanks for the quick answer, i believe adding this into the readme would be helpful to people  looking to migrate to CDK :)

skorfmann:
> @skorfmann Thanks for the quick answer, i believe adding this into the readme would be helpful to people looking to migrate to CDK :)

No, worries. Adding this to the docs would make sense, yes 👍 

Out of curiosity, could you elaborate on your use-case for variables?

vmarchaud:
Sure, here is some options (in TS) on the infrastructure i'm trying to migrate:
```ts
type InfrastructureToolingOptions = {
  environment: string
  gcp_credentials: string
  sva_dns: string
  project: string
  region: string

  gke: {
    pool_min: number
    pool_max: number
    pool_type: number
    version: string,
    nodes_cidr: string
    pods_cidr: string
    master_cidr: string
  }
}
```

[...]

Issue #247, opened 2020-07-25
Forceable Static Name for Outputs


skorfmann:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

When aiming to build a Terraform module with `cdktf`, or reference the remote state in a non automated environment, it would be good to be able to force a static name.

Given this output:

```ts
new TerraformOutput(this, 'arn', {
      value: custom.instance.arn
    })
```

This generates a name like this `cdktfhybridmodule_arn_788D4ED0`.

A new option could be:

```ts
new TerraformOutput(this, 'arn', {
      value: custom.instance.arn,
      staticName: "arn"
    })
```

Which wouldn't generate a name but use the input and would be referenceable as `arn`.

skorfmann:
Or perhaps just use the construct id?

```ts
new TerraformOutput(this, 'arn', {
      value: custom.instance.arn,
      static: true
    })
```

=> `arn`

jsteinich:
I think the construct id would work well.

skorfmann:
This can already be achieved by using a [feature flag](https://github.com/hashicorp/terraform-cdk/blob/f868537f8a954f2aaa7c22304dda24d8cfe75d1e/docs/working-with-cdk-for-terraform/feature-flags.md) (default for new apps) - It was implemented in https://github.com/hashicorp/terraform-cdk/pull/329

This applies to all root level resources. Since it's not really granular manageable, I'll leave this issue open for now.

jsteinich:
#329 also added `overrideLogicalId` which can control the name for individual resources.

PR #277, opened 2020-08-02
Terraform Variables

jsteinich:
Adds support for defining Terraform Variables #249

### DIFF
--- a/packages/cdktf/lib/terraform-variable.ts
+++ b/packages/cdktf/lib/terraform-variable.ts
@@ -1,0 +1,133 @@
+import { Construct } from "constructs";
+import { TerraformElement } from "./terraform-element";
+import { keysToSnakeCase, deepMerge } from "./util"
+import { Token } from "./tokens";
+
+export interface TerraformVariableConfig {
+    readonly default?: any;
+    readonly description?: string;
+    readonly type?: ITerraformVariableType;
+}
+
+export interface ITerraformVariableType {
+    /**
+     * @internal
+     */
+    _render(): string;
+}
+
+export class TerraformPrimitiveVariableType implements ITerraformVariableType {
+    public static readonly STRING = new TerraformPrimitiveVariableType('string');
+    public static readonly NUMBER = new TerraformPrimitiveVariableType('number');
+    public static readonly BOOL = new TerraformPrimitiveVariableType('bool');
+    public static readonly ANY = new TerraformPrimitiveVariableType('any');
+
+    constructor(private readonly type: string) {
+
+    }
+
+    /**
+     * @internal
+     */
+    _render(): string {
+        return this.type;
+    }
+}
+
+export enum TerraformCollectionType {
+    LIST = 'list',
+    MAP = 'map',
+    SET = 'set'
+}
+
+export class TerraformCollectionVariableType implements ITerraformVariableType {
+    constructor(public readonly collectionType: TerraformCollectionType, public readonly elementType: ITerraformVariableType) {
+    }
+
+    /**
+     * @internal
+     */
+    _render(): string {
+        return `list(${this.elementType._render()})`;
+    }
+}
+
+export class TerraformTupleVariableType implements ITerraformVariableType {
+    constructor(public readonly elements: ITerraformVariableType[]) {
+
+    }
+
+    /**
+     * @internal
+     */
+    _render(): string {
+        return `tuple(${this.elements.map(e => e._render()).join(", ")})`;
+    }
+}
+
+export class TerraformObjectVariableType implements ITerraformVariableType {
+    constructor(public readonly attributes: { [k: string]: ITerraformVariableType }) {
+
+    }
+
+    /**
+     * @internal
+     */
+    _render(): string {
+        return `object({${Object.keys(this.attributes).map(k => k + "=" + this.attributes[k]._render()).join(", ")}})`;
+    }
+}
+
+export class TerraformVariable extends TerraformElement {
+    public readonly default?: any;
+    public readonly description?: string;
+    public readonly type?: ITerraformVariableType;
+
+    constructor(scope: Construct, id: string, config: TerraformVariableConfig) {
+        super(scope, id);
+
+        this.default = config.default;
+        this.description = config.description;
+        this.type = config.type;
+    }
+
+    public get stringValue(): string {
+        return Token.asString(this.interpolation());
+    }
+
+    public get numberValue(): number {
+        return Token.asNumber(this.interpolation());
+    }
+
+    public get listValue(): string[] {
+        return Token.asList(this.interpolation());
+    }
+
+    public get booleanValue(): boolean {
+        return Token.asString(this.interpolation()) as any as boolean
+    }
+
+    public get value(): any {
+        return Token.asAny(this.interpolation());
+    }
+
+    private interpolation(): any {
+        return `\${var.${this.friendlyUniqueId}}`
+    }
+
+    public synthesizeAttributes(): { [key: string]: any } {
+        return {
+            default: this.default,
+            description: this.description,
+            type: this.type?._render()
+        }
+    }
+
+    public toTerraform(): any {
+        return {
+            variable: {
+                [this.friendlyUniqueId]: deepMerge(keysToSnakeCase(this.synthesizeAttributes()), this.rawOverrides)
+            }
+        };
+    }
+}
--- a/packages/cdktf/lib/terraform-variable.ts
+++ b/packages/cdktf/lib/terraform-variable.ts
@@ -6,38 +6,103 @@ import { Token } from "./tokens";
 export interface TerraformVariableConfig {
     readonly default?: any;
     readonly description?: string;
-    readonly type?: VariableType;
+    readonly type?: IVariableType;
 }
 
-export type VariableType = PrimitiveVariableType | CollectionVariableType | TupleVariableType | ObjectVariableType;
-
-export enum PrimitiveVariableType {
-    STRING = 'string',
-    NUMBER = 'number',
-    BOOL = 'bool',
-    ANY = 'any'
+export interface IVariableType {
+    /**
+     * @internal
+     */
+    _render(): string;
 }
-const PrimitiveTypeVals = ['string', 'number', 'bool', 'any'] as const;
+
+export class PrimitiveVariableType implements IVariableType {
+    public static readonly STRING = new PrimitiveVariableType('string');
+    public static readonly NUMBER = new PrimitiveVariableType('number');
+    public static readonly BOOL = new PrimitiveVariableType('bool');
+    public static readonly ANY = new PrimitiveVariableType('any');
+
+    constructor(private readonly type: string) {
+
+    }
+
+    /**
+     * @internal
+     */
+    _render(): string {
+        return this.type;
+    }
+}
 
 export enum CollectionType {
     LIST = 'list',
     MAP = 'map',
     SET = 'set'
 }
-const CollectionTypeVals = ['list', 'map', 'set'] as const;
 
-export type CollectionVariableType = [CollectionType, VariableType];
+export class CollectionVariableType implements IVariableType {
 
-export type TupleVariableType = VariableType[];
+    public static readonly LIST = new CollectionVariableType(CollectionType.LIST);
+    public static readonly MAP = new CollectionVariableType(CollectionType.MAP);
+    public static readonly SET = new CollectionVariableType(CollectionType.SET);
 
-export interface ObjectVariableType {
-    [key: string]: VariableType;
+    public static readonly LIST_STRING = new CollectionVariableType(CollectionType.LIST, PrimitiveVariableType.STRING);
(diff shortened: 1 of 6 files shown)


jsteinich:
A couple remaining pieces:

1. Generated names might work alright for future module constructs, but they aren't very useful for configuring the stack. May want to share some details with https://github.com/hashicorp/terraform-cdk/issues/247 though maybe flip the default behavior.
2. Tokens aren't really able to represent all of the variable types. Expanding them is maybe a separate task though.

packages/cdktf/test/variable.test.ts
```diff
@@ -0,0 +1,107 @@
...
+    });
+    expect(Testing.synth(stack)).toMatchSnapshot();
+});
+
+test('collection type', () => {
+    const app = Testing.app();
+    const stack = new TerraformStack(app, 'test');
+
+    new TerraformVariable(stack, 'test-variable', {
+        type: new TerraformCollectionVariableType(TerraformCollectionType.LIST, TerraformPrimitiveVariableType.STRING)
```
skorfmann, line 69: That's too much code to define `list(string)` for my taste. I'm wondering if we could simplify this?

Ignoring the `tuple` / `object` type for a moment, this could probably be one enum (e.g. `VariableType.LIST_STRING`, `VariableType.STRING`) - or am I missing something?

- For `tuple`, would this work? `[VariableType.BOOL, VariableType.LIST_NUMBER]
- For `object`, would this work? `{internal: VariableType.NUMBER, protocol: VariableType.STRING]

Then `type` on `TerraformVariable` would have to accept something like `VariableType | VariableType[] | {[key: string]: VariableType | VariableType[] }` (sketched without testing, not sure if that actually works)

jsteinich: I'll see if I can cut down on the verbosity, especially for common cases. More complex types may still end up being verbose.

One of the difficulties is that jsii can't handle everything that Typescript can do and resorting to any results in poor experience in other languages especially.

jsteinich: > Then `type` on `TerraformVariable` would have to accept something like `VariableType | VariableType[] | {[key: string]: VariableType | VariableType[] }`

I went back and testing union types again and they work better than I remembered.

- Something like that of course works in TS.
- It also seems fine in JS and Python.
- Even Java works pretty well since an overloaded builder method for each type is generated.
- C# is pretty terrible. The result is just an `object`.

Ultimately I want to be writing in C# which is why I went with the implementation that I did; however, I'm not sure that justifies the added verbosity in other languages.

There is hope of [jsii generating more useful union code](https://github.com/aws/aws-cdk-rfcs/blob/feature/remove-type-unions/text/0193-fixing-type-unions.md), but I'm not sure how far out that is. @eladb any more context on that?

skorfmann: > I went back and testing union types again and they work better than I remembered.

Thanks for testing this. 

> Ultimately I want to be writing in C# which is why I went with the implementation that I did; however, I'm not sure that justifies the added verbosity in other languages.

Right, I see. Personally, I don't think it'd justify the additional verbosity. In particular when it looks like there's a  [pathway](https://github.com/aws/aws-cdk-rfcs/blob/feature/remove-type-unions/text/0193-fixing-type-unions.md) to a better experience in the future. 

The next step for language support will be Java (which comes more or less for free with #276). C# will have to be added to `jsii-srcmak`.

jsteinich:
I experimented quite a bit with less verbose ways to define the variable type. Commits https://github.com/hashicorp/terraform-cdk/pull/277/commits/80389e3b568972de1b894c25b4af0e0ab1db392a and https://github.com/hashicorp/terraform-cdk/pull/277/commits/56a4ebc32b5b97f5e05b132ec6205c438db335e5 show a couple iterations on it.
Ultimately I ended up back at my original version with a few tweaks:

- Dropped `Terraform` from many of the names to make them shorter
- Added 2 ways for shorter collections
1. Static constants for primitive element types
2. An `of` method that when coupled with static base isn't really shorter, but somehow looks nicer to me

Some of the issues:

- Collections can actually contain any other type, so it isn't possible to enumerate all of them
- I tried a tuple type for defining a collection which looked alright in TS (probably better in TS 4), but completely failed in Java
- I tried defining the tuple type as an array, but at that point it is a recursive type. Recursive types sometimes work, but it caused tsc to spin idefinitely
- I tried object types as an index type which works fine in TS, but is unusable in Java
- Trying to mix enums/interfaces/type aliases made it rather difficult to get the synthesizing logic correct, while it is quite simple with classes

Someone with more Typescript experience might be able to get something cleaner, but with the limitations of jsii and a desire to not make a terrible experience in other languages, I think this is a good balance.

jsteinich:
I made variables default to using the `id` as the name. I imagine the common use case being defining root input variables, so that's why I went that route.
I did make it possible to use generated names by setting `staticName` to `false`.

jsteinich:
I think any token expansion should be done as a separate task. I'm not very familiar with how they work under the hood and I'm also not sure how best to represent generic types without generics.

skorfmann:
> limitations of jsii and a desire to not make a terrible experience in other languages, I think this is a good balance.

Could you give an example for the "terrible experience" in other languages which you're trying to avoid here?

jsteinich:
> Could you give an example for the "terrible experience" in other languages which you're trying to avoid here?

```typescript
export interface ObjectVariableType {
    [key: string]: VariableType;
}
```

translates to
```java
package org.terrastack;

/**
 * EXPERIMENTAL
 */
@javax.annotation.Generated(value = "jsii-pacmak/1.7.0 (build 179a3a5)", date = "2020-08-10T12:36:14.736Z")
@software.amazon.jsii.Jsii(module = org.terrastack.$Module.class, fqn = "@terrastack/core.ObjectVariableType")
@software.amazon.jsii.Jsii.Proxy(ObjectVariableType.Jsii$Proxy.class)
@software.amazon.jsii.Stability(software.amazon.jsii.Stability.Level.Experimental)
public interface ObjectVariableType extends software.amazon.jsii.JsiiSerializable {

    /**
     * @return a {@link Builder} of {@link ObjectVariableType}
     */
    @software.amazon.jsii.Stability(software.amazon.jsii.Stability.Level.Experimental)
    static Builder builder() {
        return new Builder();
    }
    /**
     * A builder for {@link ObjectVariableType}
     */
    @software.amazon.jsii.Stability(software.amazon.jsii.Stability.Level.Experimental)
    public static final class Builder implements software.amazon.jsii.Builder<ObjectVariableType> {

        /**
         * Builds the configured instance.
         * @return a new instance of {@link ObjectVariableType}
         * @throws NullPointerException if any required attribute was not provided
         */
        @software.amazon.jsii.Stability(software.amazon.jsii.Stability.Level.Experimental)
        @Override
        public ObjectVariableType build() {
            return new Jsii$Proxy();
        }
    }

    /**
     * An implementation for {@link ObjectVariableType}
     */
    @software.amazon.jsii.Stability(software.amazon.jsii.Stability.Level.Experimental)
    final class Jsii$Proxy extends software.amazon.jsii.JsiiObject implements ObjectVariableType {

        /**
         * Constructor that initializes the object based on values retrieved from the JsiiObject.
         * @param objRef Reference to the JSII managed object.
         */
        protected Jsii$Proxy(final software.amazon.jsii.JsiiObjectRef objRef) {
            super(objRef);
        }

        /**
         * Constructor that initializes the object based on literal property values passed by the {@link Builder}.
         */
        private Jsii$Proxy() {
            super(software.amazon.jsii.JsiiObject.InitializationMode.JSII);
        }

        @Override
        public com.fasterxml.jackson.databind.JsonNode $jsii$toJson() {
            final com.fasterxml.jackson.databind.ObjectMapper om = software.amazon.jsii.JsiiObjectMapper.INSTANCE;
            final com.fasterxml.jackson.databind.node.ObjectNode data = com.fasterxml.jackson.databind.node.JsonNodeFactory.instance.objectNode();


            final com.fasterxml.jackson.databind.node.ObjectNode struct = com.fasterxml.jackson.databind.node.JsonNodeFactory.instance.objectNode();
            struct.set("fqn", om.valueToTree("@terrastack/core.ObjectVariableType"));
            struct.set("data", data);

            final com.fasterxml.jackson.databind.node.ObjectNode obj = com.fasterxml.jackson.databind.node.JsonNodeFactory.instance.objectNode();
            obj.set("$jsii.struct", struct);

            return obj;
        }
    }
}

```

jsteinich:
I poked around a bit more and I may have found a way to get `tuples` to work as just an array, but I'm still at a loss for how to get `objects` working as desired in TS while still being usable in Java/C#.

jsteinich:
I'll probably make some adjustments to the id/name assigned to variables once #329 is resolved, but any other next steps @skorfmann ?

cmclaughlin:
Would this be a good place to add support for CLI vars?

i.e. like the Terraform CLI has support for `-var 'foo=bar'` and `-var-file=foo`

Or perhaps we should open a new issue for that?

skorfmann:
> I'll probably make some adjustments to the id/name assigned to variables once #329 is resolved, but any other next steps @skorfmann ?

I'll have a look at this soon. Gonna try it out locally

packages/cdktf/lib/terraform-variable.ts
```diff
@@ -0,0 +1,170 @@
...
+     * @internal
+     */
+    _render(): string;
+}
+
+export class PrimitiveVariableType implements IVariableType {
+    public static readonly STRING = new PrimitiveVariableType('string');
+    public static readonly NUMBER = new PrimitiveVariableType('number');
+    public static readonly BOOL = new PrimitiveVariableType('bool');
+    public static readonly ANY = new PrimitiveVariableType('any');
```
skorfmann, line 17: Could this be an alternative way to express this? 

```ts
type PrimitiveVariableType = "string" | "number" | "bool" | "any"
```

jsteinich: I can double check, but I don't believe jsii supports that. 

It also makes the synth a little more complicated since it needs a special case (I had a similar version with an enum).

skorfmann:
My feeling is still, that we shouldn't make it more complicated than the underlying implementation. 

Perhaps taking a step back will help us here. What is the actual goal of having typed variable types? Is it just for the sake of auto completion, or are we're actually using them for validation? 

Let's take a look at two examples:

### 1.

```
// rather than
const imageId = new TerraformVariable(this, 'imageId', {
    type: PrimitiveVariableType.STRING,
    default: 'ami-abcde123',
    description: 'What AMI to use to create an instance'
});

// which is equal to

const imageId = new TerraformVariable(this, 'imageId', {
    type: 'string',
    default: 'ami-abcde123',
    description: 'What AMI to use to create an instance'
});
```

### 2.
Or a more complex example to demonstrate my pain points:

```ts
// rather than
    new TerraformVariable(stack, 'test-variable', {
        type: new CollectionVariableType(CollectionType.MAP, PrimitiveVariableType.BOOL)
    });

// which is equal to

    new TerraformVariable(stack, 'test-variable', {
        type: "map(bool)"
    });
```

A user would have to know about and understand how to put together the following types:

- `CollectionVariableType` 
- `CollectionType` 
- `PrimitiveVariableType` 

At that point, I would think it's way easier for the user to supply a plain HCL compatible string. In particular, when considering that Terraform Variables are not a feature which I would expect a common user to use frequently. I think that's more a niche feature for creating modules and migrating a HCL project to cdktf. For that case, it's actually way easier to go with a plain string as well.

Therefore, I'd suggest to drop the typed Variable types and replace it by inline documentation for now (so that it shows up in intellisense. 

Anything I missed and / or other thoughts?

packages/cdktf/lib/terraform-variable.ts
skorfmann: See other comment https://github.com/hashicorp/terraform-cdk/pull/277#issuecomment-685965999

jsteinich:
> I think that's more a niche feature

I want to believe this, but I'm not quite convinced it won't be used a bit more than that.



> Anything I missed and / or other thoughts?

I hold strongly to not relying on "magic" strings. I pushed what I think is a reasonable compromise though. You can pass the raw string for the hcl type or you can use constants / factory methods on an abstract (static) class to get valid values.

jsteinich:
> Could you provide a few use-cases you have in mind?

Mainly thinking about a synth once, deploy multiple workflow.

We currently use https://www.runatlantis.io/docs/custom-workflows.html#use-cases with multiple tfvar files. It would be pretty straightforward to add a synth to every workflow, but I'm also exploring running the synth separately and keeping the atlantis workflow as is.

skorfmann:
> We currently use https://www.runatlantis.io/docs/custom-workflows.html#use-cases with multiple tfvar files. It would be pretty straightforward to add a synth to every workflow, but I'm also exploring running the synth separately and keeping the atlantis workflow as is.

Hm, that's interesting - are you using workspaces as well?

jsteinich:
> Hm, that's interesting - are you using workspaces as well?

Yes, with S3 backend state.

skorfmann (approved):
Cool, thanks for adding this 👍  Let's merge it and get some real world experience with that :)

PR #278, opened 2020-08-03
FEATURE: Add python example aws eks

smalltown:
Add one python example to demonstrating below CDK for Terraform features:

- TerraformOutput
- Data
- Tokens


examples/python/aws-eks/Pipfile
```diff
@@ -0,0 +1,11 @@
...
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[requires]
+python_version = "3.7"
+
+[packages]
+constructs="*"
+cdktf="*"
```
skorfmann, line 11: Please use the relative reference here `cdktf = {path = "./../../../dist/python/cdktf-0.0.0-py3-none-any.whl"}`.

examples/python/aws-eks/Pipfile
```diff
@@ -0,0 +1,11 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[requires]
+python_version = "3.7"
+
+[packages]
+constructs="*"
```
skorfmann, line 10: I think this should be pinned to version 3

examples/python/aws-eks/.gitignore
```diff
@@ -0,0 +1,3 @@
+dist/
+imports/
+terraform.tfstate*
```
skorfmann, line 3: Adding `cdktf.out` would be good, I think.

smalltown:
@skorfmann Thank for your review very much ^^
I have push another commit to fix it！

smalltown:
@skorfmann Thank for your review very much ^^
I have pushed another commit to fix it !!

examples/python/aws-eks/main.py
```diff
@@ -0,0 +1,55 @@
...
+
+class MyStack(TerraformStack):
+    def __init__(self, scope: Construct, ns: str):
+        super().__init__(scope, ns)
+
+        AwsProvider(self, 'Aws', region='us-west-2')
+        LocalProvider(self, 'Local')
+        NullProvider(self, 'Null')
+        RandomProvider(self, 'Random')
+        TemplateProvider(self, 'Template')
```
cmclaughlin, line 26: Perhaps this is more of a Terraform question and less of a cdktf question, but do the last four providers do anything?

skorfmann: Hm, not sure - I think these providers would be pulled in automatically if needed - aren't they?

smalltown: Hi @cmclaughlin and @skorfmann,

Thank for your reviewing,
Yes, the four providers not real take any effect during develop cdk for terraform,
and the provider will be downloaded when user perform `terraform init`,
hence, I have removed the four provider declaration and including from the example.

skorfmann (approved):
Awesome, thanks again 👍

Issue #293, opened 2020-08-07
ckdtf-cli@next with python is broken


skorfmann:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### cdktf & Language Versions

- `cdktf-cli@next` (pre release) 
- Python

### Affected Resource(s)

- `init`

### Debug Output

![Screenshot 2020-08-07 at 16 36 27](https://user-images.githubusercontent.com/136789/89656847-766cef00-d8cc-11ea-8d72-948982664c6d.png)

### Expected Behavior

An `cdktf init --template python` should work when using a pre release of the cli

### Actual Behavior

it's exiting with an error, since we don't publish pre releases for python 

### Steps to Reproduce

- install `npm install -g cdktf-cli@next`
- run `cdktf init --template python --local`

anubhavmishra:
We discussed this and we want to publish the next verbose version to PyPi for the users using the next version of the CLI.

Issue #298, opened 2020-08-11
Warning when using a new version of cdktf with older node modules


pmartindev:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

When executing cdktf synth, cdktf deploy, etc. from the command line, it would be helpful if there was some type of warning saying that the cdktf-cli and cdktf packages are outdated, or a prompt asking if the user would like to update these packages.  Something like

`cdktf synth
WARNING: Terraform CDK is at version 0.0.14, but you are using cdktf-cli 0.0.13 and cdktf 0.0.13 node modules in your current cdktf project.`

skorfmann:
Thanks for creating this issue - that would be pretty good to have!

DanielMSchmidt:
We have this in the CLI now, closing the issue :)

PR #311, opened 2020-08-14
Prepare release 0.0.15

skorfmann:
Since #309 fixed a Python bug and we don't provide pre-release versions in Python (see #293), let's do another release.


skorfmann:
The more releases we're doing, the more urgent this becomes https://github.com/hashicorp/terraform-cdk/issues/298 - Do we have an upgrade documentation somewhere?

skorfmann:
Gonna merge, since the fix is required to build Python packages in #98

PR #314, opened 2020-08-15
Make all Examples build / synth via Lerna

skorfmann:
This allows to run `lerna examples:integration` in the project root dir to check if the examples are working properly. 

Could be easily added as integration test step as well. Thoughts?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,9 +17,35 @@ $ yarn install
 $ yarn build
 ```
 
-This will also build all [Typescript examples](./examples/typescript).
+## Examples
 
-## Development 
+We have a few top level script commands which are executed with Lerna to make the handling of examples easier:
+
+```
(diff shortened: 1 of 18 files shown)


jsteinich:
Looks nice.

It would be nice if there was an example template...maybe `cdktf init --example ...`.

Having it run in ci is nice. Making sure examples build was very helpful when editing the code generator.

skorfmann:
> It would be nice if there was an example template...maybe `cdktf init --example ...`.

What would this example template do?

jsteinich:
> > It would be nice if there was an example template...maybe `cdktf init --example ...`.
> 
> What would this example template do?

It would generate a project with a package.json file (and other files) setup to match the format required for examples.

Currently `cdktf init` produces a project that doesn't quite match requirements for examples and this change makes it farther from correct.

PR #316, opened 2020-08-17
enable Dependabot v2

sullis:
https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"


skorfmann:
Could you provide a description for the intended change, please?

.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
```
skorfmann, line 3: I'm curious why it's configuring github actions only

skorfmann:
I think if we're going to add a [dependabot config](https://docs.github.com/en/github/administering-a-repository/customizing-dependency-updates), we should configure the `npm` package ecosystem as well. Thoughs @sullis ?

skorfmann:
As I mentioned earlier, the biggest pain point are NPM package upgrades, which needs some configuration to ignore critical dependencies.

Just looked what [AWS CDK](https://github.com/aws/aws-cdk/blob/443948164e09aaa81c094c013b32aa1f67b69570/.dependabot/config.yml) is doing, and I think we could likely adapt their configuration for this project. 

@sullis It would be great if you could look into that.

skorfmann:
Closing for now - coming back to this at a later stage.

Issue #325, opened 2020-08-18
Project Metrics


anubhavmishra:
During the experimental period of this project, we would like to collect telemetry reporting such as the cdktf version, language, provider name, etc to help us guide the product features and roadmap. 

We will do this using [Checkpoint](https://checkpoint.hashicorp.com) which is already used by all HashiCorp projects including Terraform that CDK for Terraform depends on to generating and deploying infrastructure code.

No personal data needs to be collected through this telemetry option. Like all projects using Checkpoint, we will have an environmental variable `CDKTF_CHECKPOINT_DISABLE` for opting out of this reporting.

PR #326, opened 2020-08-18
Checkpoint telemetry integration

anubhavmishra:
Integrating [Checkpoint](https://checkpoint.hashicorp.com/) telemetry to report project version, provider name, provider version, etc.

Fixes: #325


packages/cdktf-cli/lib/checkpoint.ts
```diff
@@ -0,0 +1,80 @@
...
+        reportParams.arch = os.arch();
+    }
+
+    if (!reportParams.os) {
+        reportParams.os = os.platform();
+    }
+
+    const postData = JSON.stringify(reportParams);
+
+    await post(`${BASE_URL}telemetry/${reportParams.product}`, postData)
```
skorfmann, line 75: I think this should catch potential errors, so it's not preventing a user from continuing just because there's some issue around tracking.

packages/cdktf-cli/lib/checkpoint.ts
```diff
@@ -0,0 +1,80 @@
+import https = require('https');
+import { format } from 'url';
+import { v4 as uuidv4 } from 'uuid';
+import * as os from 'os';
+
+const BASE_URL = `https://checkpoint-api.hashicorp.com/v1/`;
+
+const VALID_ERROR_CODES = [200, 201];
```
skorfmann, line 8: why is it called error codes?

packages/cdktf-cli/lib/checkpoint.ts
```diff
@@ -0,0 +1,80 @@
...
+
+        req.write(data);
+
+        req.end();
+    })
+}
+
+export async function ReportRequest(reportParams: ReportParams): Promise<void> {
+    // we won't report when checkpoint is disabled.
+    if (process.env.CDKTF_CHECKPOINT_DISABLE == "1") {
```
skorfmann, line 53: should this be set for our tests - so pretty much for all CI jobs?

skorfmann (changes requested):
Besides from the comments, are we going to document this somehow?

packages/cdktf-cli/lib/checkpoint.ts
anubhavmishra: Yes, I think that would help give with generating accurate telemetry.

packages/cdktf-cli/lib/checkpoint.ts
anubhavmishra: thanks! fixed.

packages/cdktf-cli/lib/checkpoint.ts
anubhavmishra: good catch. this should be resolved.

anubhavmishra:
I have documented the telemetry [here](https://github.com/hashicorp/terraform-cdk/pull/326/files#diff-04c6e90faac2675aa89e2176d2eec7d8R70).

skorfmann (commented):
Looks good. 

What do you think about adding one explicit (integration) test for this telemetry stuff?

skorfmann (approved):
As just discussed, just merge after adding the integration test

anubhavmishra:
am seeing errors that aren't related to the PR. I think merging #276 has started causing these errors.

jsteinich:
> I am seeing errors that aren't related to the PR. I think merging #276 has started causing these errors.

I'm looking into it.

PR #329, opened 2020-08-20
Support for matching existing Terraform names

jsteinich:
This provides ways to better support mapping from existing Terraform stacks to cdktf. All elements (resources, data sources, etc) declared as direct children to the `TerraformStack` will use their construct id (what's past in the constructor) as their Terraform name. This is directly inspired from [AWS CDK logical ids](https://github.com/aws/aws-cdk/blob/b5f2f75128e156c04bcd4b4f3513d7387515be1d/packages/%40aws-cdk/core/lib/stack.ts#L775).

The naming restrictions have also been loosened to allow '_' and '-' in logical ids (names). It might make sense to loosen farther since I believe Terraform has even fewer restrictions, but this at least allows common naming patterns.

Since both of those changes are breaking (Terraform resources had different local ids / local names), this also introduces the concept of [feature flags](https://docs.aws.amazon.com/cdk/latest/guide/featureflags.html). This is accomplished through construct runtime context, though support is limited at this time.

Since construct ids need to be unique and Terraform local names don't completely need to be, this also adds `overrideLogicalId` method for complete control over local names. Also, `allocateLogicalId` is a new overridable method for controlling logical id generation for an entire stack.


skorfmann:
Yes! That's a good idea 👍

skorfmann:
That's gonna be a breaking change though - not a big issue, but we should think about way to deal with that.

cmclaughlin:
Not sure if I fully understand this PR, but so far I've been depending on the stack name to uniquely identify resources in the synthesized output.

When running the CLI, I pass in two environment variables which I use in all stack names:

* ENV - dev, qa, prod, etc
* REGION - us-east-1, us-west-2, etc

I was just working on a stack named rds-api (an RDS cluster that our API servers use). So in our dev environment running in us-east-1 the stack name becomes rds-api-dev-us-east-1.

I thought this naming convention was a good idea so I could write my stacks once and synth/plan/apply for many environments in a few regions. If there are any best practices or if this is a bad design on my part, perhaps we should add some docs for recommendations.

So far we only have a few cdktf stacks, so breaking changes are not a big deal. 
I haven't tried importing any of our older raw HCL stacks, but anything to make that easier does sound great.

cmclaughlin:
Just took another look at the code... I think I really like it. I'd keep using my ENV vars to separate my environments in my remote S3 state, but which is probably enough.

jsteinich:
> That's gonna be a breaking change though - not a big issue, but we should think about way to deal with that.

That's a good point. A couple ideas:
* Add an option to maintain existing behavior
* Add in some of the rename capability that CDK has (https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_core.Stack.html#rename-wbr-logical-wbr-idoldid-newid)

I'll push forward with fixing the integration tests and checking that the CLI handles the format.

skorfmann:
> * Add an option to maintain existing behavior

How about something along the lines of this https://docs.aws.amazon.com/cdk/latest/guide/featureflags.html ?

skorfmann:
@cmclaughlin I understand you correctly, that you drive multiple stacks from the same Terraform CDK codebase based on injected envs? That sounds pretty cool - would love to see an example of this :)

> Not sure if I fully understand this PR, but so far I've been depending on the stack name to uniquely identify resources in the synthesized output.

Yes, the stack name would be omitted from the resource name.

jsteinich:
> > * Add an option to maintain existing behavior
> 
> How about something along the lines of this https://docs.aws.amazon.com/cdk/latest/guide/featureflags.html ?

I started porting this over from CDK, but discovered loading in `context` isn't actually something supported at the moment. I could port that over as well, or I could just do something simpler for now. @skorfmann any thoughts?

skorfmann:
> I started porting this over from CDK, but discovered loading in `context` isn't actually something supported at the moment. I could port that over as well, or I could just do something simpler for now.

Any idea already how a simpler approach would look like?

jsteinich:
> Any idea already how a simpler approach would look like?

I haven't really thought about it. Was just thinking reading in one value might be easier, but that might not be true.

I could bypass the file altogether and just set programmatically when the app is created. I don't overly like it, but would be a quick way.

skorfmann:
> I could bypass the file altogether and just set programmatically when the app is created. I don't overly like it, but would be a quick way.

You mean as part of the template?

skorfmann:
I'm not sure if making it part of the template would be less work. Even if the logic would be living in the CLI itself, the template would still have to do search and replace, which becomes more and more complex as well.

What I was thinking of the other day: Perhaps we could change the templates to do actual code generation, rather than search and replace. This might be an opportunity to think about this. But again - I doubt this would be less effort compared to driving feature flags via `cdktf.json`.

Generally speaking, I'm pretty sure feature flags should be global to the application and not based on individual stacks.

Maybe this topic should be an issue / PR on its own?

jsteinich:
> I'm not sure if making it part of the template would be less work

I think it would be very similar to something I've already done getting it into cdktf.json, so shouldn't be too hard.


> Perhaps we could change the templates to do actual code generation

I think overtime that may become necessary. Definitely impacts #339 


> Maybe this topic should be an issue / PR on its own?

I'll work on creating some new issues (and PRs) when I get some free time.

jsteinich:
I've updated the description with some changes made to make this more general purpose and support feature flags.

The python integration tests are currently failing, but if I change `pipenv run...` to `cdktf synth` then they will succeed, The reason for this is because `cdktf.json` is only loaded by the cli, so setting only apply when run that way. I'm pretty sure this matches AWS CDK behavior, but we could change it. (I think the typescript tests that do similar work because of the way environment variables are transferred).

README.md
```diff
@@ -71,6 +71,7 @@ Choose a language:
 * Using Terraform [data sources](./docs/working-with-cdk-for-terraform/data-sources.md).
 * Synthesizing Terraform configuration using CDK for Terraform [synthesize](./docs/working-with-cdk-for-terraform/synthesizing-config.md) command.
 * Project [telemetry](./docs/working-with-cdk-for-terraform/telemetry.md).
+* Handling breaking changes with [feature flags](./docs/working-with-cdk-for-terraform/feature-flags.md).
```
jsteinich, line 75: TODO: add some documentation about id overriding.

packages/cdktf-cli/test/ui/deploy.test.tsx
```diff
@@ -65,7 +65,7 @@ test("Apply", async () => {
   expect(stripAnsi(lastFrame())).toMatchInlineSnapshot(`
     "Deploying Stack: testing
     Resources
-     ⠋ foo.bar_352350
+     ⠋ FOO                  bar_352350          foo.bar_352350
```
jsteinich, line 68: It didn't seem important to maintain this behavior.

packages/cdktf/lib/app.ts
```diff
@@ -49,4 +64,25 @@ export class App extends Construct {
...
+
+    private loadContext(defaults: { [key: string]: string } = { }) {
+        const node = Node.of(this)
+
+        // prime with defaults passed through constructor
+        for (const [k, v] of Object.entries(defaults)) {
+          node.setContext(k, v);
+        }
+
+        //test.app.ts
```
jsteinich, line 76: TODO: remove this line

test/test-typescript-diff/expected/output
```diff
@@ -1,5 +1,5 @@
 Stack: hello-diff
 Resources
- + NULL_RESOURCE        test                null_resource.hellodiff_test_85E428D7
+ + NULL_RESOURCE        hellodiff_test                null_resource.hellodiff_test_85E428D7
```
jsteinich, line 3: I intentionally left this test without the new flags, so ideally this wouldn't change, but it would take some refactoring to get enough information to ResourceName in order to fix this.

README.md
jsteinich: @skorfmann I could see id overriding in a page about migrating from existing Terraform deployments, but I'm not sure we are at a point to have such a page yet. Perhaps I should just hold off on adding to the documentation.

test/test-typescript-diff/expected/output
jsteinich: @skorfmann Any thoughts on this? Basically the cli isn't able to handle pulling the stack name off when not using the new behaviors in the case that there is and underscore or dash in the stack name. We'd need to propagate the feature flag status (or the entire config) all the way to `ResourceName` in order to fix.

jsteinich:
@skorfmann any thoughts regarding https://github.com/hashicorp/terraform-cdk/pull/329#issuecomment-683373347 or anything else with this PR? This is one I'd really like to get in before long.

README.md
skorfmann: Yeah, let's wait with that for a bit 👍

packages/cdktf-cli/test/ui/deploy.test.tsx
skorfmann: 👍

test/test-typescript-diff/expected/output
skorfmann: That's a pure cosmetic UI thing, right?

skorfmann (commented):
Have only skimmed through this, looks good though. To sum up my understanding of what this does:

- Introduces Feature Flags which can be provided via `cdktf.json` or `CDKTF_CONTEXT_JSON`
- Existing apps won't be affected from the introduced changes
- New apps will add the feature flags `EXCLUDE_STACK_ID_FROM_LOGICAL_IDS` and `ALLOW_SEP_CHARS_IN_LOGICAL_IDS` to `cdktf.json` on `init`
- This will have the following effects 
  - Stack name will be omitted from resource names
  - `-` and `_` won't be removed from resources names
  - No unique id will be added at the end of root level resource names
- `overrideLogicalId` can be used to set a fixed resource name

Is that correct, did I miss anything?

Gonna give it a try locally.

skorfmann:
Just tested it locally with and without feature flags. Works as expected, I think.

Screenshot 2020-09-07 at 13 02 53

test/test-typescript-diff/expected/output
jsteinich: Yeah. Technically someone could be parsing the output, but I wouldn't expect that.

skorfmann: Yeah, I think that's not an issue then.

jsteinich:
> Is that correct, did I miss anything?

Feature flags can only be provided in `cdktf.json`. The environment variable is just used to transport the data between the cli and running the app.

There's also `allocateLogicalId`, which is a new overridable method for controlling logical id generation for an entire stack.

> @jsteinich have you seen this comment btw? [#248 (comment)](https://github.com/hashicorp/terraform-cdk/issues/248#issuecomment-678351919)

That's why I added `overrideLogicalId` at this time. Ultimately the resource type is part of the synthesized json, it's just constructs that require uniqueness on the ids.
We could could change the code generation / some fixed types to prepend to the construct id, but it seems a bit intrusive to me.

@skorfmann did you see my [question](https://github.com/hashicorp/terraform-cdk/pull/329#issuecomment-683373347) about either changing integration tests or when `cdktf.json` is loaded?

skorfmann:
> The python integration tests are currently failing, but if I change `pipenv run...` to `cdktf synth` then they will succeed, The reason for this is because `cdktf.json` is only loaded by the cli, so setting only apply when run that way. I'm pretty sure this matches AWS CDK behavior, but we could change it. (I think the typescript tests that do similar work because of the way environment variables are transferred).

yes, I think it's ok to assume that synth happens trough the CLI. If we'll see bug reports around this, we still could implement some check to produce a warning if the CLI isn't involved in synth. 

> > @jsteinich have you seen this comment btw? [#248 (comment)](https://github.com/hashicorp/terraform-cdk/issues/248#issuecomment-678351919)
> 
> That's why I added `overrideLogicalId` at this time. Ultimately the resource type is part of the synthesized json, it's just constructs that require uniqueness on the ids.
> We could could change the code generation / some fixed types to prepend to the construct id, but it seems a bit intrusive to me.

Yeah, I was thinking about the same. Probably not worthwhile doing 👍

packages/cdktf-cli/bin/cdktf.ts
```diff
@@ -32,6 +32,7 @@ const args = yargs
   .option('disable-logging', { type: 'boolean', default: true, required: false, desc: 'Dont write log files. Supported using the env CDKTF_DISABLE_LOGGING.'})
   .option('disable-plugin-cache-env', { type: 'boolean', default: false, required: false, desc: 'Dont set TF_PLUGIN_CACHE_DIR automatically. This is useful when the plugin cache is configured differently. Supported using the env CDKTF_DISABLE_PLUGIN_CACHE_ENV.'})
   .option('log-level', { type: 'string', required: false, desc: 'Which log level should be written. Only supported via setting the env CDKTF_LOG_LEVEL'})
+  .option('contextJson', { required: false, hidden: true, desc: 'Used internally for env variable' })
```
skorfmann, line 35:
```suggestion
  .option('context-json', { required: false, hidden: true, desc: 'Used internally for env variable' })
```

skorfmann (commented):
Cool, looks good. I think in general it's good to go, right? @jsteinich

jsteinich:
> Cool, looks good. I think in general it's good to go, right? @jsteinich

Yeah, just accepted your change. Assuming tests pass (they did locally), then I think everything is ready.

david92rl:
This was a blocker for us. Well done 🎉 
Do you have any plans to release a new version including this change soon?
Also, is there any documentation explaining the usage of `overrideLogicalId` and `allocateLogicalId`

Thanks a lot

jsteinich:
> Do you have any plans to release a new version including this change soon?

If you're using ts/js it should already be available using the [@next version](https://learn.hashicorp.com/tutorials/terraform/cdktf-install#install-cdktf). I'm not sure when the next stable version will be released.



> Also, is there any documentation explaining the usage of `overrideLogicalId` and `allocateLogicalId`

`overrideLogicalId` should result in the resource having exactly that terraform name.
[The CDK documentation for allocateLogicalId](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_core.Stack.html#protected-allocate-wbr-logical-wbr-idcfnelement) pretty much matches how it is used here.

david92rl:
Thank you @jsteinich 
We're using Python, but it's fine. I can use the built version of master until it gets released 👍

PR #331, opened 2020-08-20
Streamline Usage of Prebuilt Providers

### LINKED ISSUE #98: Prebuilt Provider Packages

### LINKED ISSUE #275: Document usage of cdktf.json file
We need to document the usage of `cdktf.json` file to enable users to better understand the usage and modify it as needed by them to support customizations for various languages.

### LINKED ISSUE #284: Streamline Workflow with Prebuilt Providers
### Community Note

### Description

When using a pre-built provider (early example [here](https://github.com/skorfmann/cdktf-provider-aws)) there are a few things we can streamline:

- `cdktf init` should take a flag to skip `get` - something like `cdktf init --skip-get`
- `cdktf synth` expects the imports folder (`.gen` by default) to be present. Perhaps we can skip this entirely, or we skip that in case no providers are specified in `cdktf.json`
- `cdktf.json` should allow the `providers` field to be omitted

### References

- #98

### LINKED ISSUE #308: Create Minimal Template
### Community Note

### Description

It would be great to have a template for the `cdktf init` command, which doesn't pull in any provider and therefore doesn't generate any provider bindings. This is particular useful for our integration tests, but I can imagine that this would be interesting to regular users as well.

### LINKED ISSUE #320: Document usage of provider packages
Add a section in the following document about using provider packages: https://github.com/hashicorp/terraform-cdk/blob/master/docs/working-with-cdk-for-terraform/importing-providers-and-modules.md

- Add package in the `package.json` file

skorfmann:
This streamlines the experience of using [prebuilt Terraform providers](https://cdk.tf/provider):

In `cdktf.json`:  When no `terraformProviders` and `terraformModules` are set, the check for the presence of `codeMakerOutput` is omitted. This allows running the check, even when using a prebuilt provider alongside custom generated providers or modules.

In addition, this pull request, introduces a minimal template (see #308). This is a good starter when using prebuilt providers.

Last but not least, it adds a bit of documentation. 

Fixes #98 
Fixes #284 
Fixes #275 
Fixes #308  
Fixes #320


docs/working-with-cdk-for-terraform/cdktf-json.md
```diff
@@ -0,0 +1,94 @@
...
+
+#### Modules
+
+Similar to 3rd Party providers, the full registry namespace should be provided. Please note that only modules from the registry are supported at this point.
+
+## Examples
+
+### Minimal Configuration
+
+A minimal configuration would define `app` only. This is useful, when planning to use [prebuilt providers]() and therefore no provider or modules bindings should be generated.
```
jsteinich, line 47: Fill in link

docs/working-with-cdk-for-terraform/cdktf-json.md
```diff
@@ -0,0 +1,94 @@
...
+This will synthesize JSON into `my-workdir` and all Terraform operations - such as `deploy` or `destroy` - will be performed in this directory.
+
+```json
+{
+  "app": "npm run --silent compile && node main.js",
+  "output": "my-workdir",
+}
+```
+
+### Building Providers
```
jsteinich, line 66: Might be good to have a module example as well.

docs/working-with-cdk-for-terraform/using-providers-and-modules.md
```diff
@@ -1,4 +1,26 @@
...
+
+To improve the general user experience around provider imports and to allow building further abstractions on top of the Terraform provider bindings, a few popular providers are offered as prebuilt packages. At the moment the following providers are built and published to NPM / PyPi on a regular basis automatically.
+
+- [AWS Provider](https://cdk.tf/provider/aws)
+- [Google Provider](https://cdk.tf/provider/google)
+- [Azure Provider](https://cdk.tf/provider/azurerm)
+- [Kubernetes Provider](https://cdk.tf/provider/kubernetes)
+- [Docker Provider](https://cdk.tf/provider/docker)
+- [Github Provider](https://cdk.tf/provider/github)
+- [Null Provider](https://cdk.tf/provider/null)
```
jsteinich, line 13: I"m curious how this list was chosen. I was expecting all hashicorp official providers.

packages/cdktf-cli/lib/config.ts
```diff
@@ -28,5 +29,7 @@ export function readConfigSync(): Config {
     };
   }
 
+  config.checkCodeMakerOutput = Array.isArray(config.terraformModules) || Array.isArray(config.terraformProviders)
```
jsteinich, line 32: Do you also want to check if non-empty?

docs/working-with-cdk-for-terraform/using-providers-and-modules.md
skorfmann: Mostly popularity, `Github` because I want to build an example case for [Level 2 constructs](https://github.com/skorfmann/cdktf-github-constructs) and `Null` because it's a building block which useful for things like [ecr asset](https://github.com/skorfmann/cdktf-aws-ecr-asset/blob/e60aec5fa9834101819caa62e257281ed3e7b96b/src/index.ts#L25)

The current state is an advanced proof of concept. Mid / long term I think there would be more providers if not all at some point. For now, the goal is collecting feedback and building some examples to see if the UX is going in the right direction.

Things like:

- versioning (version constraints for peer dependencies, how does it work in Python, ...)
- major version changes (e.g. AWS 2.0 -> 3.0) 
- improve the auto publish pipeline 
- building upon these providers
- adding tests to the provider publish pipeline

docs/working-with-cdk-for-terraform/cdktf-json.md
skorfmann: true 👍

packages/cdktf-cli/lib/config.ts
skorfmann: Good point, gonna fix that 👍

Issue #44, opened 2020-05-02
Provider Alias not part of Schema Anymore


skorfmann:
In Terrastack (~ a month ago) the provider `alias` attribute was part of the Schema. Now it's missing from the AWS provider schema. Wondering if that change was intentional.

PR #332, opened 2020-08-21
Fix TerraformProvider alias definition

jsteinich:
Match override behavior of generated provider schema

I'm not sure yet how this passed tests in #276, but this should fix the integration tests.

It is quite strange to me that terraform provider schema include `alias` when it is really more of a meta argument. If anyone from the terraform team could provide more context, that could make me more confident that this is the correct approach.


skorfmann:
> I'm not sure yet how this passed tests in #276, but this should fix the integration tests.

Yeah, a bit strange. I think I had a similar issue a while ago. 

> It is quite strange to me that terraform provider schema include `alias` when it is really more of a meta argument. If anyone from the terraform team could provide more context, that could make me more confident that this is the correct approach.

Yes, that's a bit weird - See https://github.com/hashicorp/terraform-cdk/issues/44

PR #345, opened 2020-08-24
Pre Release for Python

skorfmann:
Test pre releases for Python

### DIFF
--- a/.github/workflows/release_next.yml
+++ b/.github/workflows/release_next.yml
@@ -6,7 +6,7 @@ on: [pull_request]
   #     - master
 
 jobs:
-  release:
+  build_artifact:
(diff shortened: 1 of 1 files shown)


skorfmann:
Yeah, that's not going to work 

> DVCS based version labels
Many build tools integrate with distributed version control systems like Git and Mercurial in order to add an identifying hash to the version identifier. As hashes cannot be ordered reliably such versions are not permitted in the public version field.

>As with semantic versioning, the public .devN suffix may be used to uniquely identify such releases for publication, while the original DVCS based label can be stored in the project metadata.

>Identifying hash information may also be included in local version labels.

See https://www.python.org/dev/peps/pep-0440/

PR #353, opened 2020-08-28
Add typescript example  Azure App Service 😊

guan840912:
Add one typescript azure  example to demonstrating below CDK for Terraform features:

Azure App Service

### DIFF
--- a/examples/typescript/azure-app-service/package.json
+++ b/examples/typescript/azure-app-service/package.json
@@ -1,0 +1,30 @@
+{
+  "name": "azure-app-service-docker",
+  "version": "0.0.0",
+  "main": "main.js",
+  "types": "main.ts",
+  "license": "MPL-2.0",
+  "private": true,
+  "scripts": {
+    "get": "cdktf get",
+    "build": "yarn get && tsc",
+    "synth": "cdktf synth",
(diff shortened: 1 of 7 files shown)


examples/typescript/azure-app-service/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "azure-app-service-docker",
```
skorfmann, line 2: please name it `@examples/azure-app-service-docker` - see https://github.com/hashicorp/terraform-cdk/blob/7faebbe7ace248377c3ca35c7d40697e797f14e2/CONTRIBUTING.md#examples

skorfmann (changes requested):
Cool - thanks for the example @guan840912 🙌 

Please see the comment for the name change, other than that it's good to go 👍

examples/typescript/azure-app-service/package.json
guan840912: No problem !!! Sorry my mistake 😊

skorfmann: No worries, that's something which got introduced very recently

skorfmann (approved):

PR #357, opened 2020-09-01
Simple Terraform Locals

### LINKED ISSUE #281: Add Terraform Locals
### Community Note

### Description

As requested in https://github.com/hashicorp/terraform-cdk/issues/236 - Supported for Terraform locals might be useful. In particular for computations at deploy time.

### References

https://github.com/hashicorp/terraform-cdk/issues/236

jsteinich:
Closes #281 

Very simple terraform local which is just a bit more than an escape hatch at this point. We'll probably expand a bit down the road, but a local is still likely to be a bit of a wrapper. That said, it is part of hcl, so seems like something worth having for completeness.

I'll fill in tests/documentation if this seems worthwhile to others.

### DIFF
--- a/packages/cdktf/lib/terraform-local.ts
+++ b/packages/cdktf/lib/terraform-local.ts
@@ -1,0 +1,49 @@
+import { Construct } from "constructs";
+import { TerraformElement } from "./terraform-element";
+import { Token } from "./tokens";
+
+export class TerraformLocal extends TerraformElement {
+    private _expression: any;
+
+    constructor(scope: Construct, id: string, expression: any) {
+        super(scope, id);
+
+        this._expression = expression;
+    }
+
+    public set expression(value: any) {
+        this._expression = value;
+    }
+
+    public get expression() {
+        return Token.asAny(this.interpolation());
+    }
+
+    public get asString(): string {
+        return Token.asString(this.interpolation());
+    }
+
+    public get asNumber(): number {
+        return Token.asNumber(this.interpolation());
+    }
+
+    public get asList(): string[] {
+        return Token.asList(this.interpolation());
+    }
+
+    public get asBoolean(): boolean {
+        return Token.asString(this.interpolation()) as any as boolean
+    }
+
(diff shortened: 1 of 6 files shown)


skorfmann:
> I'll fill in tests/documentation if this seems worthwhile to others.

Yes, please!

skorfmann:
@jsteinich thanks - looks good 👍

packages/cdktf/lib/terraform-local.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export class TerraformLocal extends TerraformElement {
+    private _expression: any;
+
+    constructor(scope: Construct, id: string, expression: any) {
+        super(scope, id);
+
+        this._expression = expression;
+    }
+
```
skorfmann, line 13: I think it would make sense to add the following:

## Typescript 

```ts
    public toString(): string {
        return this.asString
    }
```

This would allow to use it like `my-first-table-${local}` rather than `my-first-table-${local.asString}` in Typescript. 

## Python

I did try it for Python but it doesn't work: `"my-first-sns-topic-<cdktf.TerraformLocal object at 0x10cd45190>"`

Defining the required method for Python in the Typescript class only works as `@internal` and therefore wouldn't be compiled via jsii.

```ts
 /** @internal */
    public __str__(): string {
        return this.asString
    }
```

There's an issue somehow related to this at Jsii: https://github.com/aws/jsii/issues/380

Other ideas?

jsteinich: I agree it would be nice to directly reference the local value. I don't really think `toString` is the answer since I would expect users to have already set the local to the value they desire and `${local}` seems about the same as `local.asString` to me.

We could make properties accept and `IResolvable` interface in addition to their normal type, but I think we'll run into problems with some other jsii languages.

skorfmann: Ok, let's leave it like this for now and come back to this later 👍

Issue #358, opened 2020-09-04
Allow to specify terraform binary via environment variable


dethmix:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

It is very useful to be able to set path to terraform binary if you use multiple terraform versions. 

You have already implemented this in https://github.com/hashicorp/terraform-cdk/blob/a2387f810d9ca7e4cdcfadbbecab6b43a192f6ee/packages/cdktf-cli/bin/cmds/ui/models/terraform.ts#L4

but still use hardcoded `terraform` as path to binary at 
https://github.com/hashicorp/terraform-cdk/blob/a2387f810d9ca7e4cdcfadbbecab6b43a192f6ee/packages/cdktf-cli/lib/get/generator/provider-schema.ts#L80
#

So as a result we start with `terraform_13` and finish with default `terraform`
```
CDKTF_LOG_LEVEL=debug TERRAFORM_BINARY_NAME="/usr/local/bin/terraform_13" cdktf get
[2020-09-03T23:28:37.761] [DEBUG] default - Terraform v0.13.2

⠙ downloading and generating providers...
[2020-09-03T23:28:37.973] [DEBUG] default -

[2020-09-03T23:28:37.974] [DEBUG] default - Initializing provider plugins...
- Checking for available provider plugins on https://releases.hashicorp.com...
⠇ downloading and generating providers...
[2020-09-03T23:28:38.520] [DEBUG] default - - Downloading plugin for provider "aws" (2.70.0)...
⠼ downloading and generating providers...
[2020-09-03T23:28:39.008] [DEBUG] default -

[2020-09-03T23:28:39.008] [DEBUG] default - The following providers do not have any version constraints in configuration,
so the latest version was installed.

[2020-09-03T23:28:39.008] [DEBUG] default -

[2020-09-03T23:28:39.008] [DEBUG] default - To prevent automatic upgrades to new major versions that may contain breaking

[2020-09-03T23:28:39.008] [DEBUG] default - changes, it is recommended to add version = "..." constraints to the

[...]

PR #366, opened 2020-09-08
allow to use TERRAFORM_BINARY_NAME env to set terraform binary

dethmix:
PR to fix issue https://github.com/hashicorp/terraform-cdk/issues/358


jsteinich:
I think the test failure is actually another missed case: https://github.com/hashicorp/terraform-cdk/blob/5f400e133912ddfc472f03e32ed81266ce7fe447/.github/workflows/examples.yml#L28
That internally uses Terraform, but the environment variable isn't set for that step.

skorfmann:
Cool - thank you @dethmix 👍

skorfmann (approved):

Issue #16, opened 2020-04-20
Extend Support of Module Sources beyond Terraform Registry


joatmon08:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

The CDK for Terraform only supports generation of publicly available modules hosted in [registry.terraform.io](https://registry.terraform.io/v1/modules). In order to support private Terraform registries and other module sources, we will need to extend the CDK for Terraform to obtain module metadata to private or version-controlled sources.

### References

N/A

tovbinm:
Adding local / custom modules (e.g. using Git submodules) would also be great.

skorfmann:
This will be relatively easy when we have #232 

> Adding local / custom modules (e.g. using Git submodules) would also be great.

You mean local, as in not available on Github / Gitlab / etc?

tovbinm:
Yeah, unpublished, say from an internal repo or a subfolder.

skorfmann:
> Yeah, unpublished, say from an internal repo or a subfolder.

That makes sense, yes. I think that's not much different from a git remote. If one works, the other one will be easy to do as well.

jsteinich:
Would something like `TerraformHclModule` used similar to https://github.com/jsteinich/terrastack/pull/5/files#diff-1526cfd8651d8210fb4a0decf140147aR70 be useful to anyone else while waiting for generation to be supported?

skorfmann:
> Would something like `TerraformHclModule` used similar to https://github.com/jsteinich/terrastack/pull/5/files#diff-1526cfd8651d8210fb4a0decf140147aR70 be useful to anyone else while waiting for generation to be supported?

It's also possible to go with [escape hatches](https://cdk.tf/escape-hatch). Not quite as comfortable, but still possible :)

cmclaughlin:
@jsteinich @skorfmann 

Does PR # 375 fix this?

No detailed explanation needed - I was just looking through issues and PRs and thought we might be able to close this one.

skorfmann:
> Does PR #375 fix this?

[...]

PR #375, opened 2020-09-18
Terraform hcl module

jsteinich:
Helps with #16 

This is meant to be something a bit nicer than escape hatches for referencing non-registry modules. Eventually this may be come obsolete if code generation supports non-registry modules, but that seems like a ways off yet.

I didn't add the new Terraform 0.13 meta arguments. I think that's better suited for a follow up especially since there is a bunch of overlap between different types.

### DIFF
--- a/packages/cdktf/lib/terraform-hcl-module.ts
+++ b/packages/cdktf/lib/terraform-hcl-module.ts
@@ -1,25 +1,16 @@
 import { Construct } from "constructs";
 import { TerraformModuleOptions, TerraformModule } from "./terraform-module";
-import { TerraformProvider } from "./terraform-provider";
 
 export interface TerraformHclModuleOptions extends TerraformModuleOptions {
-    readonly providers?: (TerraformProvider | TerraformModuleProvider)[];
     readonly variables?: {[key: string]: any};
 }
 
-export interface TerraformModuleProvider {
-    readonly provider: TerraformProvider;
-    readonly moduleAlias: string;
-}
-
 export class TerraformHclModule extends TerraformModule {
-    private _providers?: (TerraformProvider | TerraformModuleProvider)[];
     private _variables?: {[key: string]: any};
 
     constructor(scope: Construct, id: string, options: TerraformHclModuleOptions) {
         super(scope, id, options);
 
-        this._providers = options.providers;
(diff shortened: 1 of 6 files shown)


packages/cdktf/lib/terraform-module.ts
```diff
@@ -4,13 +4,13 @@ import { deepMerge } from "./util";
 
 export interface TerraformModuleOptions {
   readonly source: string;
-  readonly version: string;
+  readonly version?: string;
```
jsteinich, line 7: This isn't a required property for using modules.

packages/cdktf/lib/terraform-hcl-module.ts
```diff
@@ -0,0 +1,81 @@
+import { Construct } from "constructs";
+import { TerraformModuleOptions, TerraformModule } from "./terraform-module";
+import { TerraformProvider } from "./terraform-provider";
+
+export interface TerraformHclModuleOptions extends TerraformModuleOptions {
+    readonly providers?: (TerraformProvider | TerraformModuleProvider)[];
+    readonly variables?: {[key: string]: any};
```
jsteinich, line 7: It's a bit unfortunate to name this, but it won't translate well to java/c# if left as an indexer.

packages/cdktf/lib/terraform-hcl-module.ts
```diff
@@ -0,0 +1,81 @@
...
+}
+
+export interface TerraformModuleProvider {
```
jsteinich, line 10: I'm not completely sold on this interface, but it implementation a bit easier while still not forcing a module alias.

packages/cdktf/lib/terraform-module.ts
skorfmann: 👍

skorfmann (commented):
Cool - This allows the full module functionality including remote / git modules, I suppose?

jsteinich:
> Cool - This allows the full module functionality including remote / git modules, I suppose?

Yeah. It's basically just as if you were writing hcl for using a module. I think we should keep #16 open since this is kinda like writing hcl without an IDE plugin.

skorfmann (approved):
👍

Issue #216, opened 2020-07-20
Add Terraform Version Check


skorfmann:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Reduce confusion to users running an older version of Terraform by printing out a warning notice when Terraform < 0.12 

The functionality could also be useful when new versions of Terraform are being released (e.g. `0.13 ` with the changed provider naming schema).

### References

https://github.com/hashicorp/terraform-cdk/issues/208

bverhoeve:
Hi @skorfmann I'm looking to contribute to the repository, what exactly needs to be done?

skorfmann:
> Hi @skorfmann I'm looking to contribute to the repository, what exactly needs to be done?

For the case where a user uses something below `0.12`, it would be great to display a clear error message and exit the process. The supported Terraform versions at the moment are `0.12` and `0.13`.  I think the check should go like this:

- `< SUPPORTED => exit` with error message
- `>= SUPPORTED => continue` without further message.

We do have a [terraform check](https://github.com/hashicorp/terraform-cdk/blob/175c3872827c1325d32ecc7ea7150a2774da0121/packages/cdktf-cli/bin/cmds/terraform-check.ts#L6) already, which checks for the presence of the Terraform CLI only. This could be expanded to check against supported versions and print a helpful error message.

--- 

ps: I was thinking about if it would make sense to print a warning when a user is using `> SUPPORTED` (e.g. alpha of `0.14` at the moment). However, I think it's not important for now and I'd ignore this case until we actually run into issues with a case like that.

bverhoeve:
Okay, thanks for the explanation! I'll check to tend to it.

skorfmann:
This was addressed in #378

PR #378, opened 2020-09-19
Terraform version check

bverhoeve:
#216 

This is meant to reduce confusion to users running an older version of Terraform by printing out a warning notice when Terraform < 0.12. It could be extended in the future as well by printing a warning when a user is using a version higher than the minimally supported version.

I'm not entirely sure if the minimally supported version should be declared here or somewhere else.


skorfmann:
Awesome - thank you! I'll check it out locally.

packages/cdktf-cli/bin/cmds/terraform-check.ts
```diff
@@ -1,9 +1,21 @@
...
 
 export const terraformCheck = async (): Promise<void> => {
   try {
-    const terraform = new Terraform('./')
-    await terraform.version()
+    const terraform: Terraform = new Terraform('./')
+
+    // string in format of 'Terraform v0.13.3'
+    const terraformVersion = await terraform.version()
+    const cleanTerraformVersion = semver.clean(terraformVersion.substring(terraformVersion.indexOf('v')))
```
skorfmann, line 12: I've tried `0.13.0`, `0.12.29` and `0.11.14`, where I'd have expected the last one to fail. However, it didn't. It looks like `cleanTerraformVersion` is always null.

Some logging for this `terraformVersion`:

```
{
  terraformVersion: 'Terraform v0.11.14\n' +
    '+ provider.aws v2.70.0\n' +
    '\n' +
    'Your version of Terraform is out of date! The latest version\n' +
    'is 0.13.4. You can update by downloading from www.terraform.io/downloads.html\n',
}
```

Looks like `semver.clean` doesn't handle arbritary strings. Perhaps it would make sense to parse `Terraform v<version regex>` explicitly und use the match group as input for the semver parsing. What do you think?

skorfmann (changes requested):
@bverhoeve sorry for the delay. I just tried it locally, and it doesn't seem to work as expected. See inline comment for more details.

packages/cdktf-cli/bin/cmds/terraform-check.ts
bverhoeve: @skorfmann sorry have been swamped at work the previous days. Yes that makes sense since there might be other ways of formatting the version message. I'll check to implement the change.

packages/cdktf-cli/bin/cmds/terraform-check.ts
```diff
@@ -1,9 +1,27 @@
...
-    await terraform.version()
+    const terraform: Terraform = new Terraform('./')
+
+    const terraformVersion = await terraform.version()
+    const terraformVersionMatches = terraformVersion.match(VERSION_REGEXP)
+
+    if ( terraformVersionMatches !== null) {
+
+      // Should always be the first match found in the string
+      let cleanTerraformVersion = semver.clean(terraformVersionMatches[0].substring(terraformVersionMatches[0].indexOf('v')))
```
bverhoeve, line 17: @skorfmann I've explicitly parsed the string so it becomes something like `Terraform v13.0.1` which can then be compared against by semver.

skorfmann (approved):
Sweet - looks good. Tried it locally and works as expected. Thank you!

PR #379, opened 2020-09-21
Fixes CloudFront Proxy Example

poldridge:
AcmCertificate DNS validation must use the DVO's resource_record_name


skorfmann (approved):
cool, thanks for the fix @poldridge  👍

PR #360, opened 2020-09-07
Java support

jsteinich:
Closes #197 

Adds:

- Core library output to Java
- Ability to generate modules/providers in Java
- Java app template with maven
- Basic documentation
- Simple examples
- Setup for release

Setting up the maven keys and profile is probably best left to the internal team.

PR #380, opened 2020-09-22
Increase Maven max memory

jsteinich:
The jsii/superchain image sets `MAVEN_OPTS` with a max heap size that is too small for generating some of the larger providers. I've worked around it in #360, but for https://github.com/terraform-cdk-providers/cdktf-provider-project/pull/14 it will be easier to just have the value increased.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM jsii/superchain
 
-RUN yum install -y unzip jq && curl https://raw.githubusercontent.com/pypa/pipenv/master/get-pipenv.py | python
+RUN yum install -y unzip jq && curl https://raw.githubusercontent.com/pypa/pipenv/master/get-pipenv.py | python3
 
 ENV DEFAULT_TERRAFORM_VERSION=0.13.0                                \
     TF_PLUGIN_CACHE_DIR="/root/.terraform.d/plugin-cache"           \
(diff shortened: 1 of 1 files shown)


jsteinich:
When building the docker image locally I'm getting:
```
Could not import setuptools which is required to install from a source distribution.
Please install setuptools.
```

I'll try to look into it more later, but we may want to be a bit more explicit with base image version and pipenv install script version.

jsteinich:
Switching to python3 for the pip install fixed it.

skorfmann:
We can merge this, but we still don't have automatically docker builds /cc @anubhavmishra

skorfmann (approved):

Dockerfile
```diff
@@ -1,9 +1,10 @@
 FROM jsii/superchain
 
-RUN yum install -y unzip jq && curl https://raw.githubusercontent.com/pypa/pipenv/master/get-pipenv.py | python
+RUN yum install -y unzip jq && curl https://raw.githubusercontent.com/pypa/pipenv/master/get-pipenv.py | python3
 
-ENV DEFAULT_TERRAFORM_VERSION=0.13.0
-ENV TF_PLUGIN_CACHE_DIR="/root/.terraform.d/plugin-cache"
+ENV DEFAULT_TERRAFORM_VERSION=0.13.0                                \
+    TF_PLUGIN_CACHE_DIR="/root/.terraform.d/plugin-cache"           \
+    MAVEN_OPTS="-Xms256m -Xmx3G"
```
anubhavmishra, line 7: @jsteinich can add a comment about the maven option being set?

anubhavmishra (changes requested):

anubhavmishra (approved):

skorfmann:
Merging, will be published manually or via #396

PR #417, opened 2020-10-31
Use docker mirror

mdeggies:
CHANGELOG: no impact

Dockerhub is going to rate limit unauthenticated pulls on November 1st. IPs from CI machines will be near their limit all the time. We're moving projects to use a public un-rate-limited Mirror of these images instead. Let me know if you have any q's, otherwise feel free to merge when you can!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM jsii/superchain
+FROM hashicorp.jfrog.io/docker/jsii/superchain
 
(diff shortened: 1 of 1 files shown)


anubhavmishra:
thanks, @mdeggies! I hope the move goes over smoothly!

anubhavmishra (commented):
👍 we might want to rebuild and push up a new image.

Issue #382, opened 2020-09-23
Terraform Cloud Support


anubhavmishra:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Background

CDK for Terraform currently only support local operations using Terraform Cloud and uses it as a remote state store. The CDK for Terraform CLI supports commands like `cdktf diff` that runs a `terraform plan` in the background and saves the plan to allow for parsing it to show a simplified diff to the user. This can only be done for Terraform projects that are using local operations.
Currently, the CDK for Terraform CLI can be used to initialize a new project using the `cdktf init` command. The command also gives the user an option to create a new workspace for the project. While creating the workspace, the CDK for Terraform CLI sets up the workspace to local operations only. This approach works well for single user or maybe a smaller team but might not be suitable for large groups of users to collaborate on infrastructure.

There is a need to enable CDK for Terraform to use the full capabilities of Terraform Cloud to enable users to perform remote operations using the CDK for Terraform CLI (cdktf).

### Description

The proposal is that we support Terraform Cloud remote operations using the CDK for Terraform CLI by enabling the cdktf to read details about a Terraform Cloud/Enterprise organization and workspace using the `cdk.tf.json` file generated as part of synthesizing the application and using those details to trigger remote plans and applies. To support generating diffs and run deploy using the `cdktf diff` and `cdktf deploy` commands for remote operations, we also want to propose the CDK for Terraform CLI to be able to retrieve plans from Terraform Cloud/Enterprise APIs. This is due to the limitations of the Terraform CLI that doesn’t allow for saving a Terraform Plan when using remote operations. The [GitHub issue](https://github.com/hashicorp/terraform/issues/22779) highlights the current limitation of Terraform to not allow saving Terraform Plans for projects that have a Terraform Cloud workspace that has remote operations enabled for it.

### UX

**Getting Started**

A getting started workflow would be similar to what already exists but in this case will add a few minor variations to how we create a Terraform workspace and save Terraform Cloud organization and workspace information. For the following command we assume that there is no Terraform Cloud token present in “~/.terraform.d/credentials.tfrc.json” file present.

```bash
[...]

PR #449, opened 2020-11-30
Terraform Cloud Support

skorfmann:
This adds support for Terraform Cloud. It largely follows the workflow described [here](https://www.terraform.io/docs/cloud/run/api.html). It's using a fork of https://github.com/mijailr/terraform-cloud with the goal to upstream changes. 

Screenshot 2020-11-30 at 17 14 58

- [x] Separate CLI / Cloud Executors 
- [x] Detect if Remote Backend is configured for stack
- [x] Read credentials from terraform config 
- [x] Support credentials from env
- [x] Upload synthesized stack to configured Terraform Cloud workspace 
- [x] Plan 
- [x] Deploy
- [x] Destroy 
- [x] Outputs
- [x] ~~Cancel Deploy / Destroy~~ (outscoped to #482)
- [x] Integration test
- [x] ~~Switch Local / Remote State with confirmation if local state exists (mostly moved to #490 - see 462782c)~~

If there are variables necessary, they'll have to be added manually for now. There will be a follow pull request, to make this more integrated (see #485)

Fixes: #382 

Here's the [terraform cloud PR](https://github.com/mijailr/terraform-cloud/pull/114)


skorfmann:
A few notes regarding the current state:

- Right now there aren't streaming logs for deploy / destroy. The cli output is updated when applied. 
- There are likely enterprise edge cases not considered at the moment (sentinel / cost estimation, .. ?)
- There's no handling of interruptions in the CLI. Means, the apply will continue in Terraform Cloud

mijailr:
🎉 Oh interesting! I'm glad to see this!

skorfmann:
> * There's no handling of interruptions in the CLI. Means, the apply will continue in Terraform Cloud

The Terraform CLI does ask if the TF Cloud run should be cancelled. So, that's likely the way to go - However, I'd consider to postpone this to another iteration.

skorfmann:
- [ ] Changes to outputs only needs to be confirmed as well


Screenshot 2020-12-15 at 17 26 34

skorfmann:
> * Changes to outputs only needs to be confirmed as well

Unfortunately, the JSON plan representation doesn't fully cover this. See here https://github.com/hashicorp/terraform/issues/24628

Gonna have to calculate this manually

skorfmann:
@jsteinich / @anubhavmishra feel free to review already. Gonna wrap this up tomorrow and it would be good to have some feedback on it.

packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx
```diff
@@ -330,36 +397,66 @@ export const useTerraform = ({ targetDir, synthCommand }: UseTerraformInput) =>
     return state
```
jsteinich, line 397: Should `planDestroy` also be broken up with a synth check first? I'm not certain the reasons why others were split apart.

packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx
```diff
@@ -330,36 +397,66 @@ export const useTerraform = ({ targetDir, synthCommand }: UseTerraformInput) =>
...
-  const deploy = (plan?: TerraformPlan) => {
+  const deploy = () => {
     React.useEffect(() => {
       const invoke = async () => {
-        if (plan) {
-          await execTerraformApply(plan)
-        } else {
-          throw new Error("expected plan to be present but was undefined")
-        }
+        await execTerraformSynth()
```
jsteinich, line 403: Given how slow synth is, adding this for every deploy seems like it will have a big impact.

packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx
```diff
@@ -330,36 +397,66 @@ export const useTerraform = ({ targetDir, synthCommand }: UseTerraformInput) =>
...
+      }
+      invoke()
+    }, [])
+
+    React.useEffect(() => {
+      const invoke = async () => {
+        await execTerraformInit()
+        await execTerraformPlan(false)
+      }
+      if (state.status === Status.SYNTHESIZED) invoke();
```
jsteinich, line 413: Am I reading correctly that those won't be ran if a plan has already been created?

packages/cdktf-cli/bin/cmds/ui/terraform-json.ts
```diff
@@ -0,0 +1,31 @@
...
+  workspaces: TerraformJsonConfigBackendRemoteWorkspace;
+}
+
+export interface TerraformJsonConfigBackend {
+  remote?: TerraformJsonConfigBackendRemote;
+}
+export interface TerraformJsonConfig {
+  backend?: TerraformJsonConfigBackend;
+}
+export interface TerraformJson {
```
jsteinich, line 27: Would be nice to reuse something from the core library for this.

jsteinich (commented):
I've never used Terraform cloud and only have a high level understanding of it, so I focused on looking for changes in behavior in the non-cloud version.

packages/cdktf-cli/bin/cmds/ui/diff.tsx
```diff
@@ -38,12 +38,22 @@ const PlanSummary = ({ resources }: PlanSummaryConfig): React.ReactElement => {
   </>)
 }
 
+export const CloudRunInfo = (): React.ReactElement => {
+  const { url } = useTerraformState()
+  if (url == undefined) return <></>;
+
+  const staticElements = [url]
+
+  return <Static>{ staticElements.map(e => (<Text key={e}>This is running remotely: { e }</Text>))} </Static>
```
anubhavmishra, line 47:
```suggestion
  return <Static>{ staticElements.map(e => (<Text key={e}>Running plan in the remote backend. To view this run in a browser, visit: { e }</Text>))} </Static>
```

.github/workflows/examples.yml
```diff
@@ -33,3 +33,4 @@ jobs:
         env:
           TERRAFORM_BINARY_NAME: "terraform${{ matrix.terraform }}"
           MAVEN_OPTS: "-Xmx3G"
+          TERRAFORM_CLOUD_TOKEN: ${{ secrets.TERRAFORM_CLOUD_TOKEN }}
```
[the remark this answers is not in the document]
skorfmann: I think we can get rid of this, shouldn't be necessary anymore.

package.json
```diff
@@ -18,7 +18,8 @@
     "integration-windows": "test\\run-against-dist.bat test\\test-all.bat",
     "release-github": "tools/release-github.sh",
     "build-docker-jsii": "docker build -t hashicorp/jsii-terraform .",
-    "push-docker-jsii": "docker push hashicorp/jsii-terraform"
+    "push-docker-jsii": "docker push hashicorp/jsii-terraform",
+    "dist-clean": "lerna run dist-clean --stream && rm -rf dist"
```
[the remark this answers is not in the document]
skorfmann: yes! :)

packages/cdktf-cli/bin/cmds/helper/synth-stack.ts
```diff
@@ -9,7 +9,7 @@ interface SynthesizedStackMetadata {
   "//"?: {[key: string]: TerraformStackMetadata };
 }
 
-interface SynthesizedStack {
+export interface SynthesizedStack {
```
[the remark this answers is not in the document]
skorfmann: 👍

packages/cdktf-cli/bin/cmds/ui/models/terraform.ts
```diff
@@ -54,84 +49,17 @@ export interface TerraformOutput {
...
-
-  public get applyableResources(): PlannedResource[] {
-    const applyActions = [PlannedResourceAction.UPDATE, PlannedResourceAction.CREATE, PlannedResourceAction.DELETE, PlannedResourceAction.READ];
-    return this.resources.filter(resource => (applyActions.includes(resource.action)));
-  }
-
-  public get needsApply(): boolean {
-    return this.applyableResources.length > 0
-  }
+export interface TerraformPlan {
```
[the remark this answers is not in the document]
skorfmann: Currently, there are subtile differences in the implementation (e.g. camel cased in the cloud context). And since the Cloud story isn't fully complete with this PR, I wouldn't necessarily assume these are the only differences. However, later on, it could certainly make sense to extract parts of the implementation into an abstract class.

packages/cdktf-cli/bin/cmds/ui/models/terraform-cloud.ts
```diff
@@ -0,0 +1,241 @@
...
+    stdout(Buffer.from(logs.data, 'utf8'))
+
+    switch (result.attributes.status) {
+      case 'applied': console.log('done') ; break;
+      default: throw new  Error(`error: ${result.attributes.status}`);
+    }
+  }
+
+  public async version(): Promise<string> {
+    return ''
```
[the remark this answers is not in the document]
skorfmann: Yes, good catch. Will fill this in 👍

packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx
skorfmann: Hm, in general absolutely yes. Was just wondering why I haven't stumbled upon this. First and foremost certainly because this area lacks tests, but more importantly it's unused code. So, I'm just gonna drop it entirely.

packages/cdktf-cli/bin/cmds/ui/models/terraform-cloud.ts
```diff
@@ -0,0 +1,241 @@
...
+        result = await this.client.Runs.show(result.id)
+        await wait(1000);
+      }
+    }
+
+
+    const plan = await this.client.Plans.jsonOutput(result.relationships.plan.data.id)
+    this.run = result
+    const url = `https://app.terraform.io/app/${this.organizationName}/workspaces/${this.workspaceName}/runs/${result.id}`
+    return new TerraformCloudPlan('terraform-cloud', plan as unknown as any, url)
```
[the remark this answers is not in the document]
skorfmann: Yes, I agree. Gonna see if there's an easy way to type this.

packages/cdktf-cli/bin/cmds/ui/models/terraform-cloud.ts
```diff
@@ -0,0 +1,241 @@
...
+
+    if (deployingStates.includes(result.attributes.status)) {
+      result = await this.client.Runs.show(runId)
+      while (deployingStates.includes(result.attributes.status)) {
+        result = await this.client.Runs.show(runId)
+        await wait(1000);
+      }
+    }
+
+    const logs = await this.client.Applies.logs(result.relationships.apply.data.id)
```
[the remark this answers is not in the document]
skorfmann: Good point 👍  #491

packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx
skorfmann: We did synth before on every action. This was just moved into the actions explicitly, due to the dependency on the generated stack in order to figure out if we should use the CLI or not. In general, I'm not a 100% happy with the state of the this entire file. I hope this can be drastically simplified not too far away.

For the slow synth, I think we'd want to look into caching the generated stacks. However, that's for another day :)

packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx
skorfmann: Yes, whenever any of `[state.status, terraform]` changes, this will be invoked. Therefore this additional guard clause.

packages/cdktf-cli/bin/cmds/ui/terraform-json.ts
skorfmann: Yes 👍  #492

skorfmann:
> I think we should point to the instructions that we have for our project to help with state migration instead of the Terraform documentation. https://github.com/hashicorp/terraform-cdk/blob/master/docs/working-with-cdk-for-terraform/remote-backend.md#migrating-local-state-storage-to-remote

Thanks, that's a very good point. Wasn't even aware that we have this documentation 👍

packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx
skorfmann: 6d798dc

packages/cdktf-cli/bin/cmds/ui/models/terraform-cloud.ts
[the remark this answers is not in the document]
skorfmann: Let's do this as part of https://github.com/hashicorp/terraform-cdk/issues/492

PR #455, opened 2020-12-04
PyPi Pre-Releases

skorfmann:
With https://github.com/hashicorp/terraform-cdk/pull/421 we should be able to do pre-releases to PyPi now. 

Fixes https://github.com/hashicorp/terraform-cdk/issues/293

/cc @cmclaughlin @jsteinich

### DIFF
--- a/.github/workflows/release_next.yml
+++ b/.github/workflows/release_next.yml
@@ -5,19 +5,17 @@ on:
       - master
 
 jobs:
-  release:
+  build_artifact:
     if: github.repository == 'hashicorp/terraform-cdk'
     runs-on: ubuntu-latest
     container:
       image: hashicorp/jsii-terraform
     env:
       CHECKPOINT_DISABLE: "1"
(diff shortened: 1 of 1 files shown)


.github/workflows/release_next.yml
```diff
@@ -5,27 +5,56 @@ on:
...
+          path: dist
+
+  release_npm:
+    name: Release to NPM
+    needs: build_artifact
+    runs-on: ubuntu-latest
+    container:
+      image: hashicorp/jsii-terraform
+    steps:
+      - uses: actions/checkout@v2
```
jsteinich, line 39: This should be downloading the built artifacts and the checkout shouldn't be necessary.

skorfmann: good point 👍

skorfmann:
Screenshot 2020-12-04 at 16 32 58

🎉

PR #456, opened 2020-12-04
GitHub maven prerelease

### LINKED ISSUE #450: Publish Java snapshot versions
### Community Note

### Description

Publish snapshot versions of the Java binaries with each push to master. This is already done for Typescript and Python and would be useful for quicker testing of changes made as users start to try out using Java for cdktf.

jsteinich:
Closes #450 

While not an actual snapshot (or even using standard naming), this does match with other languages and should work with cdktf-cli@next without requiring extra builds or code changes.

Also a bit of cleanup on actual release that was missed when switching to jsii-release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -96,20 +96,17 @@ jobs:
 
   release_maven_github:
     name: Release to GitHub Maven
-    runs-on: ubuntu-latest
     needs: build_artifact
+    runs-on: ubuntu-latest
+    container:
+      image: hashicorp/jsii-terraform
     steps:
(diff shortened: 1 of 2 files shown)


skorfmann:
Do we need similar adjustments here? https://github.com/hashicorp/terraform-cdk/pull/457

jsteinich:
> Do we need similar adjustments here? #457

I don't believe so since the version can be used as is unlike python which internally changes it.
Definitely easier to test after the version is published, but I may be able to test before that.

skorfmann (approved):

Issue #196, opened 2020-07-16
Support C#


jsteinich:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Allow writing Terraform CDK code in C#

- Update jsii config to add in C#
- Add example C# solution
- Add C# project template
- Add tests

### References

Similar to [Python Work](https://github.com/hashicorp/terraform-cdk/pull/51)

skorfmann:
There's already progress on this by @pmRz00

skorfmann:
This was done with #477

PR #458, opened 2020-12-07
csharp support

pmRz00:
#196 

## Todo

- [ ] Documentation 
- [ ] Example
- [ ] Publishing Workflow 
- [ ] Integration Tests
- [ ] Cleanup


packages/cdktf-cli/templates/csharp/MyTerraformStack.csproj
```diff
@@ -0,0 +1,12 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
jsteinich, line 1: I think it would make sense to also have a `.sln` file as part of the template.

packages/cdktf-cli/templates/csharp/.hooks.sscaff.js
```diff
@@ -0,0 +1,64 @@
...
+
+function terraformCloudConfig(baseName, organizationName, workspaceName) {
+  template = readFileSync('./Main.cs', 'utf-8');
+
+  result = template.replace(`using Hashicorp.Cdktf.App;`, `using Hashicorp.Cdktf.App;
+using Hashicorp.Cdktf.NamedRemoteWorkspace;
+using Hashicorp.Cdktf.RemoteBackend;
+using Hashicorp.Cdktf.RemoteBackendProps;`);
+  result = result.replace(`new Main(app, "${baseName}");`, `Main stack = new Main(app, "${baseName}");
+new RemoteBackend(stack, RemoteBackendProps.builder().hostname("app.terraform.io").organization("${organizationName}").workspaces(new NamedRemoteWorkspace("${workspaceName}")).build());`);
```
jsteinich, line 61: I don't believe a builder is generated for C#. This would just be creating `RemoteBackendProps` using initializers.

packages/cdktf-cli/templates/csharp/.hooks.sscaff.js
```diff
@@ -0,0 +1,64 @@
...
+  // locally built package
+  if (cdktf_version == '0.0.0') {    
+    execSync(`dotnet add package '${pkgName}' --source '${srcFolder}'`, { stdio: 'inherit' });
+  }
+  else {
+    execSync(`dotnet add package '${pkgName}'`, { stdio: 'inherit' });
+  }
+
+  execSync(`\"${process.execPath}\" ${cli} get`, { stdio: 'inherit' });
+  //execSync(`\"${process.execPath}\" ${cli} synth`, { stdio: 'inherit' });
```
jsteinich, line 48: Why comment out the synth?

packages/cdktf-cli/templates/csharp/.hooks.sscaff.js
```diff
@@ -0,0 +1,64 @@
...
+
+  // Terraform Cloud configuration settings if the organization name and workspace is set.
+  if (options.OrganizationName != '') {
+    console.log(`\nGenerating Terraform Cloud configuration for '${options.OrganizationName}' organization and '${options.WorkspaceName}' workspace.....`)
+    terraformCloudConfig(options.$base, options.OrganizationName, options.WorkspaceName)
+  }
+
+  pkgFileName = nuget_cdktf.substring(nuget_cdktf.lastIndexOf('/') + 1);
+  pkgName = pkgFileName.substring(0, pkgFileName.lastIndexOf(".".concat(cdktf_version)));
+  srcFolder = nuget_cdktf.substring(0, nuget_cdktf.lastIndexOf('/'));
```
jsteinich, line 35: Might be better to use `path` to split rather than checking for `/` which is likely not cross-platform safe. There's also the option of explicitly passing other variables.

jsteinich: I'm also pretty sure these will fail when a version number is passed for `nuget_cdktf`.

packages/cdktf-cli/templates/csharp/.hooks.sscaff.js
```diff
@@ -0,0 +1,64 @@
...
+
+  srcFolder = "/root/workspace/terraform-cdk/dist/dotnet/";
```
jsteinich, line 37: Shouldn't use a hardcoded path here.

packages/cdktf-cli/templates/csharp/.hooks.sscaff.js
```diff
@@ -0,0 +1,64 @@
...
+
+  // locally built package
+  if (cdktf_version == '0.0.0') {    
+    execSync(`dotnet add package '${pkgName}' --source '${srcFolder}'`, { stdio: 'inherit' });
+  }
+  else {
+    execSync(`dotnet add package '${pkgName}'`, { stdio: 'inherit' });
```
jsteinich, line 44: Should explicitly say which version of the package to add.

examples/csharp/azure/cdktf.json
```diff
@@ -0,0 +1,6 @@
+{
+  "language": "python",
```
jsteinich, line 2: Not python.

examples/csharp/azure/README.md
```diff
@@ -0,0 +1,92 @@
+# python-azure
```
jsteinich, line 1: Need to update this or remove.

examples/csharp/azure/Main.cs
```diff
@@ -0,0 +1,23 @@
...
+using Hashicorp.Cdktf;
+
+
+namespace MyCompany.MyApp
+{
+    class MyApp : TerraformStack
+    {
+        public MyApp(Construct scope, string id) : base(scope, id)
+        {
+            // define resources here
```
jsteinich, line 12: I'm guessing this is just in-progress still, but definitely should have some resources in it.

examples/csharp/azure/MyTerraformStack.csproj
```diff
@@ -0,0 +1,13 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <RestoreSources>$(RestoreSources);./../../../dist/dotnet;https://api.nuget.org/v3/index.json</RestoreSources>
```
jsteinich, line 6: This does match how the other languages work so we can probably leave alone, but it would be kind of nice if examples didn't require it to better match using a release version.

jsteinich (changes requested):
It looks like everything is on the right track, but I'm not easily able to test due to the hardcoded path. Please let me know if I can help in any other way. I'm really looking forward to using this.

In addition to the TODO list and other comments I left just a couple other points:
* I think it would be nice for the example to include a `.gitignore` file. Not all examples have, but most do.
* One of the snapshot tests is out of date. Just go to `packages/cdktf-cli` and run `yarn test -u`.

skorfmann:
@jsteinich thanks for your review! @pmRz00 please correct me if I'm wrong, but I think any help around this PR would be appreciated.

packages/cdktf-cli/templates/csharp/MyTerraformStack.csproj
jsteinich: On second thought, don't really need it.

packages/cdktf-cli/templates/csharp/.hooks.sscaff.js
jsteinich: Changed to not be needed

skorfmann:
Closing in favour of #477 (which is the same code, but as a branch in this repo)

Issue #25, opened 2020-04-26
Complex Computed Optional Types


skorfmann:
There are optional computed complex types in the provider schema, e.g. in the [aws_security_group](https://www.terraform.io/docs/providers/aws/r/security_group.html) the attributes `egress` / `ingress`. Here an excerpt from the schema:

```
             "egress": {
                "type": [
                  "set",
                  [
                    "object",
                    {
                      // ... attributes 
                    }
                  ]
                ],
                "optional": true,
                "computed": true
              },
```

As I understand `optional` && `computed`, this means there could be `egress` or `ingress` rules created by AWS. Which, at least for `egress` is true. However, apparently the AWS Terraform provider removes this `computed` object (see [comment](https://www.terraform.io/docs/providers/aws/r/security_group.html#description-2))

At the moment the attribute is rendered without taking into account the `computed` option, which looks like this:

```typescript
  // egress
  private _egress?: SecurityGroupEgress[];
  public get egress() {
    return this._egress; // Getting the computed value is not yet implemented
  }
  public set egress(value: SecurityGroupEgress[]) {
    this._egress = value;
  }
```

We should probably enable access to the those `computed` rules as well. Something along the lines of this:

```typescript
  // egress
  private _egress?: SecurityGroupEgress[];
  public get egress(index?: string) {
    return this._egress || return new SecurityGroupEgressList(this, 'egress', index);
  }
  public set egress(value: SecurityGroupEgress[] | undefined) {
    this._egress = value;
  }
```

We'd have to generate:

- an interface (e.g. `SecurityGroupEgress`) for the `optional` type 
- a class (e.g. `SecurityGroupEgressList`) which implements the interface to wrap the `computed` value
- Make sure the return type of the getter function is intuitively usable

In particular the last point seems to be difficult.

For now it's probably fine as is, but when heading towards beta we should come back to this.

PR #480, opened 2021-01-11
Provide accessors to nested computed attributes

### LINKED ISSUE #172: Enable Attribute References for Nested Types
### Community Note

### Description

Some providers (at least the Kubernetes provider) allow attribute references on anonymous interfaces (see #170). It would be good to support this.

That would be something like this:

```ts
const namespace = new Namespace(this, 'foo', {})
namespace.inputs.generation
```

### References

https://github.com/hashicorp/terraform-cdk/issues/170

jsteinich:
Fixes #172 

This will make sure structs have all attributes available. Some won't actually be settable. Currently only enforced by terraform, but could easily add a check.

### DIFF
--- a/packages/cdktf-cli/lib/get/generator/emitter/struct-emitter.ts
+++ b/packages/cdktf-cli/lib/get/generator/emitter/struct-emitter.ts
@@ -27,7 +27,7 @@ export class StructEmitter {
       this.code.openBlock(`export interface ${struct.name}${struct.extends}`);
     }
 
-    for (const att of struct.assignableAttributes) {
+    for (const att of struct.accessibleAttributes) {
       if (att.description) {
         this.code.line(`/** ${att.description} */`);
       }
@@ -56,7 +56,7 @@ export class StructEmitter {
     this.code.openBlock(`function ${downcaseFirst(struct.name)}ToTerraform(struct?: ${struct.name}): any`);
     this.code.line(`if (!cdktf.canInspect(struct)) { return struct; }`);
     this.code.openBlock('return');
-    for (const att of struct.isClass ? struct.attributes : struct.assignableAttributes) {
+    for (const att of struct.assignableAttributes) {
       this.attributesEmitter.emitToTerraform(att, true);
     }
     this.code.closeBlock(';');
--- a/packages/cdktf-cli/lib/get/generator/emitter/struct-emitter.ts
+++ b/packages/cdktf-cli/lib/get/generator/emitter/struct-emitter.ts
@@ -27,7 +27,7 @@ export class StructEmitter {
       this.code.openBlock(`export interface ${struct.name}${struct.extends}`);
     }
 
-    for (const att of struct.accessibleAttributes) {
+    for (const att of struct.isComputed ? struct.nonAssignableAttributes : struct.assignableAttributes) {
       if (att.description) {
         this.code.line(`/** ${att.description} */`);
       }
@@ -36,7 +36,7 @@ export class StructEmitter {
     }
     this.code.closeBlock();
 
-    if (!(struct instanceof ConfigStruct)) {
+    if (!(struct instanceof ConfigStruct) && !struct.isComputed) {
       this.emitToTerraformFuction(struct);
     }
   }
--- a/packages/cdktf-cli/lib/get/generator/models/attribute-model.ts
+++ b/packages/cdktf-cli/lib/get/generator/models/attribute-model.ts
@@ -40,7 +40,7 @@ export class AttributeModel {
   }
 
   public get typeDefinition() {
-    const optional = this.required ? '' : '?';
+    const optional = this.optional ? '?' : '';
     return `${this.name}${optional}: ${this.type.name}`;
   }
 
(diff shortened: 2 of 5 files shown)


jsteinich:
I checked over a few examples in referenced tickets and the attributes are now in the generated interfaces, but not in the `toTerraform` functions as expected.

Since these generally show up in interfaces, the properties are optional. This is a bit disingenuous since they are both not-settable and *always available. 

There's a reasonable chance for things to not be fully accessible since lists and maps of custom types aren't fully supported, but that's a separate issue, although some of the ideas in https://github.com/hashicorp/terraform-cdk/issues/25 could solve both issues.

skorfmann:
Cool, thanks! I'll give it a try locally 👍

skorfmann:
I finally got to look at this, sorry for the delay. 

> Since these generally show up in interfaces, the properties are optional. This is a bit disingenuous since they are both not-settable and *always available.

Yes, I'd expect that these attributes wouldn't show up in the interface. Let's take [AwsInstance](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/instance#import) `root_block_device` as an example. It has two attribute references - `deviceName` and `volumeId` which are now additionally generated as part of the `InstanceRootBlockDevice` interface:

```ts
export interface InstanceRootBlockDevice {
    // ...
    readonly deviceName?: string;  
    readonly volumeId?: string;
    //...
  }
  export class Instance extends cdktf.TerraformResource {
    //...
  
    // root_block_device - computed: false, optional: true, required: false
    private _rootBlockDevice?: InstanceRootBlockDevice[];
    public get rootBlockDevice() {
      return this.interpolationForAttribute('root_block_device') as any;
    }

    //...
  }
```

My expectation would be, that I would be able to reference this along the lines of this:

```
instance.rootBlockDevice(0).volumeId
```

For reference, the schema for this attribute:

```json
              "root_block_device": {
                "type": [
                  "set",
                  [
                    "object",
                    {
                      "delete_on_termination": "bool",
                      "device_name": "string",
                      "encrypted": "bool",
                      "iops": "number",
                      "kms_key_id": "string",
                      "tags": [
                        "map",
                        "string"
                      ],
                      "throughput": "number",
                      "volume_id": "string",
                      "volume_size": "number",
                      "volume_type": "string"
                    }
                  ]
                ],
                "description_kind": "plain",
                "computed": true
              },
```

Does this match with your expectation? @jsteinich?

jsteinich:
That does match how this PR is intended to work.

The best alternative I've though of to adding to the interface would be to generate a separate class returned by the getter.
- Would allow for keeping the interface clean with what's actually settable while still allowing access to the properties.
- Cost of having another class for users to comprehend, be named, and get compiled.

Side note, that example isn't actually referenceable by index currently due to #25.

skorfmann:
> The best alternative I've though of to adding to the interface would be to generate a separate class returned by the getter.
> 
> * Would allow for keeping the interface clean with what's actually settable while still allowing access to the properties.
> * Cost of having another class for users to comprehend, be named, and get compiled.

yes, that's what I was thinking as well 👍

jsteinich:
I started down the path of generating alternate types when I again ran into getters and setters needing to be the same type.
Some options:

1. Generate another get method that gives access to the computed type
2. Always return the computed type as the normal type and force the user to cast to the generated type if they need access to it
3. Convert the interfaces to classes and just not expose setters for the only computed types

I want to test if 3 is possible without converting to a class (I'm worried jsii may not work well); otherwise, leaning towards 1.

skorfmann:
> I started down the path of generating alternate types when I again ran into getters and setters needing to be the same type.

Do you have an example?

jsteinich:
```typescript
export interface InstanceRootBlockDevice {
    // ...
}

export interface InstanceRootBlockDeviceComputed extends InstanceRootBlockDevice {
    readonly deviceName: string;  
    readonly volumeId: string;
}

export class Instance extends cdktf.TerraformResource {
    //...
  
    // root_block_device - computed: false, optional: true, required: false
    private _rootBlockDevice?: InstanceRootBlockDevice[];
    public get rootBlockDevice() {
      return this.interpolationForAttribute('root_block_device') as any as InstanceRootBlockDeviceComputed[];
    }
    public set rootBlockDevice(value: InstanceRootBlockDevice[] ) {
      this._rootBlockDevice = value;
    }
    //...
}
```

Ignoring for a moment that the `as any` results in mostly unfunctional code, the `as InstanceRootBlockDeviceComputed[]` will fail because the getter and setter types for `rootBlockDevice` will no longer be the same.

jsteinich:
> I want to test if 3 is possible without converting to a class

Turns out it's not possible to have getters/setters in a TS interface. There are readonly properties, but they are really init-only which doesn't accomplish what we are after.

skorfmann:
Gonna have to play with this locally to get a feeling for the issues involved here :)

jsteinich:
Another thought is to just change the property type to the computed version always. It does mean that the setter would allow values that can't be set (I believe Terraform will detect this), but then the getter works as expected. I believe the common pattern is to use the constructor rather than the setter, so that seems like a decent tradeoff.

skorfmann:
@jsteinich can this be closed or are there still parts which are relevant?

jsteinich:
> @jsteinich can this be closed or are there still parts which are relevant?

There are some relevant bits in the comments, but mostly I've kept this around as a reference/reminder that we haven't fully supported everything yet. Doesn't really need to be open to accomplish that though.

Issue #484, opened 2021-01-12
Document Usage of RemoteState 


skorfmann:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

We have the basic concept for a [Terraform Remote State](https://github.com/hashicorp/terraform-cdk/blob/790ab578c307b5efef62777be4ab11d62ec45b6a/packages/cdktf/lib/terraform-remote-state.ts), but it looks like it's undocumented. We should document this somehow.

jsteinich:
It's not the most detailed, but there is some documentation [here](https://github.com/hashicorp/terraform-cdk/blob/master/docs/working-with-cdk-for-terraform/data-sources.md#terraform-remote-state).

skorfmann:
> It's not the most detailed, but there is some documentation [here](https://github.com/hashicorp/terraform-cdk/blob/master/docs/working-with-cdk-for-terraform/data-sources.md#terraform-remote-state).

Ah, cool. Didn't see that :D

cmclaughlin:
I opened PR #487 with an example.

Oh, re-reading the rest of that file... I think the intention of that doc was reading state, not writing/storing it.

I'll go ahead and close out the PR.

By the way, issue #247 makes it difficult to read-in state created from cdktf because the resource IDs have a UUID in them.

jsteinich:
> By the way, issue #247 makes it difficult to read-in state created from cdktf because the resource IDs have a UUID in them.

https://github.com/hashicorp/terraform-cdk/pull/329 introduced changes that can remove the UUID* (sometimes automatically, sometimes manually)

skorfmann:
Since there is documentation, I'll close this issue here.

PR #487, opened 2021-01-12
Add Python docs/example for remote state on s3

cmclaughlin:
Closes issue #484


cmclaughlin:
Oh, re-reading the rest of that file... I think the intention of that doc was reading state, not writing/storing it.

I'll go ahead and close out the PR.

jsteinich:
I don't think there are any current samples/examples of using a remote backend in python, so perhaps having this in a different location would make sense.
https://github.com/hashicorp/terraform-cdk/blob/master/docs/working-with-cdk-for-terraform/remote-backend.md mentions that they exist and https://github.com/hashicorp/terraform-cdk/tree/master/examples/typescript/backends has some typescript examples.

PR #494, opened 2021-01-14
Add CodeQL

skorfmann:
Let's give this a try https://github.com/github/codeql-action


smorimoto:
It's probably better to run CodeQL after building the source. The autobuild step is meaningless in this case.

skorfmann:
> It's probably better to run CodeQL after building the source. The autobuild step is meaningless in this case.

I've been wondering the same. What's in this PR was pretty much automatically generated by Github. I assumed there's probably a good reason why this is scheduled without really looking into it. However, will do that now :)

skorfmann:
Hm, doesn't seem to be very useful in the context of this project without digging deeper in the configuration, see https://github.com/hashicorp/terraform-cdk/security/code-scanning?query=ref%3Arefs%2Fpull%2F494%2Fmerge+tool%3ACodeQL

Screenshot 2021-01-29 at 22 06 00

Closing for now, might come back to this at a later stage.

PR #497, opened 2021-01-15
Improve C# init behaviour for Github Package Registry

skorfmann:
This avoids init behaviour as seen here (which is due to the requirement of having to authenticate for the Github Package registry):

Screenshot 2021-01-15 at 14 27 10

This PR makes sure the `init` isn't erroring out and provides guidance to the user how to authenticate to the Github Package registry. We can revert back to the original behaviour once we got the official nuget.org registry distribution.

Note: We'll do a follow up for the Java template which has similar issues.

### DIFF
--- a/packages/cdktf-cli/templates/csharp/.hooks.sscaff.js
+++ b/packages/cdktf-cli/templates/csharp/.hooks.sscaff.js
@@ -42,12 +42,25 @@ exports.post = options => {
         <add key="NuGet official package source" value="https://api.nuget.org/v3/index.json" />
       </packageSources>
     </configuration>`, 'utf-8');
+  } else {
+    writeFileSync('./NuGet.Config', `<?xml version="1.0" encoding="utf-8"?>
+    <configuration>
+        <packageSources>
+            <add key="github" value="https://nuget.pkg.github.com/hashicorp/index.json" />
+        </packageSources>
+        <packageSourceCredentials>
+            <github>
+                <add key="Username" value="%GITHUB_USER%" />
+                <add key="ClearTextPassword" value="%GITHUB_TOKEN%" />
+            </github>
+        </packageSourceCredentials>
+    </configuration>`, 'utf-8');
   }
 
-  execSync(`dotnet restore`, { stdio: 'inherit' });
+  // execSync(`dotnet restore`, { stdio: 'inherit' });
 
   execSync(`\"${process.execPath}\" ${cli} get`, { stdio: 'inherit' });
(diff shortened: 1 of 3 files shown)


packages/cdktf-cli/templates/csharp/.hooks.sscaff.js
```diff
@@ -42,12 +42,25 @@ exports.post = options => {
...
+            <github>
+                <add key="Username" value="%GITHUB_USER%" />
+                <add key="ClearTextPassword" value="%GITHUB_TOKEN%" />
+            </github>
+        </packageSourceCredentials>
+    </configuration>`, 'utf-8');
   }
 
-  execSync(`dotnet restore`, { stdio: 'inherit' });
+  // execSync(`dotnet restore`, { stdio: 'inherit' });
```
jsteinich, line 60: Why comment out the restore and synth?

test/test-csharp-app/test.js
```diff
@@ -16,6 +16,8 @@ process.chdir(folder);
 // initialize an empty project
 execSync(`cdktf init --template csharp --project-name="csharp-test" --project-description="csharp test app" --local`, { stdio: 'inherit', env });
 
+execSync(`dotnet restore`)
```
jsteinich, line 19: Is there a reason for the explicit restore? It doesn't hurt anything, but the synth below will cause it to happen.

packages/cdktf-cli/templates/csharp/.hooks.sscaff.js
skorfmann: It's causing errors, since the `Nuget.Config` isn't complete. Essentially, the user has to authenticate manually and edit the config - unless you know a way to do this interactively? `synth` then errors, because the dependencies aren't resolved.

test/test-csharp-app/test.js
skorfmann: Haven't tested it, just assumed it would be necessary when it's omitted in the init template.

packages/cdktf-cli/templates/csharp/.hooks.sscaff.js
jsteinich: I guess it comes down to when should the user get the error messaging. Unless they have things already setup, they'll either get an error on the init or an error when trying to do anything else. 🤷

skorfmann: The reasoning is, that the help text will guide the user how to proceed without running into any error.

PR #499, opened 2021-01-16
Add test-case for equals -> equalTo

skorfmann:
Was about to fix it and realized this is already there. So, here's just an addition to the snapshot test for this.


jsteinich (approved):
Looks good

PR #510, opened 2021-01-20
Implement java gradle multi module example

marcoferrer:
This PR introduces the following:
* A cdk example using java as the target language and gradle and the build system.
  * This includes setting up Github artifact repo authentication.
* Demonstrate how to share common cdk configuration across other cdk projects.
* Demonstrate how to implement a module with HCL sources that reference declarations created by common cdk modules.

The following files are a by-product of initializing gradle and bare no significance on the implementation of the example:
* `examples/java-gradle/shared-module/gradle.properties`
* `examples/java-gradle/shared-module/gradle/wrapper/gradle-wrapper.jar`
* `examples/java-gradle/shared-module/gradle/wrapper/gradle-wrapper.properties`
* `examples/java-gradle/shared-module/gradlew`
* `examples/java-gradle/shared-module/gradlew.bat`


examples/java-gradle/shared-module/module-b/cdktf.json
```diff
@@ -0,0 +1,11 @@
+{
+  "language": "java",
+  "app": "../gradlew -q :module-b:run",
+  "output": "./",
```
jsteinich, line 4: Maybe add a comment about why you're using a non-default output. It took me a minute to release why this is necessary.

jsteinich (commented):
This is a really interesting example. Thanks for creating it.

We may want it to exist at `examples/java/grade-shared-module` or similar to keep language grouping.

Consider adding a `package.json` file so that automated tests will validate that the example works correctly.

examples/java-gradle/shared-module/module-b/cdktf.json
marcoferrer: Where would be the best place to address this? I could potentially add it to the README since this is a json file.
Thoughts?

jsteinich: The README is probably a good spot. Could also add as a `//` property above the output property.

skorfmann:
That's pretty cool - thanks for creating this example! 

How should we deal with the Github config in the Github Actions pipeline?

examples/java/gradle-shared-module/build.gradle
```diff
@@ -0,0 +1,44 @@
...
+    apply plugin: 'java'
+
+    repositories {
+        mavenCentral()
+        maven {
+            name = "GitHubPackages"
+            url = uri("https://maven.pkg.github.com/hashicorp/terraform-cdk")
+            credentials {
+                username = project.githubUsername
+                password = project.githubToken
```
marcoferrer, line 14: @skorfmann I can update this line to pull the token from an environment variable and fallback to a project property if not found. We would need a token generated and added to travis CI. It would be preferable to only grant permissions for reading GitHub package repos.

examples/java/gradle-shared-module/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "@examples/java-gradle-shared-module",
+  "version": "0.0.0",
+  "license": "MPL-2.0",
+  "scripts": {
+    "reinstall": "mvn install:install-file -Dfile=./../../../dist/java/com/hashicorp/cdktf/0.0.0/cdktf-0.0.0.jar -DgroupId=com.hashicorp -DartifactId=cdktf -Dversion=0.0.0 -Dpackaging=jar",
```
marcoferrer, line 6:
```
    "reinstall": "mvn install:install-file -Dfile=./../../../dist/java/com/hashicorp/cdktf/0.0.0/cdktf-0.0.0.jar -DgroupId=com.hashicorp -DartifactId=cdktf -Dversion=0.0.0 -Dpackaging=jar",
```
Im not entirely sure what should be defined here or what part of the CI lifecycle `reinstall` is executed.

examples/java/gradle-shared-module/module-b/cdktf.json
```diff
@@ -0,0 +1,11 @@
+{
+  "language": "java",
+  "app": "../gradlew -q :module-b:run",
+  "output": "./",
```
marcoferrer, line 4:
```suggestion
  "//": "The target output directory is set to the current working directory so that definitions generated by cdk can be referenced in local HCL sources"
  "output": "./",
```

examples/java/gradle-shared-module/build.gradle
jsteinich: I'm pretty sure the github action runner already has a user/token with permissions, so adding to environment variables should be pretty easy.

examples/java/gradle-shared-module/package.json
jsteinich: reinstall, get, and synth are ran in sequence as part of the CI lifecycle.

Most examples rely on using a version that was just built. I'm thinking that this example may be more valuable showing what it's like to pull from GitHub, in which case this could be a no-op.

examples/java/gradle-shared-module/build.gradle
marcoferrer: Sorry about that, I was thinking about another project when mentioning travis. Ill update the workflow now to add a reference to the token var

examples/java/gradle-shared-module/package.json
marcoferrer: I decided to leave it so that the example can also get tested against the local latest artifact version

examples/java/gradle-shared-module/module-b/main.tf
```diff
@@ -0,0 +1,20 @@
...
+  region                     = "us-west-1"
+  alias                      = "external"
+  skip_requesting_account_id = true
+  profile                    = "external"
+}
+
+resource "aws_secretsmanager_secret" "local-secret" {
+  name     = "internal-secret"
+  // Here we're referencing a provider generated by the
+  // base application module that was imported vis gradle
```
jsteinich, line 12:
```suggestion
  // base application module that was imported via gradle
```

jsteinich (approved):
Everything is looking good.

skorfmann (approved):

PR #512, opened 2021-01-20
TerraformHclModule outputs need to use tokens

jsteinich:
Changing output reference to use tokens similar to data sources, variables, etc., since the cast from any was causing exceptions at runtime for non-string types.


skorfmann (approved):
👍

Issue #531, opened 2021-01-30
CDK provider for DigitalOcean


jubairsaidi:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

It would be great if we could get a CDK provider for DO! I was recently introduced to CDKTF and was super excited about it until I realized that there is no DigitalOcean provider for it... Any chance we can get one? and/or an expected ETA if it's currently in the works?

skorfmann:
@jubairsaidi thanks for creating this issue. You can generate the provider bindings for every Terraform provider, see [here](https://github.com/hashicorp/terraform-cdk/blob/main/docs/working-with-cdk-for-terraform/using-providers-and-modules.md#importing-providers-and-modules). We do publish a few [pre-built providers](https://github.com/terraform-cdk-providers/), but not for DigitialOcean at this point.

I can see that this is not really clear when reading through the Readme and the first steps guide. So, I'll take this as an opportunity to improve the documentation around this.

jubairsaidi:
nice! your team f*n ROCK! thank you :)

jubairsaidi:
So I did use a generated provider for DO and it does need some work (like Droplet.image being a string or number, and When Droplet.count is specified it results in a list of resources rather than single instance) and am sure there are others.  If the maintainers of this project create a pre-built DO provider, I would be more than glad to submit a PR with fixes I find.

jsteinich:
The pre-built providers are built and updated by generating providers the same way that users are able to.
The code generated is based on [terraform providers schema](https://www.terraform.io/docs/cli/commands/providers/schema.html) output.
There are occasions where the provider itself doesn't provide the correct definition of its schema. In that case filing an issue with the provider may be appropriate.
It's also possible the code generation itself has issues, in which case specifics can be reported here.

I haven't used the DO provider before, but it looks like `Droplet.image` is a string in the provider source code.
What value are you providing for `Droplet.count`? It should result in that many instances being created.

jubairsaidi:
[...]

PR #550, opened 2021-02-05
Emphasize that we support all modules and providers of Terraform

skorfmann:
It wasn't really obvious that we're supporting the entire ecosystem of modules and providers of Terraform. I've seen feedback on various channels about this, e.g.  https://github.com/hashicorp/terraform-cdk/issues/531#issuecomment-770290934 

This PR makes this a bit more clear early on in the Readme. It could use an image to make this even more obvious (something like the image in https://cdk.tf/launch tailored to cdktf only).

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,9 +15,7 @@ HashiCorp Terraform.
 
 ## Overview
 
-The CDK for Terraform project helps users define infrastructure resources using supported programming languages and generates a Terraform configuration in JSON. Users can then use the Terraform or the CDK for Terraform CLI to deploy the application.
-
-The project uses [AWS CDK](https://aws.amazon.com/cdk/) [constructs](https://github.com/aws/constructs) to define applications.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,9 +15,7 @@ HashiCorp Terraform.
 
 ## Overview
 
-The CDK for Terraform project helps users define infrastructure resources using supported programming languages and generates a Terraform configuration in JSON. Users can then use the Terraform or the CDK for Terraform CLI to deploy the application.
-
-The project uses [AWS CDK](https://aws.amazon.com/cdk/) [constructs](https://github.com/aws/constructs) to define applications.
+CDK for Terraform apps are written in one of the supported programming languages and can leverage the entire ecosystem of Terraform [modules and providers](./docs/working-with-cdk-for-terraform/using-providers-and-modules.md). The application [synthesizes](./docs/working-with-cdk-for-terraform/synthesizing-config.md) Terraform configuration in JSON. Users can then use the Terraform or the CDK for Terraform CLI to deploy the application.
```
pkolyvas, line 18: Nice addition!

Small suggestion on the last sentence... `Users can use...`

How about:
```Terraform or the CDK for Terraform can then be used to deploy the application.```

pkolyvas (commented):
Left a small comment about a sentence, but overall this looks like a nice clarification!

README.md
skorfmann: See https://github.com/hashicorp/terraform-cdk/pull/550/commits/743073c1f69589b4f3456a53bf3a7cce08add7b1

pkolyvas (approved):

Issue #351, opened 2020-08-27
Running under the proxy


soulman-is-good:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

In our organisation we use proxy to get to internet.

When run `cdktf get` exception is thrown:
```
Error: getaddrinfo ENOTFOUND checkpoint-api.hashicorp.com
```
So I have to go to sources and add proxy to checkpoint.js request to get typescript files for azure provider.

If would be great if cli would respect HTTPS_PROXY environment variable

### References

https://github.com/hashicorp/terraform-cdk/blob/master/packages/cdktf-cli/lib/checkpoint.ts#L25

skorfmann:
Thanks for reporting this issue, I believe that's a regression of #326 - Networking issues shouldn't interrupt the workflow here.

skorfmann:
In the meantime, you can disable the metrics via the `CHECKPOINT_DISABLE` env.

soulman-is-good:
Thanks @skorfmann that works for me

sonfrau:
> 
> 
> In the meantime, you can disable the metrics via the `CHECKPOINT_DISABLE` env.

The command "CHECKPOINT_DISABLE=true cdktf get" worked for me.

Thanks!

timpur:
i just ran into this problem also, wondering if we can add the env to `cdktf.json` as context or something so that we can consistently disable it for the project ?

jsteinich:
> wondering if we can add the env to `cdktf.json` as context or something

Not currently, but that sounds like a reasonable request.

skorfmann:
As mentioned in https://github.com/hashicorp/terraform-cdk/issues/545 we should just catch the error and ignore it.

Issue #544, opened 2021-02-03
cdktf init for java fails with timeout error and .gen folder is empty


sbattagi:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### cdktf & Language Versions
cdktf version: 0.1.0
Terraform v0.14.5

### Affected Resource(s)
Below command is failing with timeout error, not generating the kubernetes provider files.
cdktf init --template="java" --local

### Debug Output
error:
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  2.805 s
[INFO] Finished at: 2021-02-03T12:20:38Z
[INFO] ------------------------------------------------------------------------
⠼ downloading and generating providers...
/home/software/node-v14.15.4-linux-x64/lib/node_modules/cdktf-cli/node_modules/yoga-layout-prebuilt/yoga-layout/build/Release/nbind.js:53
        throw ex;
        ^

Error: connect ETIMEDOUT 172.64.204.26:443
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1146:16)
Emitted 'error' event on ClientRequest instance at:
    at TLSSocket.socketErrorListener (_http_client.js:469:9)
    at TLSSocket.emit (events.js:315:20)
    at emitErrorNT (internal/streams/destroy.js:106:8)
    at emitErrorCloseNT (internal/streams/destroy.js:74:3)
    at processTicksAndRejections (internal/process/task_queues.js:80:21) {
  errno: -110,
  code: 'ETIMEDOUT',
  syscall: 'connect',
  address: '172.64.204.26',
  port: 443
}
cdktf init [OPTIONS]

Create a new cdktf project from a template.

Options:
  --version                   Show version number                                                                     [boolean]
  --disable-logging           Dont write log files. Supported using the env CDKTF_DISABLE_LOGGING.    [boolean] [default: true]
  --disable-plugin-cache-env  Dont set TF_PLUGIN_CACHE_DIR automatically. This is useful when the plugin cache is configured
[...]

PR #552, opened 2021-02-08
Checkpoint Error Handling / Timeout

skorfmann:
This PR adds proper error handling for the request object plus a timeout of 1s. This will silently ignore errors when Checkpoint isn't reachable for some reason.

Furthermore, this cleans up a bit of obsolete code and reorganizes deps / devDeps.

Fixes: https://github.com/hashicorp/terraform-cdk/issues/351
Related to: https://github.com/hashicorp/terraform-cdk/issues/544


jsteinich (approved):
I haven't really used that request method before, but seems fine.

PR #555, opened 2021-02-09
Update resource-model.ts

johnjelinek:
The existing import statement caused some headaches for my project. I was having to manually modify the generated files after running `cdktf get`. Hopefully, not anymore!

### DIFF
--- a/packages/cdktf-cli/lib/get/generator/models/resource-model.ts
+++ b/packages/cdktf-cli/lib/get/generator/models/resource-model.ts
@@ -44,7 +44,7 @@ export class ResourceModel {
     this.terraformSchemaType = options.terraformSchemaType
     this.dependencies = [
       `import { Construct } from 'constructs';`,
-      `import * as cdktf from 'cdktf';`
+      `import cdktf from 'cdktf';`
     ]
   }
 


johnjelinek:
hmm: https://github.com/hashicorp/terraform-cdk/pull/555/checks?check_run_id=1859940537#step:6:997
The tests say there is no default export, but I definitely have to  transpile to something like ES2015 instead of CommonJS, and then modify the `.ts` files with the import statement format referenced in this change. If it shouldn't be a PR, is there another way to configure the generated code to make this change? I suppose I could make a build task to generate the provider code and then make the text replacement.

johnjelinek:
(note: I'm developing my cdk code with ClojureScript)

jsteinich:
I believe your change is different since there isn't a `cdktf` exported from `cdktf`. Perhaps something like `import cdktf = require('cdktf');` would work.

How are you referencing the code in your project? Perhaps treating it as ["Foreign" code](https://clojurescript.org/reference/dependencies) would also help.

johnjelinek:
Specifically, where this is biting me is when the typescript is being
generated for the docker provider. I'm only able to consume it after
modifying the import line the way I did in the diff for each of the files.
Maybe it's a different repo to support that provider generation?

skorfmann:
Thanks for this PR @johnjelinek. While ClojureScript isn't supported by `cdktf`, I'd like to understand your use-case a bit better. So, out of curiosity - Do you have an example for this? 

> Specifically, where this is biting me is when the typescript is being generated for the docker provider.

Does this mean it works for other providers out of the box?

skorfmann:
Gonna close this for now. Please feel free to reopen if that's still something you'd like to see being discussed.

PR #559, opened 2021-02-11
Upgrade to ink v3

skorfmann:
This PR upgrades to [ink](https://github.com/vadimdemedes/ink) v3. This required a small amount of changes around removed `<Color>` elements. See more details about what changed in v3 [here](https://github.com/vadimdemedes/ink/releases/tag/v3.0.0-0).

A few neat things which are made possible by the upgrade:

### React Devtools

```
export DEV=true
cdktf diff
```

And in another terminal

```
npx react-devtools
```

Will give you 
Screenshot 2021-02-11 at 11 02 12

### Terminal Dimensions

With the capability to [measure elements](https://github.com/vadimdemedes/ink/tree/585e9cfb9e6c8c88820e8e541b21e0ce774ec73b#measureelementref) and subscribe to [terminal dimensions](https://github.com/cameronhunter/ink-monorepo/tree/master/packages/ink-use-stdout-dimensions), we are now in a position to handle [scrollback issues](https://github.com/hashicorp/terraform-cdk/issues/115)

### Better console output

Doing something like `console.log('foo')` won't be overwritten anymore by the react rendering output

### Error Boundary

A way nicer way of default handling of errors - see https://github.com/vadimdemedes/ink/pull/303

### References

fixes: #166

Issue #141, opened 2020-07-07
Test CLI commands Closer to Real User Experience


skorfmann:
The init command was broken due to the input options being colorized via ansi codes, which the user realistically can't provide. The bug wasn't surfaced in the integration tests, since colourizing is automatically disabled in CI envs (see https://github.com/chalk/supports-color)

We need some kind of integration test, which actually resembles the real user experience as close a possible, to prevent bugs like this in the future.

Related to #139 and #140

PR #567, opened 2021-02-16
Refactor Integration Tests to Jest

skorfmann:
While working on the Ink v3 upgrade (#559) I got pretty annoyed by the current state of integration tests. As part of #559, I made a little proof of concept for using Jest as the test orchestrator for the integration tests and it turned out pretty good. This PR migrates all existing integration tests.

The main benefits include: 

- Easier to manage diffs thanks to Jest snapshots 
- Runs on Linux / Windows out of the box 
- Less and simpler code / Way less bash scripting 

This is supposed to be merged into #559


test/typescript/init-with-colors/test.ts
```diff
@@ -0,0 +1,21 @@
...
+import { TestDriver } from "../../test-helper";
+
+// Since chalk auto-detects the capabilities, colored output is
+// deactivated by default in non-tty environments. We had an issue
+// in the past, where this turned out to be a problem
+
+describe("test with colors", () => {
+  let driver: TestDriver;
+
+  test("build providers", () => {
```
jsteinich, line 16: I know this matches what was there, but is it really testing anything?

test/typescript/diff-deploy-destroy/test.ts
```diff
@@ -0,0 +1,52 @@
...
+  beforeAll(() => {
+    driver = new TestDriver(__dirname)
+    driver.setupTypescriptProject()
+  });
+
+  test("diff", () => {
+    expect(driver.diff()).toMatchInlineSnapshot(`
+      "Stack: hello-deploy
+      Resources
+       + NULL_RESOURCE       test                null_resource.test
```
jsteinich, line 21: This did lose the test where the feature flags aren't enabled, but that seems fine.

.github/workflows/build.yml
```diff
@@ -18,6 +18,16 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+      - name: Get yarn cache directory path
```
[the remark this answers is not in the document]
skorfmann: Gonna check, thanks 👍

test/typescript/init-with-colors/test.ts
skorfmann: Just looked it up again - that was the original issue why this test is here https://github.com/hashicorp/terraform-cdk/issues/141 - There's probably a better way to test this though. Gonna think about it.

test/typescript/diff-deploy-destroy/test.ts
skorfmann: Hm, I'll check if that's tested somewhere else.

.github/workflows/build.yml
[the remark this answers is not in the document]
skorfmann: Hm, it's unclear what the actual benefits of changing the cache  implementation would be. The official cache action Readme [recommends](https://github.com/actions/cache/blob/main/examples.md#node---yarn) the way it was implemented here. It seems to cut the install time by about 50% in our case, so that's quite good already. Will leave it as is for now.

test/typescript/diff-deploy-destroy/test.ts
skorfmann: See 0a1ab00

test/typescript/init-with-colors/test.ts
skorfmann: Added a reference to the test. I think it would be better to test this in isolation, but didn't want to get into this right now :) f2fbc79

jsteinich (approved):
Looks good. Thanks for adding that test.

PR #570, opened 2021-02-17
Remove jfrog references

mdeggies:
This removes all references to `hashicorp.jfrog.io` which has been replaced with a self-hosted registry, `docker.mirror.hashicorp.services`


skorfmann (approved):
Thanks!

PR #576, opened 2021-02-19
feat(cli): Add inline doc links to Terraform provider docs when generating provider bindings

skorfmann:
This adds jsdoc links to generated resources / data sources for a provider.

Screenshot 2021-02-19 at 14 59 31

The url which we're using seems to be deprecated. There are redirects in place, so it's not a big issue right now. However, it should be addressed at some point. the new url structure expects the namespace to be present (something like `hashicorp/aws`).

This could be easily extend to attribute level anchor links to the Terraform registry. Something like `https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/dynamodb_table#name`

Todo:

- [x] Check Python
- [x] Check Dotnet
- [x] Check Java
- [x] Check Go (as soon as it is merged)


skorfmann:
For Python, it looks like this - the formatting is not properly transformed, but the link would be usable nevertheless. Not sure if there's a way to make appearance better. 

Screenshot 2021-05-26 at 17 05 58
 

The wording for data sources doesn't really make sense. Gonna change that.

skorfmann:
> For Python, it looks like this - the formatting is not properly transformed, but the link would be usable nevertheless.

That's the same for AWS CDK, hence I'm assuming there's no jsdoc transforming in jsii

skorfmann:
a504733 adds direct links to the attribute of resource in the Terraform Registry. It seems to work for the majority of the popular providers. 🎉 

Screenshot 2021-05-26 at 21 15 14

skorfmann:
> The wording for data sources doesn't really make sense. Gonna change that.

Changed that slightly.

skorfmann:
Java looks pretty good as well 

Screenshot 2021-05-26 at 21 48 59

skorfmann:
Go seems to pick up the resource level link, but not the attribute level ones. But that's close enough for now.

skorfmann:
csharp works as well 🎉  

Screenshot 2021-05-26 at 21 55 32

ansgarm (approved):

packages/cdktf-cli/lib/get/generator/models/resource-model.ts
```diff
@@ -91,6 +92,14 @@ export class ResourceModel {
     return this.isProvider ? this.provider : this.isDataSource ? this.terraformType.replace(/^data_/, '') : this.terraformType
   }
 
+  public get terraformDocName(): string {
+    if (this.isDataSource) {
+      return toSnakeCase(this.className.replace(new RegExp(`^data${this.provider}`, 'i'), ''))
```
jsteinich, line 97: If the name collision code hits, then the class name will no longer match. `terraformResourceType` + removing the provider might be a better match for this case.

skorfmann: thanks for the feedback - caaf6ca addressed this 👍

PR #584, opened 2021-03-01
Build Modules from Various Sources

### LINKED ISSUE #522: Module Support
The overall goal of this issue is to extend the support for modules within `cdktf`. 

At the moment, we only support generating (via `cdktf get`) packages for modules which are published in the [Terraform registry](https://registry.terraform.io/).  While this covers a good part of the public Terraform module ecosystem, it's limiting users who want to use modules in version-control systems, local disk or other sources.

We will add support for generating Typescript classes using [all module sources currently supported by Terraform Core](https://www.terraform.io/docs/language/modules/sources.html). The experience of using these generated classes should feel idiomatic within the respective target language of the user.

### Implementation 

At the moment we're using a dedicated Terraform Registry client to fetch the required meta information about a module to generate the code.

When looking at the various ways of consuming Terraform modules, it's better to delegate the work of fetching the modules to Terraform Core directly. This could be done via defining a module block in a temporary working directory, running terraform init and use [Terraform Config Inspect](https://github.com/hashicorp/terraform-config-inspect) to generate the metadata. 

Compared to the direct Terraform Registry API call, this will certainly a bit slower and use more resources, e.g. the terraform init call will download all required module dependencies as well. On the upside, we don't have to reimplement functionality which already exists and we'll always support what Terraform Core is going to support in the future.

On a high level view, this will include:

- [x] Replace the Terraform Registry Client by a Terraform workflow similar to the Provider generator step
- [x] Figure out if we can use [Terraform Config Inspect](https://github.com/hashicorp/terraform-config-inspect) directly or if it's better to reimplement the relevant parts in Typescript
- [ ] ~~Make sure Module Outputs are properly typed and are using Tokens where suitable (related to #525)~~

### Issues Resolved by this Feature

- [x] Extend Support for Modules beyond the Terraform Registry https://github.com/hashicorp/terraform-cdk/issues/16
- [x] Support Submodules (likely addressed in #16) - https://github.com/hashicorp/terraform-cdk/issues/520
- [x] Fix error prone versioning in `cdktf.json` (likely addressed in #16) - https://github.com/hashicorp/terraform-cdk/issues/425
- [ ] Variable default handling inconsistent (at least when generated from Terraform Registry) - https://github.com/hashicorp/terraform-cdk/issues/27 
- [ ] Improve typing of module outputs - https://github.com/hashicorp/terraform-cdk/issues/287

skorfmann:
This adds a new package, `@cdktf/hcl2json` which is being used to to convert HCL to JSON. This is required to get the inputs and outputs for a given module.

This aims to fully resolve #522


skorfmann:
In general, this is ready for review. I'm gonna adda a few test cases around modules, fix the existing tests and perhaps extract a few classes into dedicated files. Conceptually it's going to stay like this for now.

On a high level, this merged the module and provider schema generation. Both is now generated from a Terraform config, where the module schema is processed via `@cdktf/hcl2json`. 

The naming part for modules is a bit tricky. Since this supports various module sources now, a simple string source doesn't cut it anymore. Hence, this introduces an additional option to define modules - both options will be valid:

```js
{
  terraformModules: [
    "terraform-aws-modules/vpc/aws@2.39.0",
    {
      name: "AwsVpc",
      source: "terraform-aws-modules",
      version: "2.39.0"
    }
  ]
}
```

Furthermore, there's a breaking change regarding the generated module name. Given the following config:

```js
{
  terraformModules: [
    "terraform-aws-modules/vpc/aws@2.39.0",
  ]
}
```

Before a class `Vpc` would have been generated. Now it will generate a class name `TerraformAwsModulesVpcAws`. This makes the name handling simpler. An upgrade path for a user with existing module config would be to change to a module configuration object like 

```js
{
  terraformModules: [
    {
      name: "Vpc",
      source: "terraform-aws-modules",
      version: "2.39.0"
    }
  ]
}
```

PR #585, opened 2021-03-03
Don't treat warning lines as resources

### LINKED ISSUE #582: cdktf errors when module generates warnings
### Community Note

### cdktf & Language Versions

cdktf: 0.10
python: 3.7.3
terraform: 0.14.7

### Affected Resource(s)

terraform-google-modules/network/google@3.1.1

### Debug Output

https://gist.github.com/alexjmoore/45bc000eb62a1add1b5214247704e843

### Expected Behavior

When the referenced module generates warnings it should surface these to the user, although I see other issues for better error surfacing, but in this case specifically it shouldn't hard fail at a minimum.

### Actual Behavior

It seems like the warning being generated is causing it to fail - but this is my assumption - see gist referenced above.

### Steps to Reproduce

Use the terraform-google-modules/network/google@3.1.1 - there is a bug in this specific version (latest at the time of writing, but worth pinning to reproduce as there is a pending fix incoming) that generates this warning in typical terraform use:

```
Warning: Interpolation-only expressions are deprecated

  on .terraform/modules/vpc/modules/firewall-rules/main.tf line 18, in resource "google_compute_firewall" "rules":
  18:   for_each                = { for r in var.rules : "${r.name}" => r }
```

If you pin an earlier version of the module like terraform-google-modules/network/google@3.0.0 which doesn't exhibit the warning, then the issue does not occur.

Example python snippet to test:
```
        Network(self, "vpc-test",\
            network_name="vpc-test",
            project_id="<project_id>",
            auto_create_subnetworks=False,
            subnets=[{
                "subnet_name":"sn-test",
                "subnet_ip":"10.0.0.0/24",
                "subnet_region":"europe-west2",
            }]
        )
```

* #0000

jsteinich:
Fixes #582 

Currently any warning that is emitted by terraform during deployment or destruction will be interpreted as a resource which will cause a crash.

The check is added is certainly simplistic, but I'm not sure of a better way to definitely determine if an output line is referring to a resource.

### DIFF
--- a/packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx
+++ b/packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx
@@ -65,7 +65,7 @@ const parseOutput = (str: string): DeployingResource[] => {
         applyState = DeployingResourceApplyState.WAITING
     }
 
-    if (resourceMatch && resourceMatch.length >= 0) {
+    if (resourceMatch && resourceMatch.length >= 0 && resourceMatch[1] != "Warning") {
       return {
         id: resourceMatch[1],
         action: PlannedResourceAction.CREATE,


skorfmann:
> The check is added is certainly simplistic, but I'm not sure of a better way to definitely determine if an output line is referring to a resource.

Not sure either. Either way, gonna have a quick look is there a way to easily reproduce this in a functional test.

PR #590, opened 2021-03-05
Fix broken cdktf-json.md link

wolverian:


jsteinich (approved):
Thanks for correcting this.

PR #592, opened 2021-03-10
Update Terraform versions which are tested against

skorfmann:
This drops tests against `0.12` in favour of testing against the upcoming `0.15` version. That means with this PR being merged we're testing against `0.13`, `0.14`, and `0.15-beta`.

This depends https://github.com/hashicorp/terraform-cdk/pull/591 and will help https://github.com/hashicorp/terraform-cdk/pull/584 being merged. There are some provider schema differences with the change from `0.12` to `0.13` which don't seem worthwhile working around (I think it's purely testing related, functionality wise it's not an issue). 

I think it generally makes sense for now to track the current and prior major release of Terraform core - plus upcoming versions.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        terraform: ["0.12.29", "0.13.4", "0.14.0"]
+        terraform: ["0.13.6", "0.14.7", "0.15.0-beta1"]
     container:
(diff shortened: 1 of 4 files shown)


danieldreier:
Do we check what TF version a user is using? My gut feel here is that we should at least warn people, if not error out, if they are using an older version we don't test against

skorfmann:
Went with a warning only for now - see 2f1cc9a

Screenshot 2021-03-11 at 10 13 35

PR #608, opened 2021-03-21
Fixing aws eks example

GabrielAmazonas:
The following changes were broken on the example:

- Inside the imports folder, with the same cdktf.json, the subfolders created were: aws, vpc and eks instead of terraform_aws_modules.vpc and terraform_aws_modules.eks

- The constructor of the Eks and Vpc objects had their names updated from TerraformAwsModulesEksAws 
and TerraformAwsModulesVpcAws respectively

- The Eks constructor has the cluster_version parameter as obrigatory

### DIFF
--- a/examples/python/aws-eks/Pipfile
+++ b/examples/python/aws-eks/Pipfile
@@ -4,8 +4,8 @@ url = "https://pypi.org/simple"
 verify_ssl = true
 
 [requires]
-python_version = "3.7"
+python_version = "3.8"
 
 [packages]
 constructs = "~=3.0.0"
-cdktf = {path = "./../../../dist/python/cdktf-0.0.0-py3-none-any.whl"}
+cdktf = "*"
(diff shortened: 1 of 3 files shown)


examples/python/aws-eks/Pipfile
```diff
@@ -4,8 +4,8 @@ url = "https://pypi.org/simple"
 verify_ssl = true
 
 [requires]
-python_version = "3.7"
+python_version = "3.8"
 
 [packages]
 constructs = "~=3.0.0"
-cdktf = {path = "./../../../dist/python/cdktf-0.0.0-py3-none-any.whl"}
```
skorfmann, line 11: What's the reason to remove the path to `dist` here? The examples should be built with the artifacts generated by `yarn build && yarn package`.

skorfmann (changes requested):
Thanks - the changed module names are due to the recent module related changes.

examples/python/aws-eks/Pipfile
GabrielAmazonas: Hi @skorfmann, thanks for pointing out - this change was unintended - Added a reversion commit

guan840912 (approved):

skorfmann (approved):
Cool, thanks!

PR #614, opened 2021-03-31
Add hyphens to resource id regexp in output parser

sergeyzenchenko:
According to Terraform docs hyphens are allowed in resource ids https://www.terraform.io/docs/language/syntax/configuration.html#identifiers

Current version doesn't display apply status for 80% of resources in my deployment because of this bug.


sergeyzenchenko:
Also, according to docs "The first character of an identifier must not be a digit, to avoid ambiguity with literal numbers." I can add fix for this one in this PR if you want. Or should I make separate one?

skorfmann:
Cool, thanks for the fix! 

> Also, according to docs "The first character of an identifier must not be a digit, to avoid ambiguity with literal numbers." I can add fix for this one in this PR if you want. Or should I make separate one?

Gonna merge this now, would be great if you could create a new PR :)

PR #617, opened 2021-04-01
Update resource regexp to match terraform spec

sergeyzenchenko:
Update resource regexp to match terraform spec

### DIFF
--- a/packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx
+++ b/packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx
@@ -36,7 +36,7 @@ const parseOutput = (str: string): DeployingResource[] => {
     if (/^Outputs:/.test(line)) { return }
     if (/^data\..*/.test(line)) { return }
 
-    const resourceMatch = line.match(/^([a-zA-Z_\-][a-zA-Z\d_\-.]*):/)
+    const resourceMatch = line.match(/^([a-zA-Z_][a-zA-Z\d_\-.]*):/)
     let applyState: DeployingResourceApplyState;
 
     switch (true) {
--- a/packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx
+++ b/packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx
@@ -36,7 +36,7 @@ const parseOutput = (str: string): DeployingResource[] => {
     if (/^Outputs:/.test(line)) { return }
     if (/^data\..*/.test(line)) { return }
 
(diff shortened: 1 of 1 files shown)


packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx
```diff
@@ -36,7 +36,7 @@ const parseOutput = (str: string): DeployingResource[] => {
     if (/^Outputs:/.test(line)) { return }
     if (/^data\..*/.test(line)) { return }
 
-    const resourceMatch = line.match(/^([a-zA-Z\d_\-.]*):/)
+    const resourceMatch = line.match(/^([a-zA-Z][a-zA-Z\d_\-.]*):/)
```
jsteinich, line 39: The spec says the first character isn't a digit, but `-` and `_` may be allowed (could test to confirm).

jsteinich (commented):
Thank you for your contribution. Can you confirm if other non-letters are allowed? Making this regex more strict may eliminate false matches, but we haven't had many cases reported so I think we should be cautious here.

sergeyzenchenko:
I can add unit test for that function. This regexp is build according to terraform spec

jsteinich:
I just did a quick test and `_` is allowed for the first character of a resource identifier. `-` is not.
Error message from testing: "A name must start with a letter or underscore and may contain only letters,
digits, underscores, and dashes."

skorfmann:
> I just did a quick test and `_` is allowed for the first character of a resource identifier. `-` is not.
> Error message from testing: "A name must start with a letter or underscore and may contain only letters,
> digits, underscores, and dashes."

Would be great if you could adjust it accordingly @sergeyzenchenko

sergeyzenchenko:
Yes! I've already found that _ should be allowed. I will push update it today

packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx
skorfmann: So, like this? https://github.com/hashicorp/terraform-cdk/pull/617/commits/0e7c40d75fad6d77d563775c37112720e0e524d1

jsteinich: It actually can't start with a dash.
"A name must start with a letter or underscore and may contain only letters,
digits, underscores, and dashes."

skorfmann: Oh, thanks for clarifying :)

Issue #115, opened 2020-07-01
Handle small terminal sizes with lots of resources to display


skorfmann:
Due to the nature of the text rendering, using the `cdktf` cli for stacks with lots of resources in small terminal windows leads to lots of scrollback and flickering, broken text output.

Essentially, we'll have to adjust the output size dynamically to the terminal size. 

A few ideas:

- Define a fixed size for our canvas based on the terminal dimensions, and make sure elements exceeding the size won't be rendered. This would need us to implement some kind of scrolling mechanism
- Generally reduce dynamic elements shown on the screen, and use static text where possible
- Calculate canvas size and cut off overflowing elements

skorfmann:
> > Generally reduce dynamic elements shown on the screen, and use static text where possible
> 
> I think that's the only viable option for now.
> 
> * Resources in progress would be displayed "live"
> * Resources which got applied will be rendered as static text above the "live" view
> * When everything is done, we could render a final summary (similar to `diff`)
> 
> To keep it simple, this should be the only behaviour we support for now (so, no dynamic change of rendering modes based on terminal size and resource count). Thoughts?

This could look like this and use [static elements](https://github.com/vadimdemedes/ink/tree/259edd47968081b23b176b7a7cfecc5e5757845d#static)

```
Stack: hello-terra
Done: (a static list)
 ✔ AWS_DYNAMODB_TABLE   helloterra_Hello    aws_dynamodb_table.helloterra_Hello_698
 ✔ AWS_SNS_TOPIC        helloterra_Topic0   aws_sns_topic.helloterra_Topic0_E9AB776
 ✔ AWS_SNS_TOPIC        helloterra_Topic10  aws_sns_topic.helloterra_Topic10_F78BE0
 ✔ AWS_SNS_TOPIC        helloterra_Topic11  aws_sns_topic.helloterra_Topic11_219D04
 ✔ AWS_SNS_TOPIC        helloterra_Topic12  aws_sns_topic.helloterra_Topic12_19C0FD
 ✔ AWS_SNS_TOPIC        helloterra_Topic13  aws_sns_topic.helloterra_Topic13_FE30A3
 ✔ AWS_SNS_TOPIC        helloterra_Topic14  aws_sns_topic.helloterra_Topic14_E676CF
 ✔ AWS_SNS_TOPIC        helloterra_Topic15  aws_sns_topic.helloterra_Topic15_1E2290
[...]

PR #618, opened 2021-04-01
Improve activity indicator performance

sergeyzenchenko:
Current implementation based on ink-spinner consumes a lot of CPU if many indicators are displayed. 
There are few reasons for that.

1. Internally ink-spinner sets JS interval for each active spinner. This approach leads to constant re-render of spinners. These re-renders are not synced with each other and for each small update React and Ink need to do bunch of work. 
2. ink-spinner API doesn't allow us to specify spinner update rate. Default "dots" type sets update interval to 80ms. It's 12 updates per seconds for each spinner rendered in terminal.

How do I fixed it?

1. I replaced ink-spinner with custom version that sets less frequent update interval.
2. Instead of setInterval for each active spinner I use shared SpinnerClock that holds 1 interval for all active spinners. Now instead of tons of small updates we have 1 centralised update for all spinners. It's significantly reduced CPU usage.

I replaced default "dots" spinner with "circleQuarters" type. 
Example: ```◵ synthesizing ...``` 
It will rotate this clock like circle every 500ms.

I think it's good enough for terminal based activity indication.

I've tested it on list of 54 resource with activity indicator displayed next to each of them.
CPU usage dropped from 50-60% to 4-5% on my machine.

P.S.
I've disabled 2 ESLint rules related to React
react/prop-types: This rule force developer to write propType. PropTypes can be skipped for TS based projects.
react/display-name: This rule force developer to write displayName for functional components. It's helpful in large projects but I think it pretty useless in such small UI project like cdktf-cli. When was the last time when you debugger terminal UI from Google Chrome React inspector?)

### DIFF
--- a/packages/cdktf-cli/bin/cmds/ui/components/activity-indicator.tsx
+++ b/packages/cdktf-cli/bin/cmds/ui/components/activity-indicator.tsx
@@ -1,0 +1,83 @@
+import React, {useEffect, useState} from "react";
+import {Text} from "ink";
+import spinners from 'cli-spinners';
+
+interface Props {
+    readonly color?: string;
+}
+
+const DEFAULT_INDICATOR_COLOR = "green";
+
+type Unsubscribe = () => void;
+type Subscription = () => void;
+
+class SpinnerClock {
+
+    private subscriptions: Subscription[] = [];
+    private interval?: NodeJS.Timeout;
+
+    constructor(private spinRate: number = 1000) {
+    }
+
+    subscribe(subscription: Subscription): Unsubscribe {
+
+        this.subscriptions.push(subscription);
+
+        this.start();
+
+        return () => {
+            this.removeSubscription(subscription);
+        };
+    }
+
+    private start() {
+        if (!this.interval) {
+            this.interval = setInterval(() => {
+                this.notify();
+            }, this.spinRate);
+        }
+    }
+
+    private stop() {
+        if (this.interval) {
+            clearInterval(this.interval);
+            this.interval = undefined;
+        }
+    }
+
+    private notify() {
+        for (const subscription of this.subscriptions) {
+            subscription()
+        }
+    }
+
+    private removeSubscription(subscription: Subscription) {
+        this.subscriptions = this.subscriptions.filter((sub) => sub !== subscription);
+
+        if (this.subscriptions.length === 0) {
+            this.stop();
+        }
+    }
+}
+
+const sharedSpinnerClock = new SpinnerClock();
+
+const Spinner = () => {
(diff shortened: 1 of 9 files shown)


sergeyzenchenko:
Part of snapshot tests are failing because of different indicator symbol. I can update them, but I need to understand if you ok with using "◵". The old one with dots looks weird when updated 2 times per second instead of 12 times

skorfmann:
Wow, that's pretty cool! Thank for the PR - I'll have a look now.

> I've tested it on list of 54 resource with activity indicator displayed next to each of them.
> CPU usage dropped from 50-60% to 4-5% on my machine.

That's awesome!

> Part of snapshot tests are failing because of different indicator symbol. I can update them, but I need to understand if you ok with using "◵". The old one with dots looks weird when updated 2 times per second instead of 12 times

I'll have a look.

sergeyzenchenko:
@skorfmann There is not so much to look at. Tests just need to be updated. I can do it. If you okay with using  "◵" as indication.

skorfmann:
So, I had more in-depth a look at this. In general I really like the direction. For the `circleQuarters`, I'm not convinced that this makes a huge difference compared to the dots (did a few quick local tests) and I don't like the aesthetics of it :) In any case, I think an interval of `1000` or even `500` looks a bit odd. It starts to get easier on the eye at `250` and even more so at `125`. 

So, that I'd do: Keep the `dots` at an interval of `125`. 

A bit more background to keep in mind: For cdktf apps with many resources, there are other issues as well. Primarily scrollback issues (see #115). I'm very certain, that solving the scrollback issues, will result in somehow limiting the resources being rendered and therefore have a positive impact on the performance issues as well.

sergeyzenchenko:
I am looking at scrolling issues too :) we have more resources than my screen height. Possibly it can be related to unnecessary react updates. I've removed unnecessary updates and testing how it works. Will submit PR if it will make difference

skorfmann:
> I am looking at scrolling issues too :) we have more resources than my screen height. Possibly it can be related to unnecessary react updates. I've removed unnecessary updates and testing how it works. Will submit PR if it will make difference

For the scrollback issues, I just added an example of how this could be addressed https://github.com/hashicorp/terraform-cdk/issues/115#issuecomment-817635293

DanielMSchmidt:
Hey @sergeyzenchenko, thank you for the PR! I checked it and it looks a bit broken with such a small spin rate, increasing it lead to similar CPU usage as before. The UX for lots of changing resources is broken currently and we are planning to redo the UI soonish with the goal to reduce CPU consumption.
I'm closing this for now, thank you for your help!

Issue #403, opened 2020-10-05
Creation status doesn't update in the console for modules


vishmit123:
When [trying](url) to create an AWS vpc using vpc module, the create status is not updated onto the console. Even though the resources are created and reflecting in the state file.

<img width="622" alt="Screenshot 2020-10-05 at 4 38 11 PM" src="https://user-images.githubusercontent.com/17925519/95072928-a3514d00-0729-11eb-8834-a5c9ebb92eda.png">

skorfmann:
Not sure if that's a regression, or if resources in modules weren't working so far. Either way, that's something to be fixed :)

alexjmoore:
To add, I am seeing the same when using the google network module, I don't see any status updates and the summary counts for both deploy and destroy is always 0, yet it is creating and deleting resources.

sergeyzenchenko:
@alexjmoore probably should be fixed by #619

ansgarm:
Hi @vishmit123, Hi @alexjmoore

We just released version [`0.2.2`](https://github.com/hashicorp/terraform-cdk/releases/tag/v0.2.2) including the fix from @sergeyzenchenko (#619).

Could you check if this resolves this issue?

alexjmoore:
From a quick check I think it looks better thanks, but I'm also waiting on #630 to allow modules i'm using to work again in order to test fully.

danieldreier:
@alexjmoore #630 got merged - have you had a chance to test again since then?

charles-salmon:
@danieldreier I've just seen behaviour that appears to at least be _similar_ to the issue that was described here.

I have defined a stack which creates resources belonging to two modules:

* `terraform-aws-modules/acm/aws`: I have referenced this in my `cdktf.json` file, and executed `cdktf get` to generate TypeScript classes. This module reference is named `acm`.
* `cloudposse/cloudfront-s3-cdn/aws`: I have referenced this by defining a `TerraformHclModule`, because the module defines an input variable of type `set(string)` (see #928). This module reference is named `cdn`.

The output that I see after executing `cdktf deploy` is below:

![image](https://user-images.githubusercontent.com/28038653/140631484-6c9d3f0c-df89-49d5-bfed-0f6768c53fb4.png)

That is, the status of the module creation was not updated, the modules are not correctly listed (the `cdn` module is listed twice), and the summary shows that 0 resources were created, despite the fact that this was a first-time deployment in the workspace.

jsteinich:
[...]

PR #619, opened 2021-04-02
Add missing handling for resource status lines

sergeyzenchenko:
Still creating... and Still destroying... were not handled. If resource took long enough time to create or destroy it went from "Creating" state to "Waiting" state.

Probably cause of https://github.com/hashicorp/terraform-cdk/issues/403

### DIFF
--- a/packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx
+++ b/packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx
@@ -41,6 +41,7 @@ const parseOutput = (str: string): DeployingResource[] => {
 
     switch (true) {
       case /Creating.../.test(line):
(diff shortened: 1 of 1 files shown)


jsteinich:
Thank you for your contribution.

Do you have evidence that this fixes #403? I would have expected the switch to hit the default waiting state which I would anticipate eventually resulting in one of the completed states.

sergeyzenchenko:
I don't have clear evidence for this specific case. But in my deployment resource were in waiting state until they eventually all deployed. It looks similar, but information in in issue is not enough to fully understand it.

sergeyzenchenko:
Also I've observed same behaviour with Summary area not displaying correct numbers. After this fix I see all updates and summary properly

skorfmann (approved):

PR #624, opened 2021-04-08
Improve error handling if synthesizing fails

### LINKED ISSUE #533: Better Error Message when "app" command throws an error
The `app` configuration in the [cdktf.json](https://github.com/hashicorp/terraform-cdk/blob/35ef68e65bcaf5bee6f1ed1ab4c24c9bdf965d29/docs/working-with-cdk-for-terraform/cdktf-json.md) file defines how an app / stack is [synthesized](https://github.com/hashicorp/terraform-cdk/blob/35ef68e65bcaf5bee6f1ed1ab4c24c9bdf965d29/packages/cdktf-cli/bin/cmds/helper/synth-stack.ts#L19). 

There's no error handling around this. It would be good to handle errors and provide actionable feedback. 

I stumbled up on this when I renamed the file without changing the command. Took me a few seconds to realise what the issue is :) 

Screenshot 2021-01-31 at 16 09 07

ansgarm:
Resolves #533 by printing a more fitting error message including the command that was used for synth and the output that command wrote to stderr.

Before this change:
Screenshot 2021-04-08 at 14 39 23

After this change:
Screenshot 2021-04-08 at 14 29 24
(`index.js` did not exist)

And without anything on stderr:
Screenshot 2021-04-08 at 14 30 19

### DIFF
--- a/test/test-helper.ts
+++ b/test/test-helper.ts
@@ -43,7 +43,7 @@ export class TestDriver {
   }
 
   synth = () => {
-    execSync(`cdktf synth`, { stdio: "inherit", env: this.env });
+    execSync(`cdktf synth`, { stdio: "pipe", env: this.env });
   }
 
   diff = () => {
--- a/test/test-helper.ts
+++ b/test/test-helper.ts
@@ -11,6 +11,16 @@ export class TestDriver {
     this.env = Object.assign({CI: 1}, process.env, addToEnv);
   }
 
+  private execSync(command: string) {
+    try {
+      execSync(command, { stdio: "pipe", env: this.env });
+    } catch(e) {
+      console.log(e.stdout.toString())
+      console.error(e.stderr.toString())
+      throw e;
+    }
+  }
+
   switchToTempDir = () => {
     const pathName = path.join(os.tmpdir(), "test");
     this.workingDirectory = fs.mkdtempSync(pathName);
@@ -43,7 +53,7 @@ export class TestDriver {
   }
 
   synth = () => {
-    execSync(`cdktf synth`, { stdio: "pipe", env: this.env });
+    this.execSync(`cdktf synth`);
   }
 
   diff = () => {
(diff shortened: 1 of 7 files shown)


ansgarm:
marked this as draft, as I'm going to add an integration test

skorfmann:
Looks great! Much easier to see what's going on 👍

test/test-helper.ts
```diff
@@ -43,7 +43,7 @@ export class TestDriver {
   }
 
   synth = () => {
-    execSync(`cdktf synth`, { stdio: "inherit", env: this.env });
+    execSync(`cdktf synth`, { stdio: "pipe", env: this.env });
```
skorfmann, line 46: does this affect other existing tests in case of an error?

ansgarm: Good catch. Just had a look, it swalled the syntax error I added for testing..
I'll see what I can do to make sure the child process keeps printing to stdout / err.

ansgarm:
Just added another commit that makes sure stdout and stderr of the child process are printed if that child process failed with an error. This actually improves the current implementation as the syntax error (which I used for testing) was not visible in the current implementation.

before the new commit (with a syntax error in the `main.ts` of the CDK stack under test):
Screenshot 2021-04-09 at 17 41 15

with the new commit:
Screenshot 2021-04-09 at 16 53 04

I'm a bit unsure about the minor drawback of having stdout and stderr separate and not in order as the logs appeared.  
Secondly, I might want to change the logging of the stdout to `console.log()` instead of `console.error()` because jest prints the location where `console.error()` was called which adds some noise.  
However, in the end these are just tests and not the behavior of the cdktf-cli itself 😅 

Thoughts, @skorfmann?

skorfmann:
> I'm a bit unsure about the minor drawback of having stdout and stderr separate and not in order as the logs appeared.
> Secondly, I might want to change the logging of the stdout to `console.log()` instead of `console.error()` because jest prints the location where `console.error()` was called which adds some noise.
> However, in the end these are just tests and not the behavior of the cdktf-cli itself 😅

Sounds good to me. If it really turns out to be an issue, we can still iterate later.

skorfmann (approved):

Issue #488, opened 2021-01-13
Handle Not Existing Org / Workspaces in Terraform Cloud


skorfmann:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

When using Terraform Cloud as [remote](https://www.terraform.io/docs/cloud/run/index.html) runner, `cdktf` will zip the `./cdktf.out` folder contents and uploads it to the TerraformCloud API. The entire `diff` / `deploy` / `destroy` cycle is then running remotely. That's done in via the [TerraformCloud](https://github.com/hashicorp/terraform-cdk/blob/35ef68e65bcaf5bee6f1ed1ab4c24c9bdf965d29/packages/cdktf-cli/bin/cmds/ui/models/terraform-cloud.ts) handler within the `cdktf-cli` package, which is relying on this [api wrapper](https://github.com/skorfmann/terraform-cloud). All interactions with the Terraform Cloud API are relying on valid credentials and correct configuration for the organization / workspace. 

The organization and workspace configuration is [sourced](https://github.com/hashicorp/terraform-cdk/blob/35ef68e65bcaf5bee6f1ed1ab4c24c9bdf965d29/packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx#L238) from the synthesized stack. When one the values is misconfigured, it looks like the raw error from `axios` is being printed out. This should be more expressive and actionable to the user.

![Screenshot 2021-01-13 at 14 48 18](https://user-images.githubusercontent.com/136789/104460544-850fae80-55ae-11eb-92bb-796c172117ed.png)

I think it would make sense to add some basic error handling to the [TerraformCloud](https://github.com/hashicorp/terraform-cdk/blob/35ef68e65bcaf5bee6f1ed1ab4c24c9bdf965d29/packages/cdktf-cli/bin/cmds/ui/models/terraform-cloud.ts) class, which takes care of making the errors more expressive.

### References

- #449

ansgarm:
Resolved via #626

PR #626, opened 2021-04-12
improve error message for 404 errors whilst querying Terraform Cloud …

ansgarm:
…workspace

Before:
Screenshot 2021-04-12 at 15 28 54

After:
Screenshot 2021-04-12 at 15 28 18

### DIFF
--- a/packages/cdktf-cli/bin/cmds/ui/models/terraform-cloud.ts
+++ b/packages/cdktf-cli/bin/cmds/ui/models/terraform-cloud.ts
@@ -9,9 +9,9 @@ import archiver from 'archiver';
 import { WritableStreamBuffer } from 'stream-buffers';
 
 export class TerraformCloudPlan implements TerraformPlan {
-  constructor(public readonly planFile: string, public readonly plan: {[key: string]: any}, public readonly url: string) {}
+  constructor(public readonly planFile: string, public readonly plan: { [key: string]: any }, public readonly url: string) { }
 
-  public get resources(): PlannedResource[]  {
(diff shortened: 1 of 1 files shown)


ansgarm:
@skorfmann in #488 you mentioned also handling other possible errors. Do you have some example of how to reproduce some?
The response seems to contain some well formed error object which could be printed if any 4xx or 5xx error happens.
Unfortunately it does not differentiate whether the organizationName or workspace caused the 404 so the wording of the error message states both at possibly incorrect.

skorfmann (approved):
Looks good - let's keep it simple and get it merged 👍

ansgarm:
Example for printing the whole error object because its structure differs between endpoints / methods:
Screenshot 2021-04-12 at 18 11 37

packages/cdktf-cli/bin/cmds/ui/models/terraform-cloud.ts
```diff
@@ -64,7 +64,30 @@ const wait = (ms = 1000) => {
   });
 }
 
-export class TerraformCloud implements Terraform  {
+function BeaufifyErrors(target: any, propertyKey: string, descriptor: PropertyDescriptor) {
```
skorfmann, line 67: This should probably be `BeautifyErrors`

ansgarm: 🙈

skorfmann:
Cool - let's ship it 👍

PR #633, opened 2021-04-14
Fix module without newline

### LINKED ISSUE #595: cdktf unable to parse modules without newlines at EOF
### Community Note

### cdktf & Language Versions

using `typescript@4.2.3` and `cdktf 0.2.0`

### Affected Resource(s)

cdktf importing of custom modules

### Debug Output

i'm using the new module improvements from #584 to source modules locally and from github. It is works as expected unless a module's variables and/or outputs tf files do not have a newline at EOF. cdktf fails to parse those modules. Add a newline to EOF for suspect files and then cdktf will successful parse and import the module.

[error gist](https://gist.github.com/celliott/9be823e114e6ab9a32c7de91102b0f8d)

**example terraformModules**

```
"terraformModules": [
  {
    "name": "s3",
    "source": "git@github.com:mycompany/terraform-aws-s3.git"
  }
],
```

### Expected Behavior

I would expect to be able to import a module that works without error when using tarraform directly. It's easy to fix but it wasn't clear at first what the issue was. If there is no fix, then better documentation and error messaging would help.

### Actual Behavior

A source a custom tf module with the improved module importing and any module that doesn;t have new lines at EOF fail to parse. I have tested with both local or github sourced modules and as expected, the behavior is the same.

### Steps to Reproduce

Source a module from local or github using `terraformModules` in `cdktf.sjon`. Make sure that the file containing either `outputs` or `variables` does NOT have a newline at EOF. Run `cdktf get` and the module with fail to import. Add a newline to the end of the file(s), reimport and it will work as expected.

jsteinich:
Fixes #595

Also picked up no variable module support since the test case I was working with didn't have any.


packages/cdktf-cli/lib/get/generator/provider-schema.ts
```diff
@@ -75,34 +75,36 @@ interface ModuleIndex {
...
-  for (const name of Object.keys(variables)) {
-    const variable = variables[name][0]
-    let variableType: string;
-    // eslint-disable-next-line no-prototype-builtins
-    if (variable.hasOwnProperty('type') == false && variable.hasOwnProperty('default') == true) {
-      switch (typeof variable['default']) {
-        case "boolean": variableType = 'bool' ; break;
-        case "number": variableType = 'number' ; break;
-        default: variableType = 'any';
+  if (variables) {
```
skorfmann, line 78: I think a guard clause would be nicer here, something like `if (!variables) return result;`  Or are there other changes than this if clause?

jsteinich: That was the only change. I switched to a guard, though that did require being more explicit about type.

skorfmann: Cool - thanks 👍

skorfmann (approved):

PR #640, opened 2021-04-15
rename instance variable to a more specific name to avoid name collisions

### LINKED ISSUE #634: Property name and type conflict between generated code and cdktf.ComplexComputedList
### Community Note

### cdktf & Language Versions

- cdktf: 0.2.1
- typescript: 4.2.4

### Debug Output

https://gist.github.com/yamamoto-febc/2a29df295eee3a768d8f1259747473bd

### Expected Behavior

It should compile.

### Actual Behavior

Unable to compile with TS2416 and TS2411 errors.

### Steps to Reproduce

Generate `sacloud/sakuracloud` provider bindings and compile - See here for a ready to go example: https://github.com/yamamoto-febc/cdktf-sakuracloud-example

---

In the community provider [sacloud/sakuracloud](https://registry.terraform.io/providers/sacloud/sakuracloud/latest), there is a field with the key name `index`, `computed:true` and `TypeInt`.  
ref: https://github.com/sacloud/terraform-provider-sakuracloud/blob/4063b51a35a88163163b0c70cafd86702543863d/sakuracloud/data_source_sakuracloud_vpc_router.go#L101-L105

Here is the code generated using it: https://github.com/yamamoto-febc/cdktf-sakuracloud-example/blob/987794ad53d57fb596a15181d5809f06976fd4b9/.gen/providers/sakuracloud/data-sakuracloud-vpc-router.ts#L166-L169

The Class which extended `cdktf.ComplexComputedList` was generated, and a getter named `index` was generated.
In the original provider, it is defined as TypeInt, so it returns a number.

However, in cdktf.ComplexComputedList, `index` is defined as a string, which causes an error.

ansgarm:
Resolves #634 which was caused by a Terraform data source which had an attribute containing a list of items which each have a key `index` which collided with its base class `ComplexComputedList`.

### DIFF
--- a/packages/cdktf-cli/test/get/generator/module-generator.test.ts
+++ b/packages/cdktf-cli/test/get/generator/module-generator.test.ts
@@ -6,7 +6,7 @@ import { TerraformModuleConstraint  } from '../../../lib/config'
 import { expectModuleToMatchSnapshot } from '../util';
 
 test('generate some modules', async () => {
-  jest.setTimeout(20000)
+  jest.setTimeout(60000)
 
   const workdir = fs.mkdtempSync(path.join(os.tmpdir(), 'module-generator.test'));
   const constraint = new TerraformModuleConstraint('terraform-aws-modules/eks/aws@7.0.1')
--- a/packages/cdktf-cli/test/get/read-schema.test.ts
+++ b/packages/cdktf-cli/test/get/read-schema.test.ts
@@ -4,7 +4,7 @@ import { TerraformModuleConstraint, TerraformProviderConstraint } from "../../li
 
 describe("readSchema", () => {
   beforeAll(() => {
-    jest.setTimeout(30000)
+    jest.setTimeout(40000)
   })
 
   it("generates a single provider schema", async () => {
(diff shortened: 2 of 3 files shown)


packages/cdktf-cli/test/get/generator/module-generator.test.ts
```diff
@@ -6,7 +6,7 @@ import { TerraformModuleConstraint  } from '../../../lib/config'
 import { expectModuleToMatchSnapshot } from '../util';
 
 test('generate some modules', async () => {
-  jest.setTimeout(20000)
+  jest.setTimeout(60000)
```
skorfmann, line 9: Mostly out of curiosity: Do you know why the timeout has to be that high?

ansgarm: I don't. But I had other tests failing with 40s so I figured to just give it a bit more.
But I'm happy to "downgrade" it.

skorfmann: All good 👍

skorfmann (approved):
:shipit:

ansgarm:
🚀

PR #645, opened 2021-04-19
feat: add support for remote templates

### LINKED ISSUE #339: Remote Templates
### Community Note

### Description

The [cdktf-cli](https://github.com/hashicorp/terraform-cdk/tree/main/packages/cdktf-cli) ships with [templates](https://github.com/hashicorp/terraform-cdk/tree/b5d06f879fb585142487de711042528e3f3f9be9/packages/cdktf-cli/templates) to bootstrap a project via `cdktf init --template [template name]`. Right now we have the following templates:

- csharp
- java
- python
- python-pip
- typescript

Since these templates are bundled with the `cdktf-cli` we should do as little assumptions as possible in these templates (e.g. we don't want to pre-define a Terraform module or provider in the `cdktf.json` files). We already maintain two different template for Python, which demonstrates that there are different opinions about setting up even a very basic project. Since we shouldn't maintain all possible ways of setting up a project in one of the target languages, we want to enable users to define their own templates.

### User Defined Template Sources

We should enable the user to pull templates from a URI, e.g. `cdktf init --template https://gist.github.com/foo/bar` or  `cdktf init --template file:///foo/bar`

### User Defined Template Format

The goal for user defined templates from my point of view:

- Easy to create, maintain and consume a template for users
- Keep our code base simple

#### Current State 

When looking at one of the [examples](https://github.com/hashicorp/terraform-cdk/tree/b5d06f879fb585142487de711042528e3f3f9be9/packages/cdktf-cli/templates/typescript), it's an entire directory which will be copied in the current working directory of the `cdktf-cli`. That's done via [sscaff](https://github.com/awslabs/node-sscaff) where the [.hooks.sscaff.js](https://github.com/hashicorp/terraform-cdk/blob/b5d06f879fb585142487de711042528e3f3f9be9/packages/cdktf-cli/templates/typescript/.hooks.sscaff.js) file contains `pre` / `post` hooks to be executed. 

#### Zip 

One option would be to allow a zip file (and / or similar types) which follows a similar structure as the embedded templates. At least Github and Gitlab are supporting to download a repository or gist as archived file in various formats. From a user point of view, I think it'd be pretty straightforward to take a local folder, zip it and upload it wherever it's needed.

#### Other options?

?

##### Single JS Generator File

A slimmed down version of what [Ruby on Rails App templates](https://guides.rubyonrails.org/rails_application_templates.html) are doing could be an option. Compared to the `zip` file approach, the implemenation would likely be more complex. On the plus side, it would be easier to check for a user what's going to happen by just looking at the file.

With the goal of simplicity in mind, I'd try to avoid building yet another template generator.

##### Projen 

Also thought briefly about using [projen](https://github.com/projen/projen) in this context. However, `projen` is explicitly not just a template generator, but actively maintains the generated files. Hence, it's not what we're looking for in this context.

ansgarm:
Currently supports specifying a public url pointing to a zip file.

Refactored some test helper methods to be async because else they block the js main loop and the template-server utility class could not serve the request for a template zip file.

Screenshot 2021-04-20 at 00 46 48

Closes #339.

To test this locally, you can use this url to a copy of our typescript template:
```
https://github.com/ansgarm/cdktf-template-example/archive/refs/tags/v0.0.1.zip
```

# Open tasks
- [x] Finish writing docs (see todos in markdown file)
- [x] Create follow up issue to add a full example / tutorial of how to create a remote template to the docs -> #652
- [x] Check for possible issues with templates not using cdktf versions we supply as inputs to `sscaff`


packages/cdktf-cli/lib/util.ts
```diff
@@ -91,4 +92,38 @@ export async function readCDKTFVersion(outputDir: string): Promise<string> {
...
+export class HttpError extends Error {
+  constructor(message?: string, public statusCode?: number) {
+    super(message); // 'Error' breaks prototype chain here
+    Object.setPrototypeOf(this, new.target.prototype); // restore prototype chain
+    // see: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-2.html#support-for-newtarget
+  }
+}
+
+export async function downloadFile(url: string, targetFilename: string): Promise<void> {
+  const client = url.startsWith('http://') ? http : https;
```
skorfmann, line 106: I think it would be good to support file URI as well, so something like `file:///foo/bar/archive.zip`

skorfmann (commented):
left a few minor comments, but looks good!

docs/working-with-cdk-for-terraform/remote-templates.md
```diff
@@ -0,0 +1,32 @@
...
+### Remote Templates
+Currently the `cdktf` supports downloading and extracting a zip archive containing the files for the template. When extracting the archive, it searches for the `cdktf.json` file. If that file cannot be found in the root directory, it walks all directories until it finds the file. This allows creating an archive that contains e.g. a `README.md` in the root directory explaining things which itself won't turn up in the created project directory. However, most templates won't make use of it.  
+
+If you're using a Github repository for your template, you can create URLs to your repo as follows
+#### main branch
+`https://github.com/<user or organization>/<repo>/archive/refs/heads/main.zip`  
+#### tag `v.0.0.1`
+`https://github.com/<user or organization>/<repo>/archive/refs/tags/v0.0.1.zip`
+
+**Please Note:** Currently only public accessible zip archives are supported. If you need support for private packages, please [file an issue](https://github.com/hashicorp/terraform-cdk/issues/new?labels=enhancement%2C+new&template=feature-request.md).
```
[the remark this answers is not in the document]
ansgarm: Doesn't basic auth need a header to be specified? Or is there a way to encode that into urls?

packages/cdktf-cli/bin/cmds/init.ts
```diff
@@ -185,31 +192,117 @@ If you want to exit, press {magenta ^C}.
...
+    }
   }
+}
 
-  const templatePath = path.join(templatesDir, templateName);
+async function fetchRemoteTemplate(templateUrl: string): Promise<Template> {
+  console.log(chalkColour`Fetching remote template from: {whiteBright ${templateUrl}}`);
+  try {
+    const url = new URL(templateUrl);
+    const remoteFileName = url.pathname.substring(url.pathname.lastIndexOf('/') + 1) || 'template.zip';
```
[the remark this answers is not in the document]
ansgarm: It is 👍

packages/cdktf-cli/bin/cmds/init.ts
```diff
@@ -185,31 +192,117 @@ If you want to exit, press {magenta ^C}.
...
+
+    if (!templatePath) {
+      throw new Error(chalkColour`Could not find a {whiteBright cdktf.json} in the extracted directory`);
+    }
 
-  return {
-    'Name': templateName,
-    'Path': templatePath
+    return {
+      'Name': remoteFileName.endsWith('.zip') ? remoteFileName.substring(0, remoteFileName.length - 4) : remoteFileName,
```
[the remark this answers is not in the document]
ansgarm: 👍 implemented

packages/cdktf-cli/lib/util.ts
ansgarm: I created a follow up issue for this enhancement #653

test/csharp/synth-app/test.ts
```diff
@@ -8,13 +8,13 @@ import { TestDriver } from "../../test-helper";
 describe("csharp full integration test synth", () => {
   let driver: TestDriver;
 
-  beforeAll(() => {
+  beforeAll(async () => {
     driver = new TestDriver(__dirname)
-    driver.setupCsharpProject()
-  });
+    await driver.setupCsharpProject()
+  }, 180_000);
```
skorfmann, line 14: would it be possible to define a global timeout for these `beforeAll` hooks?

ansgarm: Yeah, we should probably override that globally somewhere 🙈

skorfmann (approved):
Looks great! 👍

PR #673, opened 2021-05-03
fix(docs): rename `aws` to `aws-multiple-stacks` and add `aws-prebuilt` link

neilkuan:
fix(docs): remove readme not exist path
- Typescript
   -  aws
截圖 2021-05-03 上午10 35 32

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,8 @@ Choose a language:
 
 ### Typescript
 
+* [aws-prebuilt](./examples/typescript/aws-prebuilt)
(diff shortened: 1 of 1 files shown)


jsteinich:
Looks like this example was renamed to `aws-multiple-stacks`. The `aws-prebuilt` example might be a better fight for the first typescript example rather than just removing.

neilkuan:
ok, I will update it, thanks @jsteinich  👍

skorfmann (approved):
Thanks :)

PR #691, opened 2021-05-07
Stack: ignore state files

DanielMSchmidt:
Previously the files were not successfully ignored


danieldreier:
When I tested this (just by coping the gitignore line into a local one, my tfstate files were not ignored:

```
Untracked files:
  (use "git add <file>..." to include in what will be committed)
        terraform.tfstate
        terraform.tfstate.backup
```

ansgarm (approved):
Just tested it successfully with a project that was initialized with cdktf 0.3 (supporting multiple stacks).
As these are the templates for initializing new projects, this PR won't affect existing projects such as that from @danieldreier.

PR #692, opened 2021-05-07
fix integration test if no stack was supplied to command

ansgarm:
Related to #689. I don't know why that branch was green on the PR but not after being merged.. I thought integration tests would run when submitting PRs 🤔


DanielMSchmidt (approved):

jsteinich:
> I thought integration tests would run when submitting PRs

They do run for PRs.
Sometimes a non-conflicting file (from git's point of view) can cause tests to fail if the branch wasn't completely up to date with main.
Could also be an issue in how the tests are ran on main. Looks like the parallelization work wasn't done in that workflow.

Issue #690, opened 2021-05-07
Automatically run get when running deploy / destroy / plan


DanielMSchmidt:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

When I first clone a new project and try out the examples I run into an error as I did not run get before. If I run get every time before the mentioned commands I get errors thrown if there are only prebuilt providers.

jsteinich:
I don't think it should be ran every time. It is slow and could cause unexpected changes when there isn't an exact version pinned.

If changes were detected / generated code was missing than it could make sense. Some options in increasing complexity:

1. Only run if no code generated
2. Run if no code for any provider / module
3. Run if version changed for any provider / module

Tangential would be running incremental get operations (I thought there was an issue about that already).

DanielMSchmidt:
Sorry for the late answer:

I agree, I think the options you outlined are a great roadmap 👍

skorfmann:
We used to have a basic check that checks for the presence of generated code folders in case there was something specified in `cdktf.json`. Looks like we might have lost that check as part of one of the last refactorings. 

Trying to do smart guesses about which code needs to be rebuild seems like an overkill to me at this point. So, why not just re-add this little check for existence the generated output folders?

cognitiaclaeves:
I'd like to add to this: Don't run get if using python and package for provider is installed with pip.

PR #698, opened 2021-05-10
feat(lib): Asset construct

### LINKED ISSUE #648: Asset Construct
### Community Note

### Description

When dealing with scenarios where files are either copied or generated in a construct, it'd be great to have a helper Construct to move a given path into the synthesized stack output. 

Conceptually, I'd follow what the AWS CDK has already:

```ts
const fileAsset = new Asset(this, 'SampleSingleFileAsset', {
  path: path.join(__dirname, 'file-asset.txt')
});
```

With this `fileAsset` construct, I'd expect that the file `file-asset.txt` is present in the stack's synthesized output folder. Referencing `fileAsset.outputPath` in other constructs should work.

### References

- https://docs.aws.amazon.com/cdk/latest/guide/assets.html
- Follow up for #636

DanielMSchmidt:
Closes #648

### DIFF
--- a/docs/working-with-cdk-for-terraform/terraform-assets.md
+++ b/docs/working-with-cdk-for-terraform/terraform-assets.md
@@ -41,3 +41,7 @@ const app = new App();
 new MyStack(app, "demo");
 app.synth();
 ```
+
+## Dependencies
+
+Using `AssetType.ARCHIVE` requires `zip` to be installed and in the path.
--- a/docs/working-with-cdk-for-terraform/terraform-assets.md
+++ b/docs/working-with-cdk-for-terraform/terraform-assets.md
@@ -1,0 +1,43 @@
+# Terraform Assets
+
+A Terraform Asset takes a file or directory outside of the CDK for Terraform context and moves it into it. Assets copy referenced files into the stacks context for further usage in other resources.
+
+```typescript
+import * as path from "path";
+import { Construct } from "constructs";
+import { App, TerraformStack, TerraformAsset, AssetType } from "cdktf";
+import { AwsProvider } from "./.gen/providers/aws/aws-provider";
+import { S3BucketObject } from "./.gen/providers/aws/s3-bucket-object";
+import { S3Bucket } from "./.gen/providers/aws/s3-bucket";
+
+
+class MyStack extends TerraformStack {
+  constructor(scope: Construct, name: string) {
+    super(scope, name);
+
+    new AwsProvider(this, "provider", {
+      region: "us-west-2",
+    });
+
+    const bucket = new S3Bucket(this, "bucket", {
+      bucket: "demo",
+    });
+
+    const asset = new TerraformAsset(this, "lambda-asset", {
+      path: path.resolve(__dirname, "../lambda"),
+      type: AssetType.ARCHIVE, // if left empty it infers directory and file
+    });
+
+    new S3BucketObject(this, "lambda-archive", {
+      bucket: bucket.bucket,
+      key: "lambda.zip",
+      source: asset.path,
+    });
+  }
+}
+
+
+const app = new App();
+new MyStack(app, "demo");
+app.synth();
+```
--- a/docs/working-with-cdk-for-terraform/terraform-assets.md
+++ b/docs/working-with-cdk-for-terraform/terraform-assets.md
@@ -35,7 +35,7 @@ class MyStack extends TerraformStack {
 
     new S3BucketObject(this, "lambda-archive", {
       bucket: bucket.bucket,
-      key: "lambda.zip",
+      key: asset.fileName,
       source: asset.path,
     });
   }
--- a/docs/working-with-cdk-for-terraform/terraform-assets.md
+++ b/docs/working-with-cdk-for-terraform/terraform-assets.md
@@ -36,7 +36,7 @@ class MyStack extends TerraformStack {
     new S3BucketObject(this, "lambda-archive", {
       bucket: bucket.bucket,
       key: asset.fileName,
-      source: asset.path,
+      source: asset.path, // returns a posix path
     });
   }
 }
@@ -46,7 +46,3 @@ const app = new App();
 new MyStack(app, "demo");
 app.synth();
 ```
-
-## Dependencies
-
-Using `AssetType.ARCHIVE` requires `zip` to be installed and in the path.
--- a/docs/working-with-cdk-for-terraform/terraform-assets.md
+++ b/docs/working-with-cdk-for-terraform/terraform-assets.md
@@ -2,6 +2,11 @@
 
 A Terraform Asset takes a file or directory outside of the CDK for Terraform context and moves it into it. Assets copy referenced files into the stacks context for further usage in other resources.
 
+This is useful for:
+
+- Lambda functions for copying over previously generated zip files
+- S3 Content for deploying static local files
+
 ```typescript
 import * as path from "path";
 import { Construct } from "constructs";
--- a/packages/cdktf/lib/private/fs.ts
+++ b/packages/cdktf/lib/private/fs.ts
@@ -1,0 +1,24 @@
+import * as fs from "fs";
+import * as path from "path";
+
+// Full implementation at https://github.com/jprichardson/node-fs-extra/blob/master/lib/copy-sync/copy-sync.js
+export function copySync(src: string, dest: string, _opts: {} = {}) {
+  function copyItem(p: string) {
+    const sourcePath = path.resolve(src, p);
+    const stat = fs.statSync(sourcePath);
+    if (stat.isFile()) {
+      fs.copyFileSync(sourcePath, path.resolve(dest, p));
+    }
+    if (stat.isDirectory()) {
+      walkSubfolder(p);
+    }
+  }
+
+  function walkSubfolder(p: string) {
+    const sourceDir = path.resolve(src, p);
+    fs.mkdirSync(path.resolve(dest, p), { recursive: true });
+    fs.readdirSync(sourceDir).forEach((item) => copyItem(path.join(p, item)));
+  }
+
+  walkSubfolder(".");
+}
--- a/packages/cdktf/lib/private/fs.ts
+++ b/packages/cdktf/lib/private/fs.ts
@@ -1,8 +1,10 @@
+import { execSync } from "child_process";
 import * as fs from "fs";
 import * as path from "path";
+import * as os from "os";
 
 // Full implementation at https://github.com/jprichardson/node-fs-extra/blob/master/lib/copy-sync/copy-sync.js
-export function copySync(src: string, dest: string, _opts: {} = {}) {
+export function copySync(src: string, dest: string) {
   function copyItem(p: string) {
     const sourcePath = path.resolve(src, p);
     const stat = fs.statSync(sourcePath);
@@ -21,4 +23,19 @@ export function copySync(src: string, dest: string, _opts: {} = {}) {
   }
 
   walkSubfolder(".");
+}
+
+export function archiveSync(src: string, dest: string) {
+  try {
+    if (os.platform() === "win32") {
+      execSync("where zip");
+    } else {
+      execSync("which zip");
+    }
+  } catch {
+    console.error(`Unable to find "zip".`);
+    process.exit(1);
+  }
+
+  execSync(`zip -r ${dest} ${src}`);
 }
--- a/packages/cdktf/lib/private/fs.ts
+++ b/packages/cdktf/lib/private/fs.ts
@@ -29,18 +29,25 @@ export function copySync(src: string, dest: string) {
 }
 
 export function archiveSync(src: string, dest: string) {
-  try {
-    if (os.platform() === "win32") {
-      execSync("where zip");
-    } else {
-      execSync("which zip");
+  if (os.platform() === "win32") {
+    try {
+      execSync("where tar.exe");  
+    } catch {
+      console.error(`Unable to find "tar.exe".`);
+      process.exit(1);
     }
-  } catch {
-    console.error(`Unable to find "zip".`);
-    process.exit(1);
-  }
+    
+    execSync(`tar.exe -a -c -f ${dest} ${src}`);
+  } else {
+    try {
+      execSync("which zip");
+    } catch {
+      console.error(`Unable to find "zip".`);
+      process.exit(1);
+    }
 
-  execSync(`zip -r ${dest} ${src}`);
+    execSync(`zip -r ${dest} ${src}`);
+  }
 }
 
 export function hashPath(src: string) {
--- a/packages/cdktf/lib/private/fs.ts
+++ b/packages/cdktf/lib/private/fs.ts
@@ -2,6 +2,9 @@ import { execSync } from "child_process";
 import * as fs from "fs";
 import * as path from "path";
 import * as os from "os";
+import * as crypto from "crypto";
+
+const HASH_LEN = 32;
 
 // Full implementation at https://github.com/jprichardson/node-fs-extra/blob/master/lib/copy-sync/copy-sync.js
 export function copySync(src: string, dest: string) {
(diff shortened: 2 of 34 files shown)


docs/working-with-cdk-for-terraform/terraform-assets.md
```diff
@@ -0,0 +1,47 @@
...
+    });
+
+    const asset = new TerraformAsset(this, "lambda-asset", {
+      path: path.resolve(__dirname, "../lambda.zip") // can also point to a folder
+    });
+
+    new S3BucketObject(this, "lambda-archive", {
+      bucket: bucket.bucket,
+      key: "lambda.zip",
+      source: asset.path,
```
skorfmann, line 38: Would be neat if there was something like `asset.fileName` if the path represents a file

test/python/asset/cdktf.json
```diff
@@ -0,0 +1,10 @@
+{
+  "language": "python",
+  "app": "pipenv run python main.py",
+  "terraformProviders": ["aws@~> 2.0"],
+  "terraformModules": ["terraform-aws-modules/vpc/aws@2.77.0"],
```
skorfmann, line 5: Let's speed the test up and drop the aws provider / module in favour of the null provider.

skorfmann (commented):
Left a few comments, but looks very good already 👍

packages/cdktf/lib/terraform-asset.ts
```diff
@@ -0,0 +1,93 @@
...
+}
+
+export enum AssetType {
+  FILE,
+  DIRECTORY,
+}
+
+const ASSETS_DIRECTORY = "assets";
+
+export class TerraformAsset extends TerraformElement {
```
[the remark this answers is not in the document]
DanielMSchmidt: Lack of knowledge, will switch out for `Resource` 👍

docs/working-with-cdk-for-terraform/terraform-assets.md
DanielMSchmidt: Just to be sure, you mean as an additional property not instead of `path`, right?

test/python/asset/cdktf.json
DanielMSchmidt: Good point 🙌

docs/working-with-cdk-for-terraform/terraform-assets.md
skorfmann: yes, in addition to path. Essentially to use if as `key` in this example.

packages/cdktf/lib/terraform-asset.ts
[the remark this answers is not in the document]
skorfmann: Ok, cool - I think you'll need to extend `Resource` then to get `friendlyUniqueId` :)

test/python/asset/cdktf.json
jsteinich: Why use a provider at all? Or why not use a prebuilt one?

packages/cdktf/lib/terraform-stack.ts
```diff
@@ -11,6 +11,26 @@ import { Manifest } from './manifest'
...
 
+
+/**
+ * Returns the naming scheme used to allocate logical IDs. By default, uses
+ * the `HashedAddressingScheme` but this method can be overridden to customize
+ * this behavior.
+ *
+ * @param tfElement The element for which the logical ID is allocated.
+ */
+export function allocateLogicalId(node: Node, stack: TerraformStack): string {
```
jsteinich, line 22: By changing this to a function, I'm no longer sure that users can actually override the strategy being used.

test/python/asset/cdktf.json
DanielMSchmidt: Right :D there is no real need, will remove it

packages/cdktf/lib/terraform-stack.ts
DanielMSchmidt: How would a user override this and why? (Sorry I lack context here 🙈 )

Would this solve the problem?

```ts
export class TerraformAsset extends Resource {

  protected allocateLogicalId = allocateLogicalId;
```

test/python/asset/cdktf.json
DanielMSchmidt: Ok I won't, the tests call `cdktf get` which returns exit code 1 if no modules / providers are specified. I wouldn't want to 'fix' this inside the test harness since this is also annoying for CLI users. #690 captures that

packages/cdktf/lib/terraform-stack.ts
DanielMSchmidt: I understand your comment now, will try to use stack.getLogicalID in the asset :)

packages/cdktf/lib/terraform-asset.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    if (!path.isAbsolute(config.path)) {
+      throw new Error(
+        `TerraformAsset path needs to be absolute, got relative path: '${config.path}'`
+      );
+    }
+
+    const stat = fs.statSync(config.path);
+    const inferredType = stat.isFile() ? AssetType.FILE : AssetType.DIRECTORY;
+    this.type = !config.type ? inferredType : config.type;
```
jsteinich, line 37: Minor: any reason not to use `this.type = config.type ?? inferredType`?

packages/cdktf/lib/terraform-stack.ts
```diff
@@ -85,12 +85,13 @@ export class TerraformStack extends Construct {
    *
    * @param tfElement The element for which the logical ID is allocated.
    */
-  protected allocateLogicalId(tfElement: TerraformElement): string {
-    const node = tfElement.constructNode
+  protected allocateLogicalId(tfElement: TerraformElement | Node): string {
+    const node =  "constructNode" in tfElement ? tfElement.constructNode : tfElement;
+    const stack = "stack" in tfElement ? tfElement.stack : this;
```
jsteinich, line 90: Probably more standard Typescript to have something along the lines of:
```typescript
if (tfElement instanceof TerraformElement) {
...
}
else {
...
}
```

packages/cdktf/lib/private/fs.ts
```diff
@@ -0,0 +1,48 @@
+import * as fs from "fs";
+import * as path from "path";
+import * as crypto from "crypto";
+
+const HASH_LEN = 32;
+
+// Full implementation at https://github.com/jprichardson/node-fs-extra/blob/master/lib/copy-sync/copy-sync.js
```
jsteinich, line 7: Any reason to not just use `fs-extra`?

test/typescript/terraform-cloud/main.ts
```diff
@@ -26,6 +27,10 @@ export class HelloTerra extends TerraformStack {
         token
       }
     });
+
+    new TerraformAsset(this, "asset-a", {
```
jsteinich, line 31: I think it's good to have this in the integration test, but it would be nice if the test actually verified the result somehow.

test/python/asset/main.py
```diff
@@ -0,0 +1,28 @@
...
+from cdktf import App, TerraformStack, Testing, TerraformAsset
+from imports.null import NullProvider, Resource
+
+class MyStack(TerraformStack):
+    def __init__(self, scope: Construct, ns: str):
+        super().__init__(scope, ns)
+
+        NullProvider(self, "null")
+        Resource(self, "null-resource")
+        self.add_override('terraform.backend', {
```
jsteinich, line 13: If specifying a remote backend is useful here, should probably use the code defined version rather than the override. If isn't really part of the test, might be best to just remove.

packages/cdktf/lib/terraform-asset.ts
```diff
@@ -0,0 +1,102 @@
...
+import * as fs from "fs";
+import * as path from "path";
+import { Manifest } from "./manifest";
+import { copySync, hashPath } from "./private/fs";
+import { Resource } from "./resource";
+
+export interface TerraformAssetConfig {
+  readonly path: string;
+  readonly type?: AssetType;
+  readonly assetHash?: string;
```
jsteinich, line 11: What scenario do you see where a user would pass this configuration in?

packages/cdktf/lib/terraform-asset.ts
```diff
@@ -0,0 +1,102 @@
...
+export enum AssetType {
+  FILE,
+  DIRECTORY,
+}
+
+const ASSETS_DIRECTORY = "assets";
+
+export class TerraformAsset extends Resource {
+  private sourcePath: string;
+  public assetHash: string;
```
jsteinich, line 23: I can see how having this could be useful, but there doesn't appear to be any use of this currently.

jsteinich (changes requested):
Looks to be on the right track, just a few things to clean up.

I think we'll also want to do a follow-up PR to make local modules automatically use assets.

packages/cdktf/lib/private/fs.ts
DanielMSchmidt: I understand we would need to bundle it because of JSII and my understanding was the general course was to rather write the one function we need ourselves than adding a bundled dependency

packages/cdktf/lib/terraform-asset.ts
DanielMSchmidt: No, I'll change it

packages/cdktf/lib/terraform-stack.ts
DanielMSchmidt: Nice, I thought that would be too simple :D

test/typescript/terraform-cloud/main.ts
DanielMSchmidt: Do you mean verifying what got uploaded includes the asset?

test/python/asset/main.py
DanielMSchmidt: will remove it 👍

packages/cdktf/lib/terraform-asset.ts
DanielMSchmidt: it's intended to be used by resources that rely on assets, e.g. a lambda resource could choose not to recreate because the hash is the same

packages/cdktf/lib/terraform-asset.ts
DanielMSchmidt: A user could choose a hash based on their needs, e.g. ignoring node modules to have more fine grained control around recreation of resources

test/typescript/terraform-cloud/main.ts
jsteinich: Yes

jsteinich (changes requested):
Just a couple of small comments.
Also, make sure to fix the tests and add a link to the new documentation page to the readme.

packages/cdktf/lib/terraform-asset.ts
```diff
@@ -0,0 +1,114 @@
...
+        `TerraformAsset ${id} expects path to be a file, a directory was passed: '${config.path}'`
+      );
+    }
+  }
+
+  public get path(): string {
+    let filename = "";
+    switch (this.type) {
+      case AssetType.ARCHIVE:
+        filename = ARCHIVE_NAME;
```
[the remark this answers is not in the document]
DanielMSchmidt: every asset path has the logical id in it for uniqueness so there should be no collisions.

packages/cdktf/lib/terraform-asset.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    return path.join(
+      ASSETS_DIRECTORY,
+      this.stack.getLogicalId(Node.of(this)), // needed to get a human friendly, unique segment in the path
+      filename
+    );
+  }
+
+  public get fileName(): string | undefined {
+    return this.type === AssetType.FILE ? path.basename(this.path) : undefined;
```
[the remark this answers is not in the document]
DanielMSchmidt: I totally missed that while removing the revert commit

docs/working-with-cdk-for-terraform/terraform-assets.md
```diff
@@ -0,0 +1,52 @@
...
+
+
+const app = new App();
+new MyStack(app, "demo");
+app.synth();
+```
+
+## Dependencies
+
+Using `AssetType.ARCHIVE` requires `zip` to be installed and in the path.
```
skorfmann, line 52: or `tar` in case of windows.

packages/cdktf/lib/private/fs.ts
```diff
@@ -0,0 +1,72 @@
...
+  function walkSubfolder(p: string) {
+    const sourceDir = path.resolve(src, p);
+    fs.mkdirSync(path.resolve(dest, p), { recursive: true });
+    fs.readdirSync(sourceDir).forEach((item) => copyItem(path.join(p, item)));
+  }
+
+  walkSubfolder(".");
+}
+
+export function archiveSync(src: string, dest: string) {
```
skorfmann, line 31: Alternatively, we could try to shell out to a nodejs process, so it behaves consistently across all platforms. We would have to bundle something like [archiver](https://www.npmjs.com/package/archiver) as a [jsii dependency](https://aws.github.io/jsii/user-guides/lib-author/typescript-restrictions/) to make this work.

packages/cdktf/lib/terraform-asset.ts
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  public get fileName(): string | undefined {
+    switch (this.type) {
+      case AssetType.ARCHIVE:
+        return ARCHIVE_NAME;
+      case AssetType.FILE:
+        return path.basename(this.sourcePath);
+      case AssetType.DIRECTORY:
+        return undefined;
```
skorfmann, line 71: Perhaps fail loud and explicitly in this case?

packages/cdktf/lib/terraform-asset.ts
```diff
@@ -0,0 +1,112 @@
...
+
+      case AssetType.ARCHIVE:
+        archiveSync(this.sourcePath, targetPath);
+        break;
+      default:
+        throw new Error(`Asset type ${this.type} is not implemented`);
+    }
+  }
+
+  public toTerraform(): any {
```
skorfmann, line 109: I don't think this needs to be implemented here.

test/typescript/terraform-cloud/main.ts
skorfmann: Perhaps you could use one of the Terraform functions to check for the [presence](https://www.terraform.io/docs/language/functions/fileexists.html)?

packages/cdktf/lib/terraform-asset.ts
DanielMSchmidt: I wasn't considering this a failure, undefined signals for me that a value does not exist. Do you think people might run into problems here?

packages/cdktf/lib/private/fs.ts
DanielMSchmidt: I was trying to get around this as it seems less clean 🤔 do we do something similar already somewhere in the project?

skorfmann: No, so far we're not doing this. Why do you think it seems less clean?

DanielMSchmidt: It seems like more moving parts to me, archiver in a bundle, a local node file to execute versus two natively installed binaries seems more execution boundaries. But I don't have a super strong opinion here, we can totally move to archiver if you think it makes more sense

packages/cdktf/lib/terraform-asset.ts
skorfmann: If a user tries to get a filename for the directory, what would you think is the intention? I'd assume the user is trying to get some recognizable name. Now, I think you're right - this shouldn't result in an error. Perhaps we're more looking for something like basename?

```ts
> path.basename('/tmp/foo/bar')
'bar'
> path.basename('/tmp/foo/bar.zip')
'bar.zip'
```

DanielMSchmidt: I really did not have a use-case in mind, but yours makes sense, basename fits fine here 👍

packages/cdktf/lib/private/fs.ts
skorfmann: > It seems like more moving parts to me, archiver in a bundle, a local node file to execute versus two natively installed binaries seems more execution boundaries.

True, the upside would be that there's no burden on the user to have these binaries installed. e.g. Is `tar` installed by default on Windows? 

Haven't tried shelling out to node in the context of jsii yet. So I'm not sure if that's good or bad.

DanielMSchmidt: I understand tar is installed by default, but I also agree it could put some burden on the user. I'll move it over to shellout in a single revertable commit, if it causes problems we can fallback to this implementation

skorfmann (approved):
Awesome stuff! 🎉  Can't wait to use it :)

PR #712, opened 2021-05-18
Update UI state on deploy / destroy

DanielMSchmidt:
## Summary

- Throws prefixed errors in case of terraform cloud API errors so that there is more context
- Throws a new error when the plan errored
- Poll logs from terraform cloud to see updates inbetween
- Throws a new error when the workspace is locked


skorfmann:
I'm probably lacking a bit of context here. I think a quick summary of the intended change would help me a lot :)

DanielMSchmidt:
@skorfmann I completely forgot to add a meaningful description, I added it now 👍

skorfmann:
> @skorfmann I completely forgot to add a meaningful description, I added it now 👍

Thanks 👍  - I merged the upstream dependency, should be released soon

skorfmann (approved):

PR #714, opened 2021-05-20
Add dependabot for npm with auto merge

### LINKED ISSUE #362: Add Dependabot Configuration
### Community Note

### Description

We could configure Dependabot to update:

- Github Actions
- NPM Packages

I'm not sure how useful the github actions part is, that's something to investigate. For the NPM package ecosystem, I think it would be quite handy to automatically upgrade non-critical packages.

Critical packages I think we'd want to handle manually (taken from the [AWS CDK config](https://github.com/aws/aws-cdk/blob/443948164e09aaa81c094c013b32aa1f67b69570/.dependabot/config.yml)):

```yaml
      - match:
          dependency_name: "jsii*"
      - match:
          dependency_name: "@jsii/*"
      - match:
          dependency_name: "codemaker"
```

### References

There was an initial Pull Request with a few references https://github.com/hashicorp/terraform-cdk/pull/316

- https://docs.github.com/en/github/administering-a-repository/customizing-dependency-updates
- https://github.com/aws/aws-cdk/blob/443948164e09aaa81c094c013b32aa1f67b69570/.dependabot/config.yml

jsteinich:
Closes #362 

Checking for GitHub Action updates doesn't seem like it would add much value at this point, but I don't have a ton of evidence to back that up.


.mergify.yml
```diff
@@ -0,0 +1,11 @@
+pull_request_rules:
```
skorfmann, line 1: I don't think we have mergify in the context of the HashiCorp Github Organization. Do you have an update on this @danieldreier ?

skorfmann: Ok, since I suspect this might take a while, let's just drop the mergify config for now. We'll merge these dependabot PRs by hand.

PR #721, opened 2021-05-25
Add pr lint for semantic releases

skorfmann:


.github/workflows/pr-lint.yml
```diff
@@ -0,0 +1,42 @@
...
+            readme
+            docs
+            release
+            deps
+          # Configure that a scope must always be provided.
+          requireScope: true
+          # When using "Squash and merge" on a PR with only one commit, GitHub
+          # will suggest using that commit message instead of the PR title for the
+          # merge commit, and it's easy to commit this by mistake. Enable this option
+          # to also validate the commit message for one commit PRs.
```
DanielMSchmidt, line 41: From past experience I would suggest we add a `subjectPatternError` to make it clear why this linter exists. I'd suggest something like

> The subject "{subject}" found in the pull request title "{title}" does not match the required format. The format is "type(scope): message", supported types are feat, fix, chore, refactor, revert. The supported scopes are lib, cli, hcl2json, tests, examples, readme, docs, release, deps. Please stick to the format so that our automatic release process picks your commit up and your code makes it into the next release.

.github/workflows/pr-lint.yml
```diff
@@ -0,0 +1,42 @@
...
+      - uses: amannn/action-semantic-pull-request@v3.4.0
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          types: |
+            feat
+            fix
+            chore
+            refactor
+            revert
```
DanielMSchmidt, line 25: should we add `test` as well?

.github/workflows/pr-lint.yml
```diff
@@ -0,0 +1,42 @@
...
+          scopes: |
```
DanielMSchmidt, line 26: More of a conceptual thing, but the scopes currently correlate with directories. If we would take topics that are more in line with features we have (e.g. asset / cloud) the commits could give more context hints when debugging.

.github/workflows/pr-lint.yml
skorfmann: Yes, I like that 👍

.github/workflows/pr-lint.yml
skorfmann: That's probably difficult to scope correctly :D and it is more likely to change.  However, which kind of scopes would you list here then?

DanielMSchmidt: yes and yes :D I'd use

- cloud
- api
- each language
- tooling
- cli

Now that I listed them they don't seem so different anymore :D

skorfmann: Ok, I'll keep it as it is then. We can always come back to this and adjust it if we're not happy with it.

.github/workflows/pr-lint.yml
skorfmann: Yep, that makes sense

DanielMSchmidt (approved):

PR #737, opened 2021-05-27
fix(tests): Run integration tests based on directories

skorfmann:
We had some issues with the `group` runner approach, since it requires manual tagging of the tests which turned out to be errror prone.

One test wasn't tagged at all. And another test wasn't picked up, since the commenting format seems to be significant.

Doesn't work:

```ts
//
// Testing a full cycle of diff, deploy and destroy
//
// @group typescript
//
```

Works:

```ts
/**
 * Testing synthing typescript to json
 *
 * @group typescript
 */
```

What are your thoughts on the group runner in general. Should we keep it for local testing purposes or get rid of it entirely?


skorfmann:
This depends on #736 since it's now running all tests in the respective directories

ansgarm:
Rebased this onto my changes made in https://github.com/hashicorp/terraform-cdk/pull/735 to give the Github Actions the chance to run already while we merge the other PR first 🐌

test/typescript/multiple-stacks/test.ts
```diff
@@ -5,6 +5,9 @@
 //
 import { TestDriver } from "../../test-helper";
 
+const onWindows = process.platform === 'win32' ? it.skip : it.skip
```
jsteinich, line 8: Seems incorrect for both branches to be the same here.

ansgarm: Oh, you're right. Didn't catch that when stealing it from https://github.com/hashicorp/terraform-cdk/blob/5bb0d6877ff334869ec555af53cf885616ba1a7f/test/typescript/modules/test.ts#L12-L12

ansgarm: Added a separate PR to fix the test I stole it from: https://github.com/hashicorp/terraform-cdk/pull/740

ansgarm (commented):
Accidentally commit the fix for the other windows snapshot on this branch – isn't used here anyway 😅

PR #741, opened 2021-05-27
fix: fix github workflow for releasing cdktf go package

ansgarm:
Related Workflow failure: https://github.com/hashicorp/terraform-cdk/runs/2685171282?check_suite_focus=true


ansgarm:
@skorfmann I think we could safely disregard waiting for the Checks as only Github workflows were modified that do not run on a pull request. (Would save us the waiting time 🐌)

skorfmann (approved):

PR #742, opened 2021-05-27
chore(docs): add google cloud kubernetes example

DanielMSchmidt:


skorfmann (changes requested):
Looks great! In order to run these as part of the CI pipeline, it'd be good to add a [package.json](https://github.com/hashicorp/terraform-cdk/blob/633c9136ed241db6e3aa4e8d67e178766a63032e/examples/go/docker/package.json) (this will include this example via [lerna](https://github.com/hashicorp/terraform-cdk/blob/633c9136ed241db6e3aa4e8d67e178766a63032e/package.json#L19))

skorfmann (approved):
Cool - thanks for updating! Looks good 👍

PR #768, opened 2021-06-09
chore(docs): add google cloud python cloud function example

prashanthbasani:


examples/python/google/cloud_function_stack.py
```diff
@@ -0,0 +1,51 @@
...
+        super().__init__(scope, ns)
+
+        GoogleProvider(self, id=project_id, region="us-central1", project=project_id)
+
+        code_bucket = StorageBucket(self, "bucket", name="terraform-cf-zip-files", project=project_id)
+
+        code_object = StorageBucketObject(self, "archive", \
+            name="source.zip",
+            bucket = code_bucket.name,
+            source = os.path.join(os.path.abspath("./"), "function", "source.zip"),
```
DanielMSchmidt, line 26: Could you use [TerraformAsset](https://github.com/hashicorp/terraform-cdk/blob/main/docs/working-with-cdk-for-terraform/terraform-assets.md) here? It can even zip the folder for you and it will work properly in all contexts (e.g. with TerraformCloud)

prashanthbasani: I didn't know of that. Thanks! updated the PR :)

PR #770, opened 2021-06-09
fix(cli): adjust to JSON based terraform cloud output

DanielMSchmidt:
## TODOs

We can postpone each of these todos to unblock the other PRs :)

- [ ] Check for real type if possible
- [ ] Don't assume the structure but validate it

### DIFF
--- a/packages/cdktf-cli/bin/cmds/ui/models/terraform-cloud.ts
+++ b/packages/cdktf-cli/bin/cmds/ui/models/terraform-cloud.ts
@@ -2,7 +2,7 @@
 import * as path from 'path';
 import * as os from 'os';
 import * as fs from 'fs';
-import { Terraform, TerraformPlan, TerraformOutput, AbstractTerraformPlan } from './terraform';
+import { Terraform, TerraformPlan, TerraformOutput, AbstractTerraformPlan, DeployingResource, getResourceState, PlannedResourceAction } from './terraform';
 import { TerraformJsonConfigBackendRemote } from '../terraform-json'
 import * as TerraformCloudClient from '@skorfmann/terraform-cloud'
 import archiver from 'archiver';
@@ -25,6 +25,29 @@ export interface TerraformCredentials {
 export interface TerraformCredentialsFile {
   credentials: TerraformCredentials;
 }
+
+// This is ducktyped, we should check the real type upstream
+type CloudLog = {
+  "@level": string;
+  "@message": string;
+  "@module": string;
+  "@timestamp": string;
+  hook?: {
+    resource?: {
+      addr?: string;
+      module?: string;
+      resource?: string;
+      implied_provider?: string;
+      resource_type?: string;
+      resource_name?: string;
+    };
+    action?: string;
+    id_key?: string;
+    id_value?: string;
+  };
+  type: string;
+};
+
 
 const zipDirectory = (source: string): Promise<Buffer | false> => {
   const archive = archiver('tar', { gzip: true });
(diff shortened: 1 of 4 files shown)


packages/cdktf-cli/bin/cmds/ui/models/terraform-cloud.ts
```diff
@@ -320,4 +343,26 @@ export class TerraformCloud implements Terraform {
...
+    try {
+      message = JSON.parse(line) as CloudLog
+    } catch (error) {
+      throw new Error(`Could not parse Terraform Cloud output: ${line}`);
+    }
+
+    const resourceMatch = message.hook?.resource?.resource;
+    const applyState = getResourceState(message['@message']); // TODO: use type instead
+
+    if (applyState && resourceMatch && resourceMatch.length >= 0 && resourceMatch[1] != "Warning") {
```
jsteinich, line 358: I imagine the `Warning` check isn't needed in this case.

packages/cdktf-cli/bin/cmds/ui/models/terraform.ts
```diff
@@ -93,10 +93,35 @@ export abstract class AbstractTerraformPlan implements TerraformPlan {
...
+  }
+}
+
 export interface Terraform {
   init: () => Promise<void>;
   plan: (destroy: boolean) => Promise<TerraformPlan>;
   deploy(planFile: string, stdout: (chunk: Buffer) => any): Promise<void>;
   destroy(stdout: (chunk: Buffer) => any): Promise<void>;
   output(): Promise<{[key: string]: TerraformOutput}>;
+  outputParser: (line: string) => DeployingResource | undefined;
```
jsteinich, line 126: I'm not sure this is the best level for this abstraction. I think we'll want the same (or almost the same) parser added to the non-cloud version based on a Terraform version check so that we can use https://www.terraform.io/docs/internals/machine-readable-ui.html

packages/cdktf-cli/bin/cmds/ui/models/terraform-cloud.ts
```diff
@@ -320,4 +343,26 @@ export class TerraformCloud implements Terraform {
     await Promise.race([ready(), timeout()]);
     logger.debug('Configuration Version is ready in Terraform Cloud');
   }
+
+  public outputParser(line: string): DeployingResource | undefined {
+    let message: CloudLog;
+    try {
+      message = JSON.parse(line) as CloudLog
+    } catch (error) {
+      throw new Error(`Could not parse Terraform Cloud output: ${line}`);
```
jsteinich, line 352: We might want to just print or log the line rather than throw.

DanielMSchmidt: Good point, I will change it 👍

packages/cdktf-cli/bin/cmds/ui/models/terraform-cloud.ts
DanielMSchmidt: Yeah, we get this safely from the hook now, should not be needed anymore

packages/cdktf-cli/bin/cmds/ui/models/terraform.ts
DanielMSchmidt: I am also not 100% satisfied with this place. Where would you put it? Or should we just always assume the output is json?

jsteinich: Left a note in the alternative PR, but probably depends on if the cloud output json is exactly the same format as the cli output json. If so, then probably either always try json and fallback on failure, or just be explicit about which parsing mode (so still at this level, but not the entire parser). If they are different, then this may be the best spot.

PR #774, opened 2021-06-10
fix(cli): parse JSON output

DanielMSchmidt:
Alternative approach for #770, I think I like it better. I strongly typed every possible message mentioned in https://www.terraform.io/docs/internals/machine-readable-ui.html (thank you @jsteinich for pointing me to it).

### DIFF
--- a/packages/cdktf-cli/package.json
+++ b/packages/cdktf-cli/package.json
@@ -52,7 +52,8 @@
     "stream-buffers": "^3.0.2",
     "strip-ansi": "^6.0.0",
     "uuid": "^8.3.0",
-    "yargs": "^15.1.0"
+    "yargs": "^15.1.0",
+    "zod": "^1.11.7"
   },
   "eslintConfig": {
     "parser": "@typescript-eslint/parser",
--- a/packages/cdktf-cli/bin/cmds/ui/models/schema.ts
+++ b/packages/cdktf-cli/bin/cmds/ui/models/schema.ts
@@ -1,0 +1,169 @@
+import * as z from "zod";
+
+/* eslint-disable @typescript-eslint/camelcase */
+// Schemas from https://www.terraform.io/docs/internals/machine-readable-ui.html#change_summary
+const resource = z.object({
+  addr: z.string(),
+  module: z.string(),
+  resource: z.string(),
+  implied_provider: z.string(),
+  resource_type: z.string(),
+  resource_name: z.string(),
+  resource_key: z.string().optional().nullable()
+});
+const baseMessage = z
+  .object({
+    "@level": z.enum(["info", "error", "warn"]),
+    "@message": z.string(),
+    "@module": z.string(),
+    "@timestamp": z.string()
+  })
+  .nonstrict();
+
+const action = z.enum(["noop", "create", "read", "update", "delete"]);
+const change = z.object({
+  resource,
+  action
+});
+const replaceChange = change.extend({
+  action: z.literal("replace"),
+  reason: z.enum(["tainted", "requested", "cannot_update"])
+});
+const plannedChange = baseMessage.extend({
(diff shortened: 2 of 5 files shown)


jsteinich:
We'll need to supply the `-json` argument to the Terraform cli commands. Hopefully this is just ignored in versions < 0.15.3, but if not, will need to omit.

We also probably aren't yet at a point where we can drop support for < 0.15.3, so either need to be explicit about when to parse JSON or try always and fall back to old behavior on failure.

packages/cdktf-cli/package.json
```diff
@@ -52,7 +52,8 @@
     "stream-buffers": "^3.0.2",
     "strip-ansi": "^6.0.0",
     "uuid": "^8.3.0",
-    "yargs": "^15.1.0"
+    "yargs": "^15.1.0",
+    "zod": "^1.11.7"
```
ansgarm, line 56: That lib looks nice! Is there a reason to not use v2 or v3?

ansgarm: Ah found out why myself: Because v3 requires TypeScript 4.1+ and v2 is discontinued 😄

skorfmann: Perhaps we should upgrade Typescript - for the CLI that shouldn't be a problem.

skorfmann:
> We'll need to supply the `-json` argument to the Terraform cli commands. Hopefully this is just ignored in versions < 0.15.3, but if not, will need to omit.

Looks like the cli fails prior to `0.15.3`.

> We also probably aren't yet at a point where we can drop support for < 0.15.3, so either need to be explicit about when to parse JSON or try always and fall back to old behavior on failure.

I think we should be rather aggressive here. Probably with `0.5` we should make a cut and support Terraform `1.0` only. Since that's affecting the CLI only, users still have the option to fallback to an older Terraform CLI by using it directly.

skorfmann:
> Alternative approach for #770, I think I like it better. I strongly typed every possible message mentioned in

That's pretty cool 👍 

The tests are failing due to the log statements. Is there anything else which is to do here?

jsteinich:
> I think we should be rather aggressive here. Probably with `0.5` we should make a cut and support Terraform `1.0` only. Since that's affecting the CLI only, users still have the option to fallback to an older Terraform CLI by using it directly.

I'm a bit torn here. I understand not wanting to have baggage, but Terraform has been moving through versions quickly (0.15 came out 2 months ago, 0.14 was 6 months ago). I personally have stacks created with cdktf that are on 0.13.x.

Module are provider versions are always compatible across Terraform versions, so this can bleed outside of just the cli.

DanielMSchmidt:
> The tests are failing due to the log statements. Is there anything else which is to do here?

Mainly debug the failing tests, if it's just the logs I'll remove them now 👍

skorfmann:
> > I think we should be rather aggressive here. Probably with `0.5` we should make a cut and support Terraform `1.0` only. Since that's affecting the CLI only, users still have the option to fallback to an older Terraform CLI by using it directly.
> 
> I'm a bit torn here. I understand not wanting to have baggage, but Terraform has been moving through versions quickly (0.15 came out 2 months ago, 0.14 was 6 months ago). I personally have stacks created with cdktf that are on 0.13.x.

Is there any blocker preventing you from upgrading?

jsteinich:
> Is there any blocker preventing you from upgrading?

Time

skorfmann:
> > Is there any blocker preventing you from upgrading?
> 
> Time

Hehe, ok :) 

Right now it works for both output versions (text / json). At latest with Terraform 1.1, we're going to drop support for 0.15 anyway. However, I'd still push for an earlier move in that direction. And again, this only affects the `cdktf-cli` anyway, `deploy`, `diff` and `destroy` particularly

PR #776, opened 2021-06-13
chore(docs): python example for gcp cloud function

prashanthbasani:
- Used Terraform Asset for uploading the cloud function source code to bucket.
- Added package.json


skorfmann (approved):
Cool, thanks for the example! Made a small adjustment to use the locally built cdktf package - see cdd84c3 - Other than that it looks good 👍

PR #778, opened 2021-06-14
chore(docs): Enhance readme with Terraform Platform Illustration 

skorfmann:
The desired outcome of this Readme change is, that it's more clear to the user, that all providers can be used.

The illustration was taken from the terraform.io homepage. 

Here's the [rendered](https://github.com/hashicorp/terraform-cdk/blob/53479fe300f1c2277c2a4140f255c8a509b1c90c/README.md) version.


ansgarm (approved):
👍
Feels a bit big on a wide screen, but maybe that's just my personal preference?

DanielMSchmidt (approved):

PR #779, opened 2021-06-14
chore(deps): Upgrade yargs to v17

skorfmann:
Looks like we're running a pretty old version of yargs. One of the transitive dependencies of the old version had a security vulnerability which should be fixed by upgrading to the latest version. [It](https://snyk.io/vuln/npm:y18n) doesn't look really critical, but it's probably still a good idea to do the upgrade nevertheless.

The changes in yargs:

- [v16.0](https://github.com/yargs/yargs/pull/1698)
- [v17.0](https://github.com/yargs/yargs/pull/1876)

### DIFF
--- a/packages/cdktf-cli/package.json
+++ b/packages/cdktf-cli/package.json
@@ -52,7 +52,7 @@
     "stream-buffers": "^3.0.2",
     "strip-ansi": "^6.0.0",
     "uuid": "^8.3.0",
-    "yargs": "^15.1.0"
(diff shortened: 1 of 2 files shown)


ansgarm:
This currently fails because yargs dropped support for NodeJS v10. As v10 had its end of life a few months ago I think we could just drop support for v10 as well.

ansgarm:
rebased this onto `main` as we now are on Node 14 and this should resolve the Node 10 issue

PR #781, opened 2021-06-14
chore(tests): Provide global PR workflow status

skorfmann:
It's difficult to determine all jobs for a PR see https://docs.mergify.io/conditions/#about-status-checks 

Since we have this 'checkrun-timechart' job, which essentially knows all the statuses (see https://github.com/WyriHaximus/github-action-wait-for-status#output) we can leverage it to provide a global status to mergify.

### DIFF
--- a/.github/workflows/timechart.yml
+++ b/.github/workflows/timechart.yml
@@ -8,6 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: wait
+        id: waitforstatuschecks
         uses: "WyriHaximus/github-action-wait-for-status@v1.2"
         with:
           ignoreActions: checkrun-timechart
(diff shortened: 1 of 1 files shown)


.github/workflows/timechart.yml
```diff
@@ -20,4 +21,16 @@ jobs:
...
+      # see https://docs.mergify.io/conditions/#about-status-checks
+      # Since we have this 'checkrun-timechart' job, which essentially
+      # knows all the statuses (see https://github.com/WyriHaximus/github-action-wait-for-status#output),
+      # we can leverage it to provide a global status to mergify.
+      - name: 'mergify status'
+        uses: actions/github-script@v3
+        if: steps.waitforstatuschecks.outputs.status != 'success'
+        with:
+          script: |
+            core.setFailed('One of the tests failed')
```
ansgarm, line 36: Could this also be something other than a test? I.e. compiling and synthezising an example?

ansgarm (approved):

.github/workflows/timechart.yml
skorfmann: better? 8e68df7

ansgarm: yes! Besides the typos 😅 (fixed them already)

Issue #558, opened 2021-02-11
Testing Modules and Providers


skorfmann:
> Also, would you agree with the assumption that it's good enough to test the provider generation process for all providers to Typescript only? Or would you see a huge benefit in testing all providers in all languages?

Given that we have encountered bugs that are language specific (equals in Java AWS, comments in Python Google, etc), I think we need to test at least some of the providers in all languages. Maybe that is just the big 3 cloud providers (examples currently cover that), but maybe it should be the full official provider list.

> Same question for Linux / Windows.

I think we need to generate at least one provider in all languages on Windows. Probably doesn't need to be AWS (I imagine a smaller provider would be much quicker).

I'm not sure what would make sense for module testing. Maybe ones with 500k+ downloads (currently 19).

_Originally posted by @jsteinich in https://github.com/hashicorp/terraform-cdk/issues/553#issuecomment-777483808_

skorfmann:
When approaching this issue, we should make sure to cleanup these tests as well https://github.com/hashicorp/terraform-cdk/pull/713#discussion_r636019727

PR #784, opened 2021-06-15
fix(docs): add mandatory cdktf get step

### LINKED ISSUE #612: from imports.aws import Instance, AwsProvider
I always get:

File "main.py", line 4, in <module>
    from imports.aws import Instance, AwsProvider

non of the examples work

What step am I missing?

#!/usr/bin/env python
from imports.aws import Instance, AwsProvider

    super().__init__(scope, ns)

    AwsProvider(self, 'Aws', region='us-east-1')
    Instance(self, "hello", ami="ami-2757f631", instance_type="t2.micro")

app = App()
MyStack(app, "hello-terraform")

app.synth()

DanielMSchmidt:
Closes #612

### DIFF
--- a/docs/getting-started/csharp.md
+++ b/docs/getting-started/csharp.md
@@ -124,7 +124,9 @@ namespace MyCompany.MyApp
 
 ## Using Terraform Providers
 
-As you can see in the example above we want to deploy AWS resources, therefore we need to use the [Terraform AWS Provider](https://registry.terraform.io/providers/hashicorp/aws/latest/docs). The default way to work with a provider is to specify it in your local `cdktf.json` file, like this:
+As you can see in the example above we want to deploy AWS resources, therefore we need to use the [Terraform AWS Provider](https://registry.terraform.io/providers/hashicorp/aws/latest/docs). Every [Terraform Provider](https://registry.terraform.io/browse/providers) and [Terraform Module](https://registry.terraform.io/browse/modules) can be used with the workflow described here.
+
+You specify it in your local `cdktf.json` file, like this:
 
 ```jsonc
 {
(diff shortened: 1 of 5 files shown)


skorfmann:
Looks a bit funny when rendered 

Screenshot 2021-06-15 at 21 51 58

Other than that, I think it's good 👍

docs/getting-started/csharp.md
```diff
@@ -122,6 +122,30 @@ namespace MyCompany.MyApp
 }
 ```
 
+## Using Terraform Providers
+
+As you can see in the example above we want to deploy AWS resources, therefore we need to use the [Terraform AWS Provider](https://registry.terraform.io/providers/hashicorp/aws/latest/docs). The default way to work with a provider is to specify it in your local `cdktf.json` file, like this:
```
jsteinich, line 127: If this is really the default way, then I think we should change the templates back to include a provider (maybe even one chosen during init flow).

However, I actually believe that using pre-built providers will be the common case. Using just a major cloud provider and a couple auxiliary (random, null) providers will cover a large amount of workflows. In this case, we should really have the examples using pre-built providers, but that hinges some on #558.

I do like being more explicit about to specify modules/providers though.

DanielMSchmidt: Maybe default is the wrong wording here 🤔 I think a lot of people are confused on what providers / modules can be used. This is why I wanted to say default since all providers work this way. But I think you are totally right, a lot of use cases can be covered with prebuilt providers alone, so maybe I just say one way and mention it as another way 🤔

skorfmann: Yes, I'd default to pre-built providers and use customly built providers if there are either special version requirements for the provider, or the provider is not available (yet) as a pre-built provider.

DanielMSchmidt: I think I got this now, I made clear that every provider and module can be used while avoiding default entirely :)

skorfmann (approved):
Cool 👍

PR #786, opened 2021-06-18
chore(docs): Add serverless application guide

### LINKED ISSUE #733: End-to-end serverless workflow
### Community Note

### Description
This is a draft! Let's talk before starting work on this.

The purpose of this project is to design a good end-to-end workflow for building and deploying serverless applications to popular cloud providers. This is simultaneously a documentation and design exercise. In the first iteration, the documentation should describe how to implement these workflows using CDK for Terraform as it exists today.

While writing that documentation, consider what else we could do to simplify the workflow and make it more idiomatic in each language and cloud provider. File enhancement requests, bugs, and link them back to this issue.

Use case: user wants to deploy a simple serverless application without having to stand up or operate other infrastructure.

#### Goals
* All resources are defined in a single codebase
* Examples available for AWS Lambda, Azure Functions, and (stretch goal) Google Cloud Functions
* Deploy functions, load balancers, security groups, storage buckets etc as needed - no manual setup outside of `cdktf deploy` should be needed.
* The serverless functions, as well as the CDK for Terraform code examples, should be implemented in TypeScript, Python, and Go

#### Stretch goals
As time permits, more advanced permutations would also be helpful, such as:
* deploying and configuring persistence (db, storage buckets) including setting up authentication to the persistence
* separately deploying a toy backend API and front-end as separate containers
* Provide a clear developer workflow for publishing to development/staging/production environments
* When the deploy completes, show (using an output?) the URL the web application is available at

#### Ideas for sample applications
In order to illustrate serverless use cases, we need some concrete use cases that are representative of common scenarios. Below, I've proposed some ideas. I'll leave it to whoever picks up this ticket to define the exact cases. Please pick one or two and add a more detailed definition (edit this issue as needed) prior to starting work.

* Simple REST endpoint that uses the cloud provider's key-value store for persistence
  * possibly accompanied by a simple statically hosted (cloud storage bucket) front-end app that pulls data from the endpoint, to show how both can be deployed as separate stacks
* Recurring (e.g. cron job) lambda that polls something and stores the result (probably in cloud provider's key-value store)
* Simple file processing pipeline where:
  * a file is POSTed to a first HTTP endpoint, which saves it to a cloud storage bucket
  * the upload event triggers another function, which compresses (gzip is fine) the file and deletes the original uncompressed file
  * a third HTTP endpoint shows a list of compressed files which are available for download
  * every day any compressed files over 24 hours old are deleted

Please feel free to suggest other toy use cases instead of these! This is intended as a prompt.

### Deliverables
The immediate practical output will be end-to-end documentation published in this GitHub. File any ideas you have for improving it as enhancement requests linked to this issue for ease of tracking.

ansgarm:
Adds a guide explaining the end to end serverless example.

[rendered version](https://github.com/hashicorp/terraform-cdk/blob/add-serverless-application-guide/docs/full-guide/serverless-application-typescript.md)
Repo with end to end serverless example: https://github.com/hashicorp/cdktf-integration-serverless-example (also linked in guide)

What do you think having this guide inside the newly created `docs/full-guide` directory? Happy to move it elsewhere if that seems off.

Resolves #733

### DIFF
--- a/docs/full-guide/serverless-application-typescript.md
+++ b/docs/full-guide/serverless-application-typescript.md
@@ -1,9 +1,304 @@
-# Deploying a serverless application with CDK for Terraform using TypeScript
+# Building a serverless application with CDK for Terraform using TypeScript
 
-Use case: user wants to deploy a simple serverless application without having to stand up or operate other infrastructure.
+This guide walks you through a codebase that uses the CDK for Terraform to
+deploy a serverless application to AWS. The application features a small
+frontend written in [React](https://reactjs.org/) that connects to an API
+powered by AWS APIGateway and AWS Lambda.
 
-> This is a draft and work in progress
+The repository can be found here:
+https://github.com/hashicorp/cdktf-integration-serverless-example  
+(includes screenhots)
+
+There are also examples available using Docker containers on
+[AWS ECS](https://github.com/hashicorp/docker-on-aws-ecs-with-terraform-cdk-using-typescript)
+or [GCP K8S](https://github.com/hashicorp/kubernetes-on-gcp-with-terraform-cdk).
+
+## Preface
+
+This guide assumes basic familarity with the CDK for Terraform and that you have
+it installed already. If you are new to the CDKTF it is recommended to first
+have a look at the [getting started guide](../getting-started/typescript.md)
+which explains the CDK for Terraform itself in more detail and the commands used
+to deploy infrastructure. This guide will mainly explain the codebase of the
+serverless example and what it does.  
+For a simpler serverless example there is also a step by step tutorial on
+[HashiCorp Learn](https://learn.hashicorp.com/tutorials/terraform/cdktf-assets-stacks-lambda).
+
+## Further requirements
+
+- [`aws` CLI](https://docs.aws.amazon.com/cli/latest/userguide/install-cliv1.html)  
+  The deployment of the frontend requires the AWS CLI to be installed on your
+  system.
+
+## Init
+
+The project was initialized using the following `cdktf init` command:
 
 ```
 cdktf init --local --template=typescript
 ```
+
+You can either clone the whole
+[repository](https://github.com/hashicorp/cdktf-integration-serverless-example)
+or initialize a new one with the aforementioned command and copy the code from
+the example repo as we move along and discuss it.
+
+### AWS Credentials
+
+The example uses environment variables to configure the access credentials for
+the AWS Terraform provider. You can refer to
+[its docs](https://registry.terraform.io/providers/hashicorp/aws/latest/docs#environment-variables)
+to see which variables to set.
+
+## Stacks
+
+in the root directory of the example you can find the `main.ts` file which
+contains the configuration for the stacks that are going to be deployed. A stack
+has its own Terraform state and is deployed separately from other stacks.  
+The example consists of three stacks:
+
+- FrontendStack
+- PostsStack
+- PreviewStack
+
+### FrontendStack
+
+The frontend stack manages infrastructure for a statically hosted web frontend
+and deploys that frontend via a short script.  
+The frontend itself is set up with
+[Create React App](https://create-react-app.dev/) inside the `frontend/code`
+directory and its build output is by default written into the
+`frontend/code/dist` directory. This guide will not cover the React frontend
+itself. See the "Frontend" section below for specifics.
+
+### PostsStack
+
+The posts stack manages the required infrastructure making up the serverless
+posts api. Its implementation is located inside the `posts` directory and is
+described in detail in the "Posts API" section below.
+
+### PreviewStack
+
+The preview stack is not yet used but is supposed to show that both the frontend
+and the posts API could be deployed at once while sharing a single Terraform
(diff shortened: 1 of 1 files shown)


jsteinich:
Looks interesting. Some possible future changes to the example:

- Lock down the S3 bucket so that all requests need to go through CloudFront.
- Use the S3 bucket object resource instead of syncing
- Use [IAM Floyd](https://www.npmjs.com/package/iam-floyd) or similar to provide code first iam policies
- Call out that the different environments are different stack

ansgarm:
Thank you for the great feedback, @jsteinich!

>  Lock down the S3 bucket so that all requests need to go through CloudFront.

We need to use the web hosting feature of the S3 bucket to serve the `index.html` for all unknown paths (due to the frontend being a single page application). I could not find a way to restrict the S3 bucket website hosting to only CloudFront and directly serving the files does not work for the application. Is there a way to achieve this that I overlooked?

> Use the S3 bucket object resource instead of syncing

I thought about that as well. However it didn't feel right. Currently the frontend has ~10ish files that are deployed to the S3 Bucket, however that number could grow substantially for a bigger application. This would result in the same number of (generated) S3 bucket object resources in Terraform.  
One alternative could be to build a custom resource / custom Terraform resource which takes care of the copying. As far as I know the AWS CDK bundles files in a zip and uses a Lambda that extracts the files into an S3 Bucket for a similar use case.


> Use [IAM Floyd](https://www.npmjs.com/package/iam-floyd) or similar to provide code first iam policies

Great idea. [Noted](https://github.com/hashicorp/cdktf-integration-serverless-example/commit/6b0fb158139ca50d26641cbe1cdc93c39314e77c).

>  Call out that the different environments are different stack

[Done](https://github.com/hashicorp/terraform-cdk/commit/f89dc8c6a9a2477fe25cda0f10ec498ff2420dc5).

jsteinich:
> I thought about that as well. However it didn't feel right. Currently the frontend has ~10ish files that are deployed to the S3 Bucket,

Makes sense. For some reason I was thinking it was single file

jsteinich:
> I could not find a way to restrict the S3 bucket website hosting to only CloudFront and directly serving the files does not work for the application. Is there a way to achieve this that I overlooked?

Here's some code that we have:
```c#
private VariableConditionResource<CloudfrontOriginAccessIdentity> ApplyBucketPolicy(S3Bucket bucket)
        {
            {//public
                var publicBucketPolicy = new PolicyDocument
                {
                    Statements = {new S3().ToGetObject().OnObject(Config.WebSiteDomain, "*").ForPublic()}
                };

                new VariableConditionResource<S3BucketPolicy>(!Config.UseCloudFront,
                    new S3BucketPolicy(this, "public_s3_policy", new S3BucketPolicyConfig
                    {
                        Bucket = bucket.Id,
                        Policy = publicBucketPolicy.ToJson()
                    }), true);
            }

            {//cloudfront
                new VariableConditionResource<S3BucketPublicAccessBlock>(Config.UseCloudFront,
                    new S3BucketPublicAccessBlock(this, "access_block", new S3BucketPublicAccessBlockConfig
                    {
                        Bucket = bucket.Id,
                        IgnorePublicAcls = true
                    }));

                var accessIdentity = new VariableConditionResource<CloudfrontOriginAccessIdentity>(Config.UseCloudFront,
                    new CloudfrontOriginAccessIdentity(this, "cf_access_identity", new CloudfrontOriginAccessIdentityConfig
                    {
                        Comment = "Access S3 bucket content only through CloudFront"
                    }));

                var cfBucketPolicy = new PolicyDocument
                {
                    Statements =
                    {
                        new S3().ToGetObject().OnObject(Config.WebSiteDomain, "*")
                            .For(accessIdentity.StringValue(nameof(accessIdentity.Resource.IamArn)))
                    }
                };
                
                new VariableConditionResource<S3BucketPolicy>(Config.UseCloudFront,
                    new S3BucketPolicy(this, "cf_s3_policy", new S3BucketPolicyConfig
                    {
                        Bucket = bucket.Id,
                        Policy = cfBucketPolicy.ToJson()
                    }));

                return accessIdentity;
            }
        }

//within cloud front distribution
Origin = new ICloudfrontDistributionOrigin[]
                {
                    new CloudfrontDistributionOrigin
                    {
                        DomainName = bucket.BucketDomainName,
                        OriginId = $"s3-origin-{bucket.Id}",
                        OriginPath = "",
                        
                        S3OriginConfig = new ICloudfrontDistributionOriginS3OriginConfig[]
                        {
                            new CloudfrontDistributionOriginS3OriginConfig
                            {
                                OriginAccessIdentity = cfAccessIdentity.StringValue(nameof(cfAccessIdentity.Resource.CloudfrontAccessIdentityPath))
                            }
                        }
                    }
                },
```

ansgarm:
Thanks @jsteinich!
[Added it](https://github.com/hashicorp/cdktf-integration-serverless-example/commit/195737c51d79fe253377ee24fd5288f4be163f3b) to the ideas for future iterations.

DanielMSchmidt (approved):
Quite a different style from my version, but I like it!

PR #796, opened 2021-06-24
feat(cli): implement hcl2cdk

DanielMSchmidt:
`pbpaste | cdktf convert --language=...` should print the CDKTF-ified version of what you have in your clipboard

convert-bigger


packages/@cdktf/hcl2cdk/lib/schema.ts
```diff
@@ -0,0 +1,26 @@
+import * as z from "zod";
+
+const outputConfig = z.array(
+  z.object({
```
jsteinich, line 4: Outputs also support depends_on

DanielMSchmidt: Thank you!

packages/@cdktf/hcl2cdk/lib/schema.ts
```diff
@@ -0,0 +1,43 @@
...
+
+const outputConfig = tfObject({
+  value: z.string(),
+  description: z.string().optional(),
+  sensitive: z.boolean().optional(),
+  depends_on: z.string().optional(),
+});
+export type Output = z.infer<typeof outputConfig>;
+
+const variableConfig = tfObject({
```
jsteinich, line 15: Could have `validation`, though we don't currently support that.

packages/@cdktf/hcl2cdk/lib/schema.ts
```diff
@@ -0,0 +1,43 @@
...
+
+const moduleConfig = z.array(z.object({ source: z.string() }).nonstrict());
+export type Module = z.infer<typeof moduleConfig>;
+
+const resourceConfig = z.array(z.record(z.any()));
+export type Resource = z.infer<typeof resourceConfig>;
+export type Data = Resource;
+
+export const schema = z
+  .object({
```
jsteinich, line 34: What about the [Terraform block](https://www.terraform.io/docs/language/syntax/json.html#terraform-blocks)?

packages/@cdktf/hcl2cdk/lib/index.ts
```diff
@@ -0,0 +1,470 @@
...
+  key: string,
+  _id: string,
+  item: Provider,
+  graph: DirectedGraph
+): t.Statement {
+  const nodeIds = graph.nodes();
+  const props = item[0];
+
+  if (props.alias) {
+    throw new Error(
```
jsteinich, line 202: Would this be hard to add?

jsteinich (commented):
I'd be happy to try this out on some of our Terraform projects once it is at that point. Just let me know.

packages/cdktf-cli/bin/cmds/convert.ts
```diff
@@ -0,0 +1,66 @@
...
+      })
+      .option("file", {
+        alias: "f",
+      })
+      .showHelpOnFail(true);
+
+  public async handler({ language, file }: any) {
+    await displayVersionMessage();
+
+    const input = await readStreamAsString(process.stdin);
```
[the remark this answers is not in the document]
DanielMSchmidt: That's the next step. This part can be useful IMHO for a quick snippet from an example.
I'm planning on adding an import command that bootstraps a new project based on the input directory

packages/@cdktf/hcl2cdk/lib/schema.ts
DanielMSchmidt: That we will need for sure! I'm adding this when I'm working on the import command, we need the data to construct a complete cdktf.json modules / providers list

packages/@cdktf/hcl2cdk/lib/index.ts
DanielMSchmidt: No, I just wanted to be explicit about what is currently supported :)

skorfmann (commented):
That's pretty amazing, looks great! Built and tried it locally as well, seems to work as expected in general. For complex examples, there are a few issues though:

```hcl
resource "aws_acm_certificate" "example" {
  domain_name       = "example.com"
  validation_method = "DNS"
}

data "aws_route53_zone" "example" {
  name         = "example.com"
  private_zone = false
}

resource "aws_route53_record" "example" {
  for_each = {
    for dvo in aws_acm_certificate.example.domain_validation_options : dvo.domain_name => {
      name   = dvo.resource_record_name
      record = dvo.resource_record_value
      type   = dvo.resource_record_type
    }
  }

  allow_overwrite = true
  name            = each.value.name
  records         = [each.value.record]
  ttl             = 60
  type            = each.value.type
  zone_id         = data.aws_route53_zone.example.zone_id
}

resource "aws_acm_certificate_validation" "example" {
  certificate_arn         = aws_acm_certificate.example.arn
  validation_record_fqdns = [for record in aws_route53_record.example : record.fqdn]
}

resource "aws_lb_listener" "example" {
  # ... other configuration ...

  certificate_arn = aws_acm_certificate_validation.example.certificate_arn
}
```

produces 

```ts
const awsAcmCertificateExample = new aws.AcmCertificate(this, "example", {
  domainName: "example.com",
  validationMethod: "DNS",
});
const dataAwsRoute53Zone = new aws.Route53Zone(this, "example", {
  name: "example.com",
  privateZone: false,
});
const awsRoute53RecordExample = new aws.Route53Record(this, "example", {
  allowOverwrite: true,
  name: "${each.value.name}",
  records: ["${each.value.record}"],
  ttl: 60,
  type: "${each.value.type}",
  zoneId: dataAwsRoute53Zone.example.zoneId,
});
awsRoute53RecordAwsRoute53RecordExample.addOverride(
  "for_each",
  `\${{
    for dvo in ${awsAcmCertificateExample.domainValidationOptions} : dvo.domain_name => {
      name   = dvo.resource_record_name
      record = dvo.resource_record_value
      type   = dvo.resource_record_type
    }
  }}`
);
const awsAcmCertificateValidationExample = new aws.AcmCertificateValidation(
  this,
  "example",
  {
    certificateArn: awsAcmCertificateExample.arn,
    validationRecordFqdns: `\${[for record in ${awsRoute53RecordExample.fqn} : record.fqdn]}`,
  }
);
new aws.LbListener(this, "example", {
  certificateArn: awsAcmCertificateValidationExample.certificateArn,
});
```

The thing which doesn't fit there:

```ts
const awsRoute53RecordExample = new aws.Route53Record(this, "example", {
  allowOverwrite: true,
  name: "${each.value.name}",
  records: ["${each.value.record}"],
  ttl: 60,
  type: "${each.value.type}",
  zoneId: dataAwsRoute53Zone.example.zoneId,
});
awsRoute53RecordAwsRoute53RecordExample.addOverride(
  "for_each",
  `\${{
    for dvo in ${awsAcmCertificateExample.domainValidationOptions} : dvo.domain_name => {
      name   = dvo.resource_record_name
      record = dvo.resource_record_value
      type   = dvo.resource_record_type
    }
  }}`
);
```

Also, the `imports` attribute seems to be empty in this case. I'd have expected that the AWS provider would have been returned for this example.

DanielMSchmidt:
> Also, the imports attribute seems to be empty in this case. I'd have expected that the AWS provider would have been returned for this example.

Currently I find the provider from the provider config. The imports fields goal is more in the import complete project part, the code part is for on the spot conversion :) Does this make sense?

> The thing which doesn't fit there:

Sorry I'm blind to the error, could you push a failing snapshot test for me to fix? :D

skorfmann:
The override combines resource class name with the variable name: `awsRoute53RecordAwsRoute53RecordExample` =>  `Route53Record ` + `awsRoute53RecordExample`

DanielMSchmidt:
@skorfmann Ah thank you! Fixed that one :)

docs/cli-commands.md
```diff
@@ -17,22 +17,32 @@ $ cdktf --help
...
-  cdktf login              Retrieves an API token to connect to Terraform Cloud.
-  cdktf synth [OPTIONS]    Synthesizes Terraform code for the given app in a directory.                                                                                                    [aliases: synthesize]
+  cdktf convert [OPTIONS]                        Converts a single file of HCL configuration to Terraform
+                                                 CDK. Takes the file to be converted on stdin.
+  cdktf deploy [stack] [OPTIONS]                 Deploy the given stack                      [aliases: apply]
+  cdktf destroy [stack] [OPTIONS]                Destroy the given stack
+  cdktf diff [stack] [OPTIONS]                   Perform a diff (terraform plan) for the given stack
+                                                                                              [aliases: plan]
+  cdktf get [OPTIONS]                            Generate CDK Constructs for Terraform providers and modules.
+  cdktf import [OPTIONS] <source> <destination>  Takes the Terraform project and converts it to a CDKTF
```
danieldreier, line 28: I think we talked about this previously, but this shouldn't be called `import` since `import` is used in core to import state and that would be confusing

jsteinich:
I gave this a couple test runs and so far running into a couple issues:

1. My first attempt was a project still using the Terraform 0.12 required_providers syntax. The error I got had `invalid_union at terraform.0.required_providers`. Would it be possible to get line numbers for the error?
2. After switching the provider definition, I ran into an issue with `terraform.workspace` being an unknown reference. There are a couple other built in references as well. See https://www.terraform.io/docs/language/expressions/references.html#filesystem-and-workspace-info
3. Next I ran into problems accessing a module defined in a git repo. It changed casing and removed the repo url and is trying to import that

packages/@cdktf/hcl2cdk/lib/schema.ts
```diff
@@ -0,0 +1,55 @@
+import * as z from "zod";
+import { ZodRawShape } from "zod/lib/src/types/base";
+
+const tfObject = <T extends ZodRawShape>(config: T) =>
+  z.array(z.object(config).partial());
+
+const outputConfig = tfObject({
+  value: z.string(),
```
jsteinich, line 8: The value of an output doesn't need to be a string. Any expression is valid.

docs/cli-commands.md
DanielMSchmidt: Thanks for pointing this out, I forgot to update the docs here :)

DanielMSchmidt:
Hey @jsteinich thanks for taking the time to review 🙇 

> My first attempt was a project still using the Terraform 0.12 required_providers syntax. The error I got had invalid_union at terraform.0.required_providers. Would it be possible to get line numbers for the error?

I don't think we can do line numbers, these get swallowed by the hcl2json tool we use. This command is build with terraform 1.0 in mind, I'll not that in the docs 👍 

> After switching the provider definition, I ran into an issue with terraform.workspace being an unknown reference. There are a couple other built in references as well

Oh wow, I just knew the path one and dealt with that, terraform.workspace was unknown to me 🤯 I'll add a testcase for this


> Next I ran into problems accessing a module defined in a git repo. It changed casing and removed the repo url and is trying to import that

Another case I completely forgot existed, I will add a testcase for this as well :)

packages/@cdktf/hcl2cdk/lib/expressions.ts
```diff
@@ -0,0 +1,390 @@
...
+    );
+  }
+
+  // resources or data sources
+  if (!type.includes("./") && type.includes(".")) {
+    const parts = type.split(".");
+    if (parts[0] === "data") {
+      return t.memberExpression(
+        t.identifier(parts[1]), // e.g. aws
+        t.identifier(pascalCase(`data_${parts[1]}_${parts[2]}`)) // e.g. NatGateway
```
ansgarm, line 212: "e.g. NatGateway" is a wrong example, isn't it? 😅 Looks like copy paste 😄

packages/@cdktf/hcl2cdk/lib/expressions.ts
```diff
@@ -0,0 +1,390 @@
+import * as t from "@babel/types";
+import { camelCase, pascalCase } from "./utils";
+import { TerraformResourceBlock } from "./types";
+
+export type Reference = {
+  start: number;
+  end: number;
+  referencee: { id: string; full: string }; // identifier for resource
```
ansgarm, line 8: some examples for `id` and `full` would be nice

packages/@cdktf/hcl2cdk/lib/expressions.ts
```diff
@@ -0,0 +1,390 @@
...
+  if (
+    refAsts.length === 1 &&
+    refAsts[0].ref.start === "${".length &&
+    refAsts[0].ref.end === input.length - "}".length &&
+    !refAsts[0].ref.useFqn
+  ) {
+    return refAsts[0].ast;
+  }
+
+  // string parts in the template string
```
ansgarm, line 280: Nice comment 👍

ansgarm (commented):
Did not quite make it through yet, but here is a first set of comments already 👍
I really like the new commands you added to the code after our video review 💪

packages/@cdktf/hcl2cdk/README.md
```diff
@@ -0,0 +1,72 @@
+# HCL2CDK
```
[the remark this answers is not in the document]
DanielMSchmidt: I like it being prefixed so the to cdk part is not ambiguous anymore

packages/@cdktf/hcl2cdk/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Known Limitations
+
+### Terraform Expressions are of the wrong TS/Java/C# type
+
+When working with typed languages the converter can run into problems where the Terraform Expression evaluates to a certain type but it's encoded in a string. Therefore the type checker of the language detects a type mismatch, resulting in an compilation error. These problems need to be manually solved by adding a typecast. One example would be:
+
+```ts
+{
+  booleanProperty: `\${${shouldBeTrue.value} ? true : false}`;
```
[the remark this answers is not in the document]
DanielMSchmidt: The point is that TS thinks the outcome is a string, but for HCL that string represents a boolean. I added some comments hope that clarifies it

packages/cdktf-cli/bin/cmds/convert.ts
```diff
@@ -0,0 +1,67 @@
...
+      stream.on("close", () => ok(string));
+      stream.on("error", (err) => ko(err));
+    }
+  });
+}
+
+class Command implements yargs.CommandModule {
+  public readonly command = "convert [OPTIONS]";
+  public readonly describe =
+    "Converts a single file of HCL configuration to Terraform CDK. Takes the file to be converted on stdin.";
```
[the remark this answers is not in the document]
DanielMSchmidt: Fine with me, I'm always confused as I see Terraform CDK, CDK for Terraform and cdktf used interchangeably :D

packages/cdktf-cli/bin/cmds/helper/init.ts
```diff
@@ -0,0 +1,470 @@
...
+        );
+
+        telemetryData.conversionStats = stats;
+      } catch (err) {
+        throw Errors.Internal("init", err, { fromTerraformProject: true });
+      }
+      execSync("npm run get", { cwd: destination });
+    } else {
+      console.error(
+        `The --from-terraform-project flag is only support with the typescript template. The command will continue and ignore the flag.`
```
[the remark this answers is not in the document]
DanielMSchmidt: I thought the next logical step after a fail here would be to retry without the flag, so having it as no-op seemed like a better fit

packages/@cdktf/hcl2cdk/lib/expressions.ts
DanielMSchmidt: Those can be found in the tests, do you think that's too hard to spot?

ansgarm: No, that's reasonable 👍

ansgarm (approved):
Nice work 💱

packages/@cdktf/hcl2cdk/test/convertProject.test.ts
```diff
@@ -0,0 +1,215 @@
...
+        language: "typescript",
+      }
+    );
+
+    fs.writeFileSync(path.resolve(targetPath, "main.ts"), code, "utf8");
+    fs.writeFileSync(path.resolve(targetPath, "cdktf.json"), cdktfJson, "utf8");
+
+    const currentPlan = getCdkPlan(targetPath);
+
+    expect(resources(currentPlan)).toEqual(resources(previousPlan));
```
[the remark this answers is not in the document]
DanielMSchmidt: Yeah, I'm really proud that this one worked 🎉

jsteinich:
> thanks for taking the time to review

I didn't actually review the code, just started trying it out on some real Terraform projects.

I tried it again this morning and ran into some new issues:

- In a project without any providers, it still tries to run cdktf get, which fails. Probably an unusual case right now, but if we update to use pre-built providers, then will likely be common.
- The git modules failed with cdktf get. Not directly related to this. More of an error with the module auto naming, Could probably generate cdktf.json with the name override matching what it was called in hcl.

DanielMSchmidt:
> In a project without any providers, it still tries to run cdktf get, which fails. Probably an unusual case right now, but if we update to use pre-built providers, then will likely be common.

Oh great point, I wonder if we should change this behaviour in get in general 🤔 but for now I can add an extra check 👍 

> The git modules failed with cdktf get. Not directly related to this. More of an error with the module auto naming, Could probably generate cdktf.json with the name override matching what it was called in hcl.

Did you try the latest version? I changed the behaviour to use `TerraformHclModule` for everything besides registry modules

jsteinich:
> I wonder if we should change this behaviour in get in general

I think so, but I was viewing that as a follow up task.



> Did you try the latest version? I changed the behaviour to use `TerraformHclModule` for everything besides registry modules

Yep. The code is updated, but it is still present in cdktf.json

DanielMSchmidt:
> Yep. The code is updated, but it is still present in cdktf.json

That is really weird, I'm using the same check for both 🤔 https://github.com/hashicorp/terraform-cdk/blob/67d261dedc880fbdf4bff1cc3f07988bf1477a84/packages/@cdktf/hcl2cdk/lib/generation.ts#L423

jsteinich:
> That is really weird, I'm using the same check for both

I'm thinking it's still in `moduleRequirements`.

DanielMSchmidt:
Oh right, thank you! Forgot the third part, will do a PR :)

PR #818, opened 2021-07-01
chore(cli): lint tsx files and clean up React Hooks code

ansgarm:
Noticed some code smells with React Hooks.  
E.g. `synth` function [being a hook](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-from-react-functions), because it called an `useEffect` hook, but not prefixed with `use` ([docs](https://reactjs.org/docs/hooks-custom.html#extracting-a-custom-hook)).

### DIFF
--- a/packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx
+++ b/packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx
@@ -321,13 +321,6 @@ export const TerraformProvider: React.FunctionComponent<TerraformProviderConfig>
       </TerraformContextState.Provider>
     );
   };
-interface UseTerraformInput {
-  targetDir: string;
-  targetStack?: string;
-  synthCommand: string;
-  isSpeculative?: boolean;
-  autoApprove?: boolean;
-}
 
 export const useTerraformState = () => {
   const state = React.useContext(TerraformContextState);
@@ -341,18 +334,8 @@ export const useTerraformState = () => {
   return state;
 };
 
-export const useTerraform = ({
-  targetDir,
-  targetStack,
-  synthCommand,
-  isSpeculative = false,
-  autoApprove = false,
-}: UseTerraformInput) => {
+const useTerraformDispatch = () => {
   const dispatch = React.useContext(TerraformContextDispatch);
-  const state = useTerraformState();
-  const [terraform, setTerraform] = React.useState<Terraform>();
-  const [confirmed, setConfirmed] = React.useState<boolean>(autoApprove);
-  const { exit } = useApp();
 
   if (dispatch === undefined) {
     throw new Error(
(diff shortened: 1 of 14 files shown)


packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx
```diff
@@ -431,7 +448,10 @@ export const useTerraform = ({
 
   const execTerraformInit = async () => {
     try {
-      if (!terraform) throw new Error("Terraform is not initialized yet");
+      if (!terraform)
+        throw new Error(
+          "Terraform is not initialized yet. Please call synth() with loadExecutor: true first"
```
DanielMSchmidt, line 453: I think this is a great internal error message, but if it makes its way to the user it could be really confusing. Is there something a user can do?

packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx
```diff
@@ -514,123 +534,131 @@ export const useTerraform = ({
...
+};
 
-  const init = () => {
-    React.useEffect(() => {
-      const invoke = async () => {
-        await execTerraformSynth();
-      };
-      invoke();
-    }, []);
+const useRunOnce = <Fn extends (...args: any[]) => any>(
```
DanielMSchmidt, line 547: Nitpick: We don't necessarily need to use any here :)

```suggestion
const useRunOnce = <T, TR, Fn extends (...args: T[]) => TR>(
```

packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx
```diff
@@ -514,123 +534,131 @@ export const useTerraform = ({
...
-        await execTerraformOutput();
-      };
-      if (confirmed && state.status === Status.PLANNED) invoke();
-    }, [terraform, confirmed, state.status]);
+  useRunOnce(synth);
+  useRunIf(state.status === Status.SYNTHESIZED, async () => {
+    await init();
+    await diff();
+  });
+  // TODO: check: old effect had reference to [terraform] aswell
```
DanielMSchmidt, line 599: Can this go now?

packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx
```diff
@@ -514,123 +534,131 @@ export const useTerraform = ({
...
+    fn(...args);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []); // only run once on mount – ignore any changes to fn
+};
 
-    React.useEffect(() => {
-      const invoke = async () => {
-        await execTerraformInit();
-      };
+const useRunIf = <Fn extends (...args: any[]) => any>(
```
DanielMSchmidt, line 557: Nitpick: If indicates a conditional for me, but in this case it's mostly temporal, right? So like useRunWhen?

packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx
ansgarm: Ah, nice!

packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx
ansgarm: I'm torn – a user should never see this message in the first place (same issue with the original error message as there is no `cdktf init` which the user could use to resolve Terraform not being initialized).

packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx
ansgarm: Hm, technically it waits for a boolean value which seems more like "if" but in practice it is going to be a temporal thing. Going to adjust it to `When` – I think that suits the context better 👍

packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx
ansgarm: Oh, yeah, it can.

packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx
ansgarm: Hm, that seems to be incomplete or am I missing something?

image

packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx
DanielMSchmidt: I see your point and I always remember the anecdote my mom told me where in an important client meeting a very internal error message was displayed :D 

Maybe we can prefix it with ÌnternalError to indicate this is not a user-facing error? I'm fine with postponing this and making it a separate issue :)

packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx
DanielMSchmidt: Hmm, seems like the type narrowing does not work as I thought it would, let's keep the anys for now

DanielMSchmidt (approved):

packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx
ansgarm: Good idea, prefixed it with "Internal error: "

Issue #234, opened 2020-07-22
Inject Defaults during Code Generation


skorfmann:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

There's at least one occasion where Terraform JSON isn't fully compatible with HCL. https://github.com/terraform-providers/terraform-provider-aws/issues/8786#issuecomment-496935442 - It's some cases expected to provide explicit `null` values for optional attributes, e.g:

<details><summary>A Security Group example by @pmartindev</summary><p>

```ts
 const ecstasksg = new SecurityGroup(this, 'ecstasksg', {
      name: "tf-ecs-task-sg",
      description: "Allows traffic to the ALB.", 
      ingress: [{
        protocol: "tcp",
        fromPort: 80,
        toPort: 80,
        cidrBlocks: ["0.0.0.0/0"]
      },{
        protocol: "tcp",
        fromPort: 22,
        toPort: 22,
        cidrBlocks: ["0.0.0.0/0"]
      },{
        protocol: "tcp",
        fromPort: 443,
        toPort: 443,
        cidrBlocks: ["0.0.0.0/0"]
      }],
      egress: [{
        protocol: "-1",
        fromPort: 0,
        toPort: 0,
        cidrBlocks: ["0.0.0.0/0"]
      }],
    });

    // Ingress
    ecstasksg.addOverride("ingress.0.description", null);
    ecstasksg.addOverride("ingress.0.ipv6_cidr_blocks", null);
    ecstasksg.addOverride("ingress.0.prefix_list_ids", null);
    ecstasksg.addOverride("ingress.0.security_groups", null);
    ecstasksg.addOverride("ingress.0.selfAttribute", null);
    ecstasksg.addOverride("ingress.0.self", null);
    ecstasksg.addOverride("ingress.1.description", null);
    ecstasksg.addOverride("ingress.1.ipv6_cidr_blocks", null);
    ecstasksg.addOverride("ingress.1.prefix_list_ids", null);
    ecstasksg.addOverride("ingress.1.security_groups", null);
    ecstasksg.addOverride("ingress.1.selfAttribute", null);
[...]

PR #821, opened 2021-07-03
fix(lib): Add in new optional AWS route attributes

### LINKED ISSUE #813: Enable null for mapped attributes
### Community Note

### cdktf & Language Versions

Probably all 

### Affected Resource(s)

AWS Route Table

### Important Factoids

As written by @wuntusk in https://github.com/terraform-cdk-providers/cdktf-provider-aws/pull/394

> Currently RouteTable is unuseable
>
> Currently you can't make a RouteTable since all routes require both a carrierGatewayId and a destinationPrefixListId. For some reason these 2 values never got an undefined check added so they never pass null but instead make a call to cdktf.stringToTerraform. I'm guessing this has never been caught as everyone is using the vpc module shown in the demos. If you try to roll your own pure Typescript VPC you'll run into this pretty quickly.

### References

Potentially related to

- https://github.com/hashicorp/terraform-cdk/issues/750
- https://github.com/hashicorp/terraform-cdk/issues/234
- https://github.com/hashicorp/terraform-cdk/pull/395

jsteinich:
Fixes #813 

This just adds in the reported missing defaults.

Would be nice if Terraform were modified to either not work this way when using json or to provide the information in the json schema.

### DIFF
--- a/packages/cdktf-cli/lib/get/generator/custom-defaults.ts
+++ b/packages/cdktf-cli/lib/get/generator/custom-defaults.ts
@@ -28,6 +28,7 @@ export const CUSTOM_DEFAULTS: { [fullName: string]: any } = {
   "aws.default_route_table.route.transit_gateway_id": null,
   "aws.default_route_table.route.vpc_endpoint_id": null,
   "aws.default_route_table.route.vpc_peering_connection_id": null,
+  "aws.default_route_table.route.destination_prefix_list_id": null,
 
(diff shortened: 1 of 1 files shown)


skorfmann:
> Would be nice if Terraform were modified to either not work this way when using json or to provide the information in the json schema.

Yes! For reference, this was the original issue for the underlying problem https://github.com/hashicorp/terraform-cdk/issues/234

DanielMSchmidt (approved):

PR #829, opened 2021-07-12
chore(cli): send telemetry on errors

DanielMSchmidt:
This PR helps us track errors to understand the problems of our user-base better. 
I added the Errors object in an adjecent effort of being deliberate about Error types


ansgarm (commented):
Only two minor things, but good to go then 🚀

ansgarm (approved):

Issue #826, opened 2021-07-09
Improve Telemetry


DanielMSchmidt:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

To improve the project we need to know what problems people are running into. The main areas of improvement are:

- [x] Errors (ideally categorised)
- [x] Escape Hatch usage
- [x] Backend usage

Additionally we need context data that helps us understand the problems folks run to and their settings

- [ ] Users (e.g. a [hashed machine id](https://github.com/automation-stack/node-machine-id))
- [ ] Codebases (e.g. a hashed git origin)
- [ ] CIs used (e.g. wth https://github.com/npm/ci-detect)

PR #830, opened 2021-07-12
chore(lib): track overrides and backend types in stack

DanielMSchmidt:
Telemetry like this is sent, giving us insights around backend types and overrides

```json
{
  "command": "synth",
  "product": "cdktf",
  "version": "0.0.0",
  "dateTime": "2021-07-12T17:17:06.396Z",
  "payload": {
    "command": "npm run --silent compile && node main.js",
    "totalTime": 4461.242155000567,
    "stackMetadata": [
      {
        "version": "0.0.0",
        "stackName": "demo-cdktf-ts-docker",
        "backendType": "local",
        "overrides": {
          "docker_image": [
            "build"
          ]
        }
      }
    ]
  },
  "runID": "62a8db03-d5cd-43bb-8e6e-cdbfce7049db",
  "arch": "x64",
  "os": "darwin"
}
```

### DIFF
--- a/packages/cdktf-cli/bin/cmds/helper/synth-stack.ts
+++ b/packages/cdktf-cli/bin/cmds/helper/synth-stack.ts
@@ -98,7 +98,6 @@ Command output on stdout:
 
     // end performance timer
     const endTime = performance.now();
-    await this.synthTelemetry(command, endTime - startTime, synthOrigin);
 
     const stacks: SynthesizedStack[] = [];
     const manifest = JSON.parse(
@@ -117,6 +116,13 @@ Command output on stdout:
         content: JSON.stringify(jsonContent, null, 2),
       });
     }
+
+    await this.synthTelemetry(
+      command,
+      endTime - startTime,
+      stacks,
+      synthOrigin
+    );
 
     if (stacks.length === 0) {
       console.error("ERROR: No Terraform code synthesized.");
@@ -137,12 +143,16 @@ Command output on stdout:
   public static async synthTelemetry(
     command: string,
     totalTime: number,
+    stacks: SynthesizedStack[],
     synthOrigin?: SynthOrigin
   ): Promise<void> {
     await sendTelemetry("synth", {
       command: command,
       totalTime: totalTime,
       synthOrigin,
+      stackMetadata: stacks.map(
+        (stack) => JSON.parse(stack.content)["//"].metadata
+      ),
     });
   }
 
--- a/packages/cdktf/lib/terraform-backend.ts
+++ b/packages/cdktf/lib/terraform-backend.ts
@@ -31,6 +31,9 @@ export abstract class TerraformBackend extends TerraformElement {
   public toMetadata(): any {
     return {
       backend: this.name,
+      ...(Object.keys(this.rawOverrides)
+        ? { overrides: { backend: this.rawOverrides } }
+        : {}),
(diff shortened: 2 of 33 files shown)


packages/cdktf/lib/terraform-resource.ts
```diff
@@ -127,6 +127,18 @@ export class TerraformResource
...
   }
 
+  public toMetadata(): any {
+    if (!Object.keys(this.rawOverrides).length) {
+      return {};
+    }
+
+    return {
+      overrides: {
+        [this.terraformResourceType]: Object.keys(this.rawOverrides),
```
jsteinich, line 137: Overrides exist on all `TerraformElements`; why only record for resources?

packages/cdktf/lib/terraform-stack.ts
```diff
@@ -140,18 +141,24 @@ export class TerraformStack extends Construct {
...
-    const tf = {
-      "//": {
-        metadata: {
-          version: this.cdktfVersion,
-          stackName: Node.of(this).id,
-        } as TerraformStackMetadata,
-      },
+    const tf = {};
+
+    const metadata: TerraformStackMetadata = {
```
jsteinich, line 146: Overrides can also be set at the stack level

packages/cdktf-cli/bin/cmds/helper/synth-stack.ts
```diff
@@ -117,14 +118,21 @@ Command output on stderr:
...
+    stacks: SynthesizedStack[]
   ): Promise<void> {
     const reportParams: ReportParams = {
       command: "synth",
       product: "cdktf",
       version: versionNumber(),
       dateTime: new Date(),
-      payload: { command: command, totalTime: totalTime },
+      payload: {
+        command: command,
```
jsteinich, line 130: I guess this isn't new, but could it potentially be exposing information that users may not want to share?

packages/cdktf/lib/terraform-resource.ts
DanielMSchmidt: Yeah good point, I'll add them across the board!

packages/cdktf-cli/bin/cmds/helper/synth-stack.ts
DanielMSchmidt: I think we need to revise what to track in general, afaik @danieldreier is making a plan here. I'd say we leave it as is for now and improve in the next iteration. #826 is the issue tracking the entire effort, we'll update this with the concrete scope also in regards of where we might be tracking secrets or similar things by accident.

packages/cdktf/lib/terraform-backend.ts
```diff
@@ -27,4 +27,13 @@ export abstract class TerraformBackend extends TerraformElement {
       },
     };
   }
+
+  public toMetadata(): any {
+    return {
+      backend: this.name,
+      ...(Object.keys(this.rawOverrides)
+        ? { overrides: { backend: this.rawOverrides } }
```
ansgarm, line 35: As this is not using `Object.keys()` here, couldn't it leak the values of the overrides? Is that a mistake or on purpose?

DanielMSchmidt: Ah forgot it here :)

ansgarm (approved):

Issue #835, opened 2021-07-15
Property Mapper should Respect Tokens


skorfmann:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### cdktf & Language Versions

- all

### Affected Resource(s)

- Lazy / Token

### Debug Output

```
node_modules/cdktf/lib/runtime.js:16
            return x.map(elementMapper);
                     ^

    TypeError: x.map is not a function
```

### Expected Behavior

`IResolvable` should be skipped here as well https://github.com/hashicorp/terraform-cdk/blob/d72fc072d3e6cf0a27ddef135942bb32f5ee8eb5/packages/cdktf/lib/runtime.ts#L50

### Actual Behavior

It's trying to map `IResolvable` values, which fails.

PR #836, opened 2021-07-15
fix(lib): Don't try to map when dealing with IResolvable

skorfmann:
Fixes: https://github.com/hashicorp/terraform-cdk/issues/835

### DIFF
--- a/packages/cdktf/lib/runtime.ts
+++ b/packages/cdktf/lib/runtime.ts
@@ -48,6 +48,9 @@ export function hashMapper(elementMapper: Mapper): Mapper {
  * True unless it's undefined
  */
 export function canInspect(x: any) {
+  // Don't inspect IResolvable
+  if (x && 'resolve' in x) return false;
+
   // Note: using weak equality on purpose, we also want to catch undefined
   return x != null;
 }
--- a/packages/cdktf/lib/runtime.ts
+++ b/packages/cdktf/lib/runtime.ts
@@ -49,7 +49,7 @@ export function hashMapper(elementMapper: Mapper): Mapper {
  */
 export function canInspect(x: any) {
   // Don't inspect IResolvable
-  if (x && 'resolve' in x) return false;
+  if (x && "resolve" in x) return false;
 
   // Note: using weak equality on purpose, we also want to catch undefined
   return x != null;


jsteinich:
Looks like the AWS CDK doesn't do this and instead resolves tokens before rendering to CloudFormation, but I'm not sure if that will cause any problems.

The exact check added feels more like a JS check than a TS check.

skorfmann:
> Looks like the AWS CDK doesn't do this and instead resolves tokens before rendering to CloudFormation, but I'm not sure if that will cause any problems.

Yes, I looked [there](https://github.com/aws/aws-cdk/blob/39aacc81e4a97f6de504de52d276a8d082059a0c/packages/%40aws-cdk/core/lib/runtime.ts#L206-L209) as well. I believe they resolve tokens beforehand. 

At least in my local test case, it was working as expected. Gonna add an explicit test here as well.

> The exact check added feels more like a JS check than a TS check.

Haven't found a way to check for interfaces at runtime. But yeah, the better way would probably be to check for [Token.isUnresolved](https://github.com/hashicorp/terraform-cdk/blob/main/packages/cdktf/lib/tokens/token.ts#L36)

jsteinich:
I can imagine cases where resolving things before this point would be necessary, so that might be the better route to go. I'm not thinking of anything that would break by checking for unresolved tokens, but probably want some tests to be sure.

packages/cdktf/lib/runtime.ts
```diff
@@ -48,6 +48,9 @@ export function hashMapper(elementMapper: Mapper): Mapper {
  * True unless it's undefined
  */
 export function canInspect(x: any) {
+  // Don't inspect IResolvable
+  if (x && "resolve" in x) return false;
```
DanielMSchmidt, line 52: We should use `Tokenization.isResolvable(x)` here

skorfmann: I believe I ran into issues there, but will check again.

jsteinich: I still wonder if we should be doing a multi-phase resolve like I believe the AWS CDK is doing.

skorfmann: @jsteinich me too - I haven't looked into this yet. Do you have more insights on reasons to do so?

jsteinich: I think it would only matter if resolving one token resulting in a different token, but I really haven't dug into it either. Really more of a hunch that as we lean harder into the token system, we'll run into issues since we aren't doing things exactly the same way.

skorfmann:
Closing in favour of https://github.com/hashicorp/terraform-cdk/pull/836

packages/cdktf/lib/runtime.ts
skorfmann: See https://github.com/hashicorp/terraform-cdk/issues/910

Issue #801, opened 2021-06-28
cdktf `watch` command


ansgarm:
The `watch` command is supposed to watch source files and to automatically do a `cdktf deploy (--auto-approve)` when they change to support quickly iterating on infrastructure (especially useful for serverless use-cases).

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Add a `watch` command to the `cdktf-cli` package. When executed, this command is
long-running and keeps watching the source directory for changes to files. When
a change is detected, it will run the `app` script of the `cdktf.json` and to
build and synthesize the application.  
After running the `synth` it will compare the `cdk.tf.json` output with a
previous version (probably using hashes kept
in-memory)
and determine whether changes need to be deployed. If there are any, it will
automatically deploy them (auto-approving), ideally also skipping the
`diff`/`plan` phase to save time.  
To make sure the user knows that the initial iteration of the `watch` command
deploys automatically, the `--auto-approve` flag is required on the `watch`
command until it supports a manual approval step before deploying. So, in the
first iteration only `cdktf watch --auto-approve` will work but `cdktf watch`
will fail with a short note about this behaviour.  
By using supplying `-refresh=false` to Terraform on subsequent deploys, we can
possibly further speed up things.

#### Determining the source directory

For a first iteration the command could watch any file in the same directory as
the `cdktf.json` file, except for the build / synth output of the CDKTF and
except for files specified in the `.gitignore` file.

If this might impose problems, the source directory could also be inferred for
the target language, however it might be useful to expose an option (on the CLI
command and/or in the `cdktf.json` config) as the code structure used by users
could differ quite substantially.

#### Handling subsequent changes

If new file updates are detected while a deploy is currently running a second
deploy should be queued so that changes can always be expected to get applied
(even if they take longer this way). In future iterations this could be made
smarter (e.g. just redoing the `synth` step if the deploy itself wasn't running
yet).

#### Handling errors

If an error happens, the `watch` should continue running but display a message
to the user that informs him that something failed. Possible errors include:

- resources can't be applied properly for some reason
- state is locked and stuck
- cloud credentials are expired

[...]

PR #843, opened 2021-07-23
chore(deps): use nodejs v14 version of aws jsii/superchain image

ansgarm:
This is mainly to test whether everything would still build and compile with this version.

We require Nodejs version 12 in our installation instructions but currently use the default jsii/superchain Docker image for builds which contains Nodejs v10. Unfortunately there is no Docker container with Nodejs v12 readily available for the jsii/superchain image.
The reason for this test is that the graphql-server, which is used in #801, needs Nodejs v12.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM docker.mirror.hashicorp.services/jsii/superchain
+FROM docker.mirror.hashicorp.services/jsii/superchain:node14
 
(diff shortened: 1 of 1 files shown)


ansgarm:
Hm, I just noticed that this might not work, we use a published version of our Docker container in our build and no version is published until we don't merge this into main 😅🤔

DanielMSchmidt (approved):

PR #845, opened 2021-07-26
fix(hcl2cdk): suffix variables and locals

DanielMSchmidt:
### Input

```hcl
variable "test" {
  type = string
}

locals {
  test = "${var.test} + 1"
}

output "test" {
  value = "${local.test}"
}
```

### Output Before PR

```ts
import * as cdktf from "cdktf";
const test = new cdktf.TerraformVariable(this, "test", {});
const test = `\${${test.value}} + 1`;
new cdktf.TerraformOutput(this, "test", {
  value: test,
});
```

### Output After PR

```ts
import * as cdktf from "cdktf";
const testVar = new cdktf.TerraformVariable(this, "test", {});
const testLocal = `\${${testVar.value}} + 1`;
new cdktf.TerraformOutput(this, "test", {
  value: testLocal,
});
```

### DIFF
--- a/packages/@cdktf/hcl2cdk/lib/expressions.ts
+++ b/packages/@cdktf/hcl2cdk/lib/expressions.ts
@@ -184,9 +184,7 @@ export function referenceToVariableName(ref: Reference): string {
 
 export function variableName(resource: string, name: string): string {
   const proposedName = camelCase(
-    ["var", "local", "module"].includes(resource)
-      ? name
-      : [resource, name].join("_")
+    resource === "module" ? name : [resource, name].join("_")
   );
 
   if (!Number.isNaN(parseInt(proposedName[0], 10))) {
--- a/packages/@cdktf/hcl2cdk/lib/expressions.ts
+++ b/packages/@cdktf/hcl2cdk/lib/expressions.ts
@@ -1,6 +1,6 @@
 import * as t from "@babel/types";
 import { camelCase, pascalCase } from "./utils";
-import { TerraformResourceBlock } from "./types";
+import { TerraformResourceBlock, Scope } from "./types";
 import isValidDomain from "is-valid-domain";
 
 export type Reference = {
@@ -175,23 +175,46 @@ export function extractReferencesFromExpression(
   }, [] as Reference[]);
 }
 
-export function referenceToVariableName(ref: Reference): string {
+export function referenceToVariableName(scope: Scope, ref: Reference): string {
   const parts = ref.referencee.full.split(".");
   const resource = parts[0] === "data" ? `${parts[0]}.${parts[1]}` : parts[0];
   const name = parts[0] === "data" ? parts[2] : parts[1];
-  return variableName(resource, name);
+  return variableName(scope, resource, name);
 }
 
-export function variableName(resource: string, name: string): string {
-  const proposedName = camelCase(
-    resource === "module" ? name : [resource, name].join("_")
-  );
+function validVarName(name: string) {
+  if (!Number.isNaN(parseInt(name[0], 10))) {
+    return `d${name}`;
+  } else {
+    return name;
+  }
+}
 
-  if (!Number.isNaN(parseInt(proposedName[0], 10))) {
-    return `d${proposedName}`;
+export function variableName(
+  scope: Scope,
+  resource: string,
+  name: string
+): string {
+  // name collision, we need to prefix the name
+  if (scope.variables[name]) {
+    if (resource === scope.variables[name].resource) {
+      return scope.variables[name].variableName;
+    }
+
+    // we only cache one per name
+    return validVarName(camelCase([resource, name].join("_")));
   }
 
-  return proposedName;
+  const variableName = validVarName(
+    camelCase(
+      ["var", "local", "module"].includes(resource)
+        ? name
+        : [resource, name].join("_")
+    )
+  );
+
+  scope.variables[name] = { variableName, resource };
+  return variableName;
 }
 
 export function constructAst(type: string, isModuleImport: boolean) {
@@ -219,11 +242,11 @@ export function constructAst(type: string, isModuleImport: boolean) {
   return t.identifier(pascalCase(type));
 }
 
-export function referenceToAst(ref: Reference) {
+export function referenceToAst(scope: Scope, ref: Reference) {
   const [resource, _name, ...selector] = ref.referencee.full.split(".");
 
   const variableReference = t.identifier(
-    camelCase(referenceToVariableName(ref))
+    camelCase(referenceToVariableName(scope, ref))
   );
 
   if (resource === "data") {
@@ -254,6 +277,7 @@ export function referenceToAst(ref: Reference) {
 }
 
 export function referencesToAst(
+  scope: Scope,
   input: string,
   refs: Reference[],
   scopedIds: readonly string[] = [] // dynamics introduce new scoped variables that are not the globally accessible ids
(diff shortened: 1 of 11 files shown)


ansgarm (approved):

jsteinich:
What's the motivation for this change?

DanielMSchmidt:
@jsteinich When converting https://github.com/futurice/terraform-examples/tree/master/azure/azure_linux_docker_app_service I found that locals, variables, and outputs can have the same name. Previously we just used the name directly, this resulted in multiple variables with the same name. Now they are suffixed so no more collisions can happen

jsteinich:
Locals are certainly often better as local variables, but I wonder if they should be translated that way or instead stay as `TerraformLocal`?

jsteinich:
> Previously we just used the name directly, this resulted in multiple variables with the same name. Now they are suffixed so no more collisions can happen

They will still collide in construct id space.

Probably other (if not all) variables that will encounter the same issue.

DanielMSchmidt:
@jsteinich Ah I see. We never translated locals into `TerraformLocal` so that's not new / the change in this PR. This one is just around variable naming. Is there a good reason to use `TerraformLocal` over local constants?

DanielMSchmidt:
I'm sorry, I meant locals and variables sharing the same name. Since locals are just constants there is no construct id involved.

jsteinich:
> Is there a good reason to use `TerraformLocal` over local constants?

Probably depends what the generation ends up looking like. As long as a Terraform expression is just passed along, then probably not.



> I'm sorry, I meant locals and variables sharing the same name. Since locals are just constants there is no construct id involved.

The variable and the output in your example overlap.

jsteinich:
> or is there a nice and easy way around it?

You could do the same suffix that you are using for the variable name, though there's no guarantee that you won't then collide with a user named resource.

DanielMSchmidt:
We have this problem across all resources, I think using the same formula we use to get variable names for scopes makes sense so that they are all properly "namespaced" by their respective resource type

packages/@cdktf/hcl2cdk/lib/expressions.ts
```diff
@@ -183,12 +183,20 @@ export function referenceToVariableName(ref: Reference): string {
 }
 
 export function variableName(resource: string, name: string): string {
-  const proposedName = camelCase(
+  let proposedName = camelCase(
     ["var", "local", "module"].includes(resource)
```
jsteinich, line 187: Should outputs be include in this check and then have a special case below?

packages/@cdktf/hcl2cdk/test/__snapshots__/hcl2cdk.test.ts.snap
```diff
@@ -12,46 +12,58 @@ exports[`convert all module types configuration 1`] = `
 "import * as cdktf from \\"cdktf\\";
 import * as AppTerraformIoExampleCorpK8SClusterAzurerm from \\"./.gen/modules/app.terraform.io/example-corp/k8s-cluster/azurerm\\";
 import * as HashicorpConsulAws from \\"./.gen/modules/hashicorp/consul/aws\\";
-new cdktf.TerraformHclModule(this, \\"consul\\", {
+new cdktf.TerraformHclModule(this, \\"cdktfTerraformHclModuleConsul\\", {
```
jsteinich, line 15: Maybe need a case for modules so this ends up as `ConsulModule`?

packages/@cdktf/hcl2cdk/test/__snapshots__/hcl2cdk.test.ts.snap
```diff
@@ -12,46 +12,58 @@ exports[`convert all module types configuration 1`] = `
...
 
 exports[`convert arithmetics configuration 1`] = `
 "import * as cdktf from \\"cdktf\\";
-const admins = new cdktf.TerraformVariable(this, \\"admins\\", {});
-const members = new cdktf.TerraformVariable(this, \\"members\\", {});
-new cdktf.TerraformOutput(this, \\"arithmetics\\", {
-  value: \`\\\\\${\${members.value} + \${admins.value}}\`,
+const adminsVar = new cdktf.TerraformVariable(
+  this,
+  \\"cdktfTerraformVariableAdmins\\",
```
jsteinich, line 50: I would have expected this to be `adminsVar`

jsteinich (changes requested):

packages/@cdktf/hcl2cdk/lib/index.ts
```diff
@@ -49,24 +49,28 @@ Please include this information:
 ${JSON.stringify((err as z.ZodError).errors)}`);
   }
 
+  // Each key in the scope needs to be unique, therefore we save them in a set
+  const scope = new Set<string>();
```
ansgarm, line 53: maybe also name this `scopeIdentifiers`? Sounds a bit more clear than just scope

ansgarm (approved):

jsteinich:
How feasible would it be to only do the variable qualifying if conflicts are detected?

packages/@cdktf/hcl2cdk/lib/index.ts
DanielMSchmidt: Yeah good idea :)

DanielMSchmidt:
@jsteinich I think this is much better, thank you!

packages/@cdktf/hcl2cdk/lib/expressions.ts
```diff
@@ -175,25 +175,46 @@ export function extractReferencesFromExpression(
...
 }
 
-export function variableName(resource: string, name: string): string {
-  const proposedName = camelCase(
-    ["var", "local", "module"].includes(resource)
-      ? name
-      : [resource, name].join("_")
-  );
+function validVarName(name: string) {
+  if (!Number.isNaN(parseInt(name[0], 10))) {
```
jsteinich, line 186: I didn't think just a number was valid in hcl, but I could imagine invalid names coming up, i.e. reserved keywords.

packages/@cdktf/hcl2cdk/lib/expressions.ts
```diff
@@ -175,25 +175,46 @@ export function extractReferencesFromExpression(
...
+  // name collision, we need to prefix the name
+  if (scope.variables[name]) {
+    if (resource === scope.variables[name].resource) {
+      return scope.variables[name].variableName;
+    }
 
-  if (!Number.isNaN(parseInt(proposedName[0], 10))) {
-    return `d${proposedName}`;
+    // we only cache one per name
+    return validVarName(camelCase([resource, name].join("_")));
```
jsteinich, line 205: Technically this could still collide, but probably don't need to worry about that.

packages/@cdktf/hcl2cdk/lib/index.ts
```diff
@@ -99,20 +104,23 @@ ${JSON.stringify((err as z.ZodError).errors)}`);
   // We recursively inspect each resource value to find references to other values
   // We add these to a dependency graph so that the programming code has the right order
   function addGlobalEdges(
+    _scopeIdentifiers: Scope,
```
jsteinich, line 107: Maybe name these `_scope`

jsteinich (approved):

packages/@cdktf/hcl2cdk/lib/expressions.ts
DanielMSchmidt: I am checking for reserved words in JS now. Rosetta will handle / check for reserved words in other languages afaik.

PR #846, opened 2021-07-26
fix(hcl2cdk): only run get conditionally

DanielMSchmidt:
Should fix running get if there are no modules and providers added to the cdktf.json

### DIFF
--- a/packages/@cdktf/hcl2cdk/lib/index.ts
+++ b/packages/@cdktf/hcl2cdk/lib/index.ts
@@ -332,7 +332,7 @@ export async function convertProject(
 
   return {
     code: prettier.format(outputMainFile, { parser: "babel" }),
-    cdktfJson: prettier.format(JSON.stringify(cdktfJson), { parser: "json" }),
+    cdktfJson,
     stats,
   };
 }
--- a/packages/@cdktf/hcl2cdk/lib/index.ts
+++ b/packages/@cdktf/hcl2cdk/lib/index.ts
@@ -301,10 +301,14 @@ export function getTerraformConfigFromDir(importPath: string) {
   return fileContents.join("\n");
 }
 
+type CdktfJson = Record<string, unknown> & {
+  terraformProviders: any[];
+  terraformModules: any[];
+};
 export async function convertProject(
   combinedHcl: string,
   inputMainFile: string,
-  inputCdktfJson: Record<string, unknown>,
+  inputCdktfJson: CdktfJson,
   { language }: ConvertOptions
 ) {
   if (language !== "typescript") {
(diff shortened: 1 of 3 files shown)


packages/cdktf-cli/bin/cmds/helper/init.ts
```diff
@@ -145,11 +145,17 @@ This means that your Terraform state file will be stored locally on disk in a fi
           "utf8"
         );
 
+        const { terraformModules, terraformProviders } = JSON.stringify(
+          cdktfJson
+        ) as any;
```
jsteinich, line 150: Why not just check the object directly?

DanielMSchmidt: you mean `config.terraformModules` etc?

jsteinich: I mean why convert `cdktfJson` to a string when it is already an object?

packages/cdktf-cli/bin/cmds/helper/init.ts
```diff
@@ -141,15 +141,22 @@ This means that your Terraform state file will be stored locally on disk in a fi
...
+        }
+
         telemetryData.conversionStats = stats;
       } catch (err) {
         throw Errors.Internal("init", err, { fromTerraformProject: true });
       }
-      execSync("npm run get", { cwd: destination });
     } else {
       console.error(
         `The --from-terraform-project flag is only support with the typescript template. The command will continue and ignore the flag.`
```
ansgarm, line 162: not new but: *supported

packages/cdktf-cli/bin/cmds/helper/init.ts
```diff
@@ -141,15 +141,22 @@ This means that your Terraform state file will be stored locally on disk in a fi
         fs.writeFileSync(path.resolve(destination, "main.ts"), code, "utf8");
         fs.writeFileSync(
           path.resolve(destination, "cdktf.json"),
-          cdktfJson,
+          JSON.stringify(cdktfJson, null, 2),
           "utf8"
         );
 
+        const { terraformModules, terraformProviders } = cdktfJson as {
+          terraformModules: string[];
```
ansgarm, line 149: Instead of using `as` you could also add proper types to the result from `convertProject` as this is not a string anymore now.

packages/@cdktf/hcl2cdk/lib/index.ts
```diff
@@ -301,10 +301,14 @@ export function getTerraformConfigFromDir(importPath: string) {
   return fileContents.join("\n");
 }
 
+type CdktfJson = Record<string, unknown> & {
+  terraformProviders: any[];
+  terraformModules: any[];
```
ansgarm, line 306: nit: they could also be undefined / not set for the input, but they won't be when they're outputted, right?

ansgarm (approved):

PR #847, opened 2021-07-26
fix(hcl2cdk): exclude non-registry modules from the cdktf.json

DanielMSchmidt:


jsteinich (approved):
I believe this will take care of the issue.
Longer term, may want to generate code for local modules.

DanielMSchmidt:
Yeah totally, that's the vision. I'd first try to get some more stability into convert, but once thats done having them as classes from resources would be really neat

PR #895, opened 2021-08-05
feat(cli): Improve telemetry

### LINKED ISSUE #826: Improve Telemetry

DanielMSchmidt:
This will help us make informed decision and gives us more context on how the CDK is used

Closes #826


packages/cdktf-cli/lib/checkpoint.ts
```diff
@@ -94,6 +102,20 @@ export async function ReportRequest(reportParams: ReportParams): Promise<void> {
...
+  const ci: string | false = ciDetect();
+  if (!reportParams.userId && !ci) {
+    reportParams.userId = machineIdSync();
+  }
+
+  if (ci) {
+    reportParams.ci = ci;
+  }
+
+  const gitRemote = await gitRemoteOriginUrl();
```
skorfmann, line 114: This will raise an error when it's not a git repo - see https://github.com/sindresorhus/git-remote-origin-url/blob/8573705e8f0f482708f0a63413cecfdde4e31d19/index.js#L12

packages/cdktf-cli/lib/checkpoint.ts
```diff
@@ -94,6 +102,20 @@ export async function ReportRequest(reportParams: ReportParams): Promise<void> {
...
+  const hash = crypto.createHash("md5");
+  hash.update(gitRemote);
+  reportParams.gitRepoId = hash.digest("hex").slice(0, 32).toUpperCase();
```
skorfmann, line 117: Why is it truncated?

packages/cdktf-cli/lib/checkpoint.ts
skorfmann: Also, isn't this something like a project fingerprint we could create once as a uuid on project init and add it to `cdktf.json`?

packages/cdktf-cli/lib/checkpoint.ts
DanielMSchmidt: I was unsure on how long it will be and I thought 32 should be enough to not run into collisions.

packages/cdktf-cli/lib/checkpoint.ts
DanielMSchmidt: That would be a different solution, but yeah we can add a project fingerprint to the cdktf.json if this seems like a better solution to you

skorfmann: Not sure if it's better. It's harder to retrofit the uuid approach for existing projects and we'd have to find a way to pass the value through to this point. The upside is, that we wouldn't have to re-run this for each and every command and that it's not depending on git.

packages/cdktf-cli/lib/checkpoint.ts
skorfmann: Asking stackoverflow yields

> A 128-bit MD5 hash is represented as a sequence of 32 hexadecimal digits

https://stackoverflow.com/questions/3394503/maximum-length-for-md5-input-output

DanielMSchmidt: huh, yeah seems like I don't need the slice then

packages/cdktf-cli/lib/checkpoint.ts
DanielMSchmidt: Downside is that if you copy and paste instead of using init we'll get wrong data. I'll add a try catch and we'll see if this approach works well enough

packages/cdktf-cli/bin/cmds/helper/init.ts
```diff
@@ -15,6 +15,19 @@ import { Errors } from "../../../lib/errors";
...
+
+function getProjectId(
+  projectName: string,
+  projectDescription: string,
+  projectPath = process.cwd()
+) {
+  const hash = crypto.createHash("md5");
+  hash.update(projectName);
+  hash.update(projectDescription);
+  hash.update(projectPath);
```
skorfmann, line 28: What's the reason for hashing these values, rather than generating a uuid?

DanielMSchmidt: There is none, could do either way

skorfmann: I think a uuid would make more sense and is easier in this case

danieldreier:
I'm working on an internal RFC and external messaging to explain what data we're collecting and what our privacy controls are, so let's hold off on merging this until that's compelte

danieldreier:
As prompted by the suggestion @DanielMSchmidt made on the internal RFC, I think we should stop sending `command` data in to telemetry, because in principal it could contain environment variables or other sensitive information. We already separately track the subcommand being invoked separately from the full command string, and don't use that `command` telemetry data. @DanielMSchmidt do you have a preference between doing that in this PR vs as a separate issue?

packages/cdktf-cli/lib/checkpoint.ts
```diff
@@ -94,7 +110,19 @@ export async function ReportRequest(reportParams: ReportParams): Promise<void> {
     reportParams.os = os.platform();
   }
 
+  const ci: string | false = ciDetect();
+  if (!reportParams.userId && !ci) {
+    reportParams.userId = machineIdSync();
```
danieldreier, line 115: In RFC review @eastebry suggested that we should add a static salt to this. I propose we use the string "cdktf-static-salt" or similar to be explicit that this is intentionally a static salt. The idea is that in case our checkpoint data were ever compromised, the hashes couldn't readily be matched to other datasets.

danieldreier (approved):
The internal RFC has been approved and this is good to go.

docs/working-with-cdk-for-terraform/cdktf-json.md
```diff
@@ -25,6 +25,7 @@ export interface Config {
   readonly language?: Language; // Target language for building provider or module bindings. Currently supported: `typescript`, `python`, `java`, `csharp`, and `go`
   readonly output: string; // Default: 'cdktf.out'. Where the synthesized JSON should go. Also will be the working directory for Terraform operations
   readonly codeMakerOutput: string; // Default: '.gen'. Path where generated provider bindings will be rendered to.
+  readonly projectId: string; // Default: generated UUID. Unique identifier for the project used to differenciate projects
```
ansgarm, line 28:
```suggestion
  readonly projectId: string; // Default: generated UUID. Unique identifier for the project used to differentiate projects
```

PR #903, opened 2021-08-10
feat(lib): expose terraform functions

### LINKED ISSUE #220: Built-in Function Compatibility
### Community Note

### Description

Currently, as far as I'm aware, the terraform built-in functions are not compatible with cdktf. 

Examples below:
https://www.terraform.io/docs/configuration/functions.html
https://www.terraform.io/docs/configuration/functions/cidrsubnet.html

DanielMSchmidt:
Closes #220

When this is done we should create an issue to improve convert to use the functions


jsteinich:
How does jsii handle this approach?

skorfmann:
FYI: The type export in jsii seems to be a bit picky - see https://github.com/aws/jsii/issues/1818

packages/cdktf/lib/terraform-remote-state.ts
```diff
@@ -40,8 +41,10 @@ export abstract class TerraformRemoteState extends TerraformElement {
     return Token.asAny(this.interpolationForAttribute(output));
   }
 
-  private interpolationForAttribute(terraformAttribute: string): any {
-    return `\${data.terraform_remote_state.${this.friendlyUniqueId}.outputs.${terraformAttribute}}`;
+  private interpolationForAttribute(terraformAttribute: string): Expression {
```
jsteinich, line 44: Should these type of functions return `Expression` or `any`? This appears to be the only one that is `Expression`; however, using `any` seems likely to cause bugs.

test/python/terraform-functions/__snapshots__/test.ts.snap
```diff
@@ -0,0 +1,48 @@
...
+          \\"path\\": \\"python-simple/CustomVpc\\",
+          \\"uniqueId\\": \\"CustomVpc\\"
+        }
+      }
+    }
+  },
+  \\"output\\": {
+    \\"computed\\": {
+      \\"value\\": {
+        \\"jsii_byref\\": \\"Object@10006\\"
```
jsteinich, line 43: This doesn't seem to have resolved correctly.

jsteinich (changes requested):

test/python/terraform-functions/__snapshots__/test.ts.snap
DanielMSchmidt: Yeah, I think it was due to a wrong return type of the functions. The current version has a different problem, so one step at a time I guess :D

packages/cdktf/lib/terraform-remote-state.ts
skorfmann: I think `Expression` would make sense - or do you see other cases?

packages/cdktf/lib/tokens/private/resolve.ts
```diff
@@ -109,6 +109,13 @@ export function resolve(obj: any, options: IResolveOptions): any {
   // string - potentially replace all stringified Tokens
   //
   if (typeof obj === "string") {
+    // If this is a "list element" Token, it should never occur by itself in string context
+    if (TokenString.forListToken(obj).test()) {
+      throw new Error(
```
jsteinich, line 114: I almost wonder if we shouldn't just make it to work instead of erroring. If the user tries to use something other than index `0` it will error out anyways, so this could maybe be a nice shortcut.

packages/cdktf/lib/tokens/token.ts
```diff
@@ -93,6 +93,27 @@ export class Token {
  * Less oft-needed functions to manipulate Tokens
  */
 export class Tokenization {
+  /**
+   * Reverse any value into a Resolvable, if possible
```
jsteinich, line 97:
```suggestion
   * Reverse any value into Resolvables, if possible
```

packages/cdktf/lib/tokens/token.ts
```diff
@@ -93,6 +93,27 @@ export class Token {
...
+   */
+  public static reverse(x: any): IResolvable[] {
+    if (Tokenization.isResolvable(x)) {
+      return [x];
+    }
+    if (typeof x === "string") {
+      return Tokenization.reverseString(x).tokens;
```
jsteinich, line 104: This won't include intrinsic values which may also be resolvable.

packages/cdktf/lib/tokens/token.ts
```diff
@@ -93,6 +93,27 @@ export class Token {
...
+    }
+    if (Array.isArray(x)) {
+      const reversedList = Tokenization.reverseList(x);
+      return reversedList ? [reversedList] : [];
+    }
+    if (typeof x === "number") {
+      const reversedNumber = Tokenization.reverseNumber(x);
+      return reversedNumber ? [reversedNumber] : [];
+    }
+    return [];
```
jsteinich, line 114: Should generic object values be checked?
Seems similar to resolve flow.

packages/cdktf/test/functions.test.ts
```diff
@@ -0,0 +1,357 @@
...
+test("spreaded mixed values", () => {
+  const app = Testing.app();
+  const stack = new TerraformStack(app, "test");
+
+  const variable = new TerraformVariable(stack, "test-var", {
+    type: "number",
+  });
+
+  new TerraformOutput(stack, "test-output", {
+    value: Fn.max([10, variable.value, 200]),
```
jsteinich, line 72: Passing an array is a bit clunky.

test/python/terraform-functions/__snapshots__/test.ts.snap
```diff
@@ -0,0 +1,46 @@
...
+        \\"metadata\\": {
+          \\"path\\": \\"python-simple/CustomVpc\\",
+          \\"uniqueId\\": \\"CustomVpc\\"
+        }
+      }
+    }
+  },
+  \\"output\\": {
+    \\"computed\\": {
+      \\"value\\": \\"\${cidrsubnet(element(merge(\\\\\\"\${module.CustomVpc.public_subnets_cidr_blocks}\\\\\\", \\\\\\"\${module.CustomVpc.public_subnets_ipv6_cidr_blocks}\\\\\\"), 3), 4, 2)}\\"
```
jsteinich, line 42: I'm not sure if the module references should be using `"${..}"` vs just direct access.

packages/cdktf/lib/terraform-functions.ts
```diff
@@ -0,0 +1,1221 @@
...
+type TFValueValidator = (value: any) => TFValue;
+
+type ExecutableTfFunction = (...args: any[]) => IResolvable;
+
+function anyValue(value: any): any {
+  return value;
+}
+
+function mapValue(value: any): any {
+  return value;
```
jsteinich, line 18: Could do some validation here

packages/cdktf/lib/terraform-functions.ts
```diff
@@ -0,0 +1,1221 @@
...
+import { TokenMap } from "./tokens/private/token-map";
+
+// We use branding here to ensure we internally only handle validated values
+// this allows us to catch usage errors before terraform does in some cases
+type TFValue = any & { __type: "tfvalue" };
+type TFValueValidator = (value: any) => TFValue;
+
+type ExecutableTfFunction = (...args: any[]) => IResolvable;
+
+function anyValue(value: any): any {
```
jsteinich, line 13: Maybe somehow group these validators to make it more clear that they are validators and not just arbitrary functions

packages/cdktf/lib/terraform-functions.ts
```diff
@@ -0,0 +1,1221 @@
...
+      }
+      args.forEach((arg, i) => argValidators[i](arg));
+      return call(name, args);
+    } else {
+      return call(name, argValidators(args));
+    }
+  };
+}
+
+export class Fn {
```
jsteinich, line 95: Might be nice to split this into multiple files.

docs/working-with-cdk-for-terraform/terraform-functions.md
```diff
@@ -0,0 +1,27 @@
...
+
+Terraform provides a set of functions that can be used through the CDK for Terraform.
+
+```ts
+import { Fn } from "cdktf";
+
+new vpc() = new VPC(this, "vpc", {});
+new LoadBalancer(this, "lb", {
+  name: "main-lb",
+  subnet: Fn.Network.cidrsubnet(
```
jsteinich, line 11: Looks like namespacing got dropped. Would be nice to have though.

packages/cdktf/lib/terraform-functions.ts
```diff
@@ -0,0 +1,1221 @@
...
+  /**
+   * {@link https://www.terraform.io/docs/language/functions/anytrue.html anytrue} returns true if all elements in a given collection are true or "true"
+   * @param {Array} values
+   */
+  public static alltrue(values: any[]) {
+    return terraformFunction("alltrue", [listOf(anyValue)])(values);
```
jsteinich, line 101: Return value not currently usable.

packages/cdktf/lib/terraform-functions.ts
```diff
@@ -0,0 +1,1221 @@
...
+  }
+
+  /**
+   * {@link https://www.terraform.io/docs/language/functions/anytrue.html anytrue} returns true if any element in a given collection is true or "true"
+   * @param {Array} value
+   */
+  public static anytrue(value: any[]) {
+    return terraformFunction("anytrue", [listOf(anyValue)])(value);
```
jsteinich, line 109: Return value not currently usable.

packages/cdktf/lib/terraform-functions.ts
```diff
@@ -0,0 +1,1221 @@
...
+  }
+
+  /**
+   * {@link https://www.terraform.io/docs/language/functions/chunklist.html chunklist} splits a single list into fixed-size chunks, returning a list of lists.
+   * @param {Array} value
+   * @param {number} chunkSize
+   */
+  public static chunklist(value: any[], chunkSize: number | IResolvable) {
+    return asList(
```
jsteinich, line 118: This will only work for lists of string currently.

packages/cdktf/lib/terraform-functions.ts
```diff
@@ -0,0 +1,1221 @@
...
+      )
+    );
+  }
+
+  /**
+   * {@link https://www.terraform.io/docs/language/functions/coalesce.html coalesce} takes any number of arguments and returns the first one that isn't null or an empty string.
+   * @param {Array} value - Arguments are passed in an array
+   */
+  public static coalesce(value: any[]) {
+    return terraformFunction("coalesce", listOf(anyValue))(...value);
```
jsteinich, line 131: Too bad we can't use generics to get more informative return type here.

packages/cdktf/lib/terraform-functions.ts
```diff
@@ -0,0 +1,1221 @@
...
+  }
+
+  /**
+   * {@link https://www.terraform.io/docs/language/functions/coalescelist.html coalescelist} takes any number of list arguments and returns the first one that isn't empty.
+   * @param Array} value - Arguments are passed in an array
+   */
+  public static coalescelist(value: any[]) {
```
jsteinich, line 138: This takes a list of lists

packages/cdktf/lib/terraform-functions.ts
```diff
@@ -0,0 +1,1221 @@
...
+    return terraformFunction("coalescelist", listOf(anyValue))(...value);
+  }
+
+  /**
+   * {@link https://www.terraform.io/docs/language/functions/compact.html compact} takes a list of strings and returns a new list with any empty string elements removed.
+   * @param {Array} value
+   */
+  public static compact(value: any[]) {
```
jsteinich, line 146: The documentation implies that this only accepts lists of strings.

packages/cdktf/lib/terraform-functions.ts
```diff
@@ -0,0 +1,1221 @@
...
+    return asList(terraformFunction("compact", [listOf(anyValue)])(value));
+  }
+
+  /**
+   * {@link https://www.terraform.io/docs/language/functions/concat.html concat} takes two or more lists and combines them into a single list.
+   * @param {Array} value
+   */
+  public static concat(value: any[]) {
```
jsteinich, line 154: Input is list of lists.

packages/cdktf/lib/terraform-functions.ts
```diff
@@ -0,0 +1,1221 @@
...
+    return asList(terraformFunction("concat", listOf(anyValue))(...value));
+  }
+
+  /**
+   * {@link https://www.terraform.io/docs/language/functions/contains.html contains} determines whether a given list or set contains a given single value as one of its elements.
+   * @param {Array} list
+   * @param {any} value
+   */
+  public static contains(list: any[] | string, value: any) {
+    return terraformFunction("contains", [listOf(anyValue), anyValue])(
```
jsteinich, line 164: Return value not currently usable.

packages/cdktf/lib/terraform-functions.ts
```diff
@@ -0,0 +1,1221 @@
...
+  public static concat(value: any[]) {
+    return asList(terraformFunction("concat", listOf(anyValue))(...value));
+  }
+
+  /**
+   * {@link https://www.terraform.io/docs/language/functions/contains.html contains} determines whether a given list or set contains a given single value as one of its elements.
+   * @param {Array} list
+   * @param {any} value
+   */
+  public static contains(list: any[] | string, value: any) {
```
jsteinich, line 163: I didn't see `contains` listed under string functions; does it accept a string?
We don't have the set type currently, but would be nice to make a note of it for the future.

packages/cdktf/lib/terraform-functions.ts
```diff
@@ -0,0 +1,1221 @@
...
+      list,
+      value
+    );
+  }
+
+  /**
+   * {@link https://www.terraform.io/docs/language/functions/distinct.html distinct} takes a list and returns a new list with any duplicate elements removed.
+   * @param {Array} list
+   */
+  public static distinct(list: any[] | string | IResolvable) {
```
jsteinich, line 174: Does this work with strings?

jsteinich (commented):
I didn't make it through every single function to make sure they are correct. I can keep going on them at a later date, but I think it's important to match how Tokens are used elsewhere. That could be not allowing them as arguments (since they aren't currently allowed in resource) or using them in the same manner as planned (never for strings/numbers and generally always an option for other types). It's important to maintain that consistency.

packages/cdktf/lib/tfExpression.ts
```diff
@@ -0,0 +1,111 @@
...
+    const expr = `${this.name}(${serializedArgs})`;
+
+    return this.isInnerTerraformExpression ? expr : `\${${expr}}`;
+  }
+}
+export function call(name: string, args: Expression[]) {
+  return new FunctionCall(name, args) as IResolvable;
+}
+
+export type Expression =
```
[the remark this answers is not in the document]
DanielMSchmidt: Yeah, we can add what we need down the road :)

packages/cdktf/lib/tfExpression.ts
```diff
@@ -0,0 +1,111 @@
...
+export function ref(identifier: string) {
+  return new Reference(identifier) as IResolvable;
+}
+
+function markAsInner(arg: any) {
+  if (arg instanceof TFExpression) {
+    arg.isInnerTerraformExpression = true;
+  }
+
+  // reverese tokens here and set inner tf expression flag
```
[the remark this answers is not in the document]
DanielMSchmidt: Yeah good point, not sure why I had it there in the first place 🙈

packages/cdktf/lib/tfExpression.ts
```diff
@@ -0,0 +1,111 @@
...
+  Tokenization.reverse(arg).map((resolvable) => {
+    if (resolvable instanceof TFExpression) {
+      resolvable.isInnerTerraformExpression = true;
+    }
+  });
+
+  if (Array.isArray(arg)) {
```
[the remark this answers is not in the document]
DanielMSchmidt: The token reversing already returns an array with all args, this is for normal arrays with no token content but plain references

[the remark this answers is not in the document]
DanielMSchmidt: And I'll add a test for generic objects, I think I haven't tested this very well

packages/cdktf/lib/tfExpression.ts
```diff
@@ -0,0 +1,111 @@
...
+  constructor(private name: string, private args: Expression[]) {
+    super({ name, args });
+
+    this.args.forEach(markAsInner);
+  }
+
+  public resolve(context: IResolveContext): string {
+    const serializedArgs = this.args
+      .map((arg) => {
+        const resolvedArg = context.resolve(arg);
```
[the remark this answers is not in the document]
DanielMSchmidt: Yeah I agree, I'm not sure how to improve it though. There are some differences in these cases here due to the handling of escaping, etc. 🤔

packages/cdktf/lib/tokens/private/resolve.ts
DanielMSchmidt: I wonder if we could mask confusing errors here if we don't error 🤔 Wdyt @ansgarm @skorfmann ?

packages/cdktf/lib/tokens/token.ts
DanielMSchmidt: No I think this is not necessary, we traverse the object elsewhere

packages/cdktf/test/functions.test.ts
DanielMSchmidt: Yeah, I'm also not a huge fan, but JSII seemed to have problems with spread parameters 🤔

packages/cdktf/lib/terraform-functions.ts
DanielMSchmidt: I tried, but JSII exploded then 🙈  Had them all namespaced in their group but JSII ran into problems

packages/cdktf/lib/tokens/private/resolve.ts
skorfmann: hm, how would the native error look like?

ansgarm: I'm currently lacking the context to even understand the problem here 😅

docs/working-with-cdk-for-terraform/terraform-functions.md
skorfmann: It might work when following this https://github.com/aws/jsii/blob/ff03310e1540ddf5364ab20390e854be9b57bc17/gh-pages/content/specification/2-type-system.md#declaration

packages/cdktf/lib/terraform-functions.ts
DanielMSchmidt: Yeah, it's really a bummer. I typed them as any now with a string representation 👍

packages/cdktf/lib/terraform-functions.ts
DanielMSchmidt: I typed them all now as boolean or any similar to how we do it for resource attributes

packages/cdktf/lib/tokens/private/resolve.ts
jsteinich: The idea would be to resolve the token rather than error. The risk I see is that if a user was iterating through an array, they might think it was working, but in reality it wouldn't be since the token array always has length 1 and index 0 would function (if we made it work). Should probably go back to the issue that I believe inspired the error to help make that call.

packages/cdktf/test/functions.test.ts
jsteinich: It looks as though jsii is intended to support variadic arguments, but there may be some issues.

skorfmann (dismissed):

packages/cdktf/lib/terraform-local.ts
```diff
@@ -16,27 +17,27 @@ export class TerraformLocal extends TerraformElement {
...
   }
 
   public get asList(): string[] {
-    return Token.asList(this.interpolation());
+    return Token.asList(ref(this.interpolation()));
   }
 
   public get asBoolean(): boolean {
-    return Token.asString(this.interpolation()) as any as boolean;
+    return Token.asString(ref(this.interpolation())) as any as boolean;
```
jsteinich, line 36: Not related to this PR, but this cast has never worked correctly and still won't after this change.

docs/working-with-cdk-for-terraform/terraform-functions.md
```diff
@@ -0,0 +1,27 @@
...
+    Fn.Collection.element(vpc.listOfSubnets, 0),
+    4,
+    2
+  ),
+});
+```
+
+To explore the functionality either check out the [Terraform Documentation](https://www.terraform.io/docs/language/functions/index.html) or use your editor autocompletion on the `Fn` object.
```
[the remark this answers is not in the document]
jsteinich: No, I mean using the Token system to get a token of a different type. I.E. using the `coalesce` function is a list of numbers will return `IResolvable` rather than `number`.

packages/cdktf/lib/terraform-functions.ts
```diff
@@ -0,0 +1,1250 @@
...
+// Tokenization
+function asString(value: IResolvable) {
+  return TokenMap.instance().registerString(value);
+}
+
+function asNumber(value: IResolvable) {
+  return TokenMap.instance().registerNumber(value);
+}
+
+function asList(value: IResolvable) {
```
jsteinich, line 87: Since this always returns `string[]` regardless of the type, using it for functions that return other types of lists will give back a not very useable array.

packages/cdktf/lib/terraform-functions.ts
```diff
@@ -0,0 +1,1250 @@
...
+  public static anytrue(value: any[]) {
+    return asBoolean(terraformFunction("anytrue", [listOf(anyValue)])(value));
+  }
+
+  /**
+   * {@link https://www.terraform.io/docs/language/functions/chunklist.html chunklist} splits a single list into fixed-size chunks, returning a list of lists.
+   * @param {Array} value
+   * @param {number} chunkSize
+   */
+  public static chunklist(value: any[], chunkSize: number | IResolvable) {
```
jsteinich, line 142: Since number tokens are represented by numbers, is `IResolvable` needed?

packages/cdktf/lib/terraform-functions.ts
```diff
@@ -0,0 +1,1250 @@
...
+    return asList(terraformFunction("distinct", [listOf(anyValue)])(list));
+  }
+
+  /**
+   * {@link https://www.terraform.io/docs/language/functions/element.html element} retrieves a single element from a list.
+   * @param {Array} list
+   * @param {number} index
+   */
+  public static element(
+    list: any[] | string | IResolvable,
```
jsteinich, line 210: Does this work for strings?

I noticed that other collection functions don't take `IResolvable`; shouldn't any array parameter support using `IResolvable`?

packages/cdktf/lib/terraform-functions.ts
```diff
@@ -0,0 +1,1250 @@
...
+        index
+      )
+    );
+  }
+
+  /**
+   * {@link https://www.terraform.io/docs/language/functions/flatten.html flatten} takes a list and replaces any elements that are lists with a flattened sequence of the list contents.
+   * @param {Array} list
+   */
+  public static flatten(list: any[] | string) {
```
jsteinich, line 225: Can this take `string`?

packages/cdktf/lib/terraform-functions.ts
```diff
@@ -0,0 +1,1250 @@
...
+    return asList(terraformFunction("flatten", [listOf(anyValue)])(list));
+  }
+
+  /**
+   * {@link https://www.terraform.io/docs/language/functions/index.html index} finds the element index for a given value in a list.
+   * @param {Array} list
+   * @param {any} value
+   */
+  public static index(list: any[] | string, value: any) {
```
jsteinich, line 234: Can this take `string`?

packages/cdktf/lib/terraform-functions.ts
```diff
@@ -0,0 +1,1250 @@
...
+    return asNumber(
+      terraformFunction("index", [listOf(anyValue), anyValue])(list, value)
+    );
+  }
+
+  /**
+   * {@link https://www.terraform.io/docs/language/functions/keys.html keys} takes a map and returns a list containing the keys from that map.
+   * @param {Object} map
+   */
+  public static keys(map: any) {
```
jsteinich, line 244: Should this take a map with string keys instead?

jsteinich (commented):
I think we need to make a very detailed pass on the function parameter and return types. We want them to be as narrow as possible, but still need full functionality.

packages/cdktf/lib/terraform-local.ts
DanielMSchmidt: I changed the boolean interfaces now to take boolean and IResolvable, so I will return the ref directly 👍

packages/cdktf/lib/terraform-functions.ts
DanielMSchmidt: I mean this is the current way how we represent lists, isn't it?

jsteinich: It is, so could definitely be tackled as a followup.

packages/cdktf-cli/test/get/generator/__snapshots__/complex-computed-types.test.ts.snap
```diff
@@ -37,7 +37,7 @@ export interface AcmCertificateConfig extends cdktf.TerraformMetaArguments {
   /**
   * Docs at Terraform Registry: {@link https://www.terraform.io/docs/providers/aws/r/acm_certificate.html#tags AcmCertificate#tags}
   */
-  readonly tags?: { [key: string]: string };
+  readonly tags?: { [key: string]: string } | cdktf.IResolvable;
```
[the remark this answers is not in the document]
DanielMSchmidt: Object :(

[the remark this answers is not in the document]
ansgarm: Hm, let's ignore this for now. I don't think it'll be a downgrade from the current implementation at least.

ansgarm (approved):

packages/cdktf/lib/tokens/private/resolve.ts
DanielMSchmidt: I think I'd go here with the earliest error possible so folks don't think they can traverse an array and can do so without an error and just the unintended HCL.json

packages/cdktf/lib/tfExpression.ts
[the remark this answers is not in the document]
DanielMSchmidt: Ah, got the context back, we just set the `isInnerTerraformExpression` function. This is needed to get rid of the extra `${}` wrapping, we just change the value, we don't do anything else with it

packages/cdktf/lib/terraform-functions.ts
```diff
@@ -0,0 +1,1248 @@
...
+function asNumber(value: IResolvable) {
+  return TokenMap.instance().registerNumber(value);
+}
+
+function asList(value: IResolvable) {
+  return TokenMap.instance().registerList(value);
+}
+
+function asBoolean(value: IResolvable) {
+  return asAny(value) as boolean;
```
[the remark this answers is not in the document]
DanielMSchmidt: Thank you, I addressed this now👍

PR #905, opened 2021-08-12
feat(cli): Add python-poetry CLI template

johnfraney:
Adds CLI template for Poetry, a popular alternative dependency and package manager for Python.

See: https://python-poetry.org/

I tested this locally by running the following in a sibling directory of `terraform-cdk` and it worked well!

```console
node ../terraform-cdk/packages/cdktf-cli/bin/cdktf.js init --template="python-poetry" --local --cdktf-version 0.5.0
```

### DIFF
--- a/docs/cli-commands.md
+++ b/docs/cli-commands.md
@@ -126,7 +126,8 @@ Options:
   --log-level                 Which log level should be written. Only supported via setting the env CDKTF_LOG_LEVEL
                                                                                                            [string]
   --template                  The template to be used to create a new project. Either URL to zip file or one of the
-                              built-in templates: ["csharp", "go", "java", "python", "python-pip", "typescript"]
+                              built-in templates: ["csharp", "go", "java", "python", "python-pip", "python-poetry",
+                              "typescript"]
                                                                                                            [string]
   --project-name              The name of the project.                                                     [string]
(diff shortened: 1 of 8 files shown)


ansgarm:
Thank you for the nice PR, @johnfraney!

We talked about this in the team as this would add a third option for Python projects. And we'd like to support Python Poetry via a [remote template](https://github.com/hashicorp/terraform-cdk/blob/main/docs/working-with-cdk-for-terraform/remote-templates.md) rather than built-in via the CLI. We'd then link to the remote template in our Readme and docs instead.

It would be the first remote template that we link to, actually.
Would you be open for hosting this on your Github account?

To change this PR into a remote template no changes to the template code itself are required. If you are open to this idea, I'm happy to assist (e.g. by send a PR to a new repo of yours, etc.) 👨‍💻

johnfraney:
> Would you be open for hosting this on your Github account?

@ansgarm Sure! I'll give that a shot. I'll push up the template here: https://github.com/johnfraney/cdktf-remote-template-python-poetry

ansgarm:
@johnfraney Thanks! Tried it locally, works great 👍
I just added a PR to link your repo in our docs #922

johnfraney:
@ansgarm That's great! I'll close this PR now that we've achieved this using the remote template

PR #911, opened 2021-08-17
fix(lib): Alternative Validation Approach

skorfmann:
While the `validation` hook seems to be a good approach to handle this, we should avoid using `toTerraform` to collect the required data. As an alternative this PR walks over all construct nodes and collects the data from the actual objects. At some point, we might wanna think about optimizing this a bit, so that we don't have to visit all nodes again and again - but for now I believe it's good enough.

For the error which gets thrown, we might want to add a handler which beautifies this a bit and gets rid of stacktraces, etc...


packages/cdktf/lib/terraform-stack.ts
```diff
@@ -2,8 +2,6 @@ import { Construct, IConstruct, ISynthesisSession, Node } from "constructs";
 import { resolve } from "./_tokens";
 import * as fs from "fs";
 import * as path from "path";
-import camelcase = require("camelcase"); // ES module interop
```
DanielMSchmidt, line 5: We can remove this from the bundledDependencies array in the package.json then

packages/cdktf/lib/validations/validate-provider-presence.ts
```diff
@@ -0,0 +1,50 @@
...
+      return !this.foundProviders.some(
+        (p) => p.terraformGeneratorMetadata?.providerName === name
+      );
+    });
+
+    if (missingProviders.length === 0) {
+      return [];
+    } else {
+      return [
+        `Found resources without a matching povider. Please make sure to add the following providers to your stack ${missingProviders.join(
```
DanielMSchmidt, line 44:
```suggestion
        `Found resources without a matching povider. Please make sure to add the following providers to your stack: ${missingProviders.join(
```

DanielMSchmidt (approved):
Two small things, but seems like a great solution :)

packages/cdktf/lib/terraform-stack.ts
skorfmann: Good point - done 👍

packages/cdktf/lib/validations/validate-provider-presence.ts
```diff
@@ -0,0 +1,50 @@
...
+      return !this.foundProviders.some(
+        (p) => p.terraformGeneratorMetadata?.providerName === name
+      );
+    });
+
+    if (missingProviders.length === 0) {
+      return [];
+    } else {
+      return [
+        `Found resources without a matching povider. Please make sure to add the following providers to your stack: ${missingProviders.join(
```
jsteinich, line 44: Perhaps in the future we could exclude providers that don't require any configuration.

DanielMSchmidt: I'm sure I ran into an issue where I forgot to initialized the docker provider and the program errored. The providers also help to set the right name in the required_providers (org is defaulted to hashicorp).

jsteinich: I think that is true, but I don't recall that being the case in hcl (maybe it changed in newer versions).

skorfmann: > Perhaps in the future we could exclude providers that don't require any configuration.

Do you know if that's documented somewhere?

jsteinich: Everything in the schema would be optional. Post Terraform 0.13, probably only works for providers in the `hashicorp` namespace.

Ygshvr:
I am using cdktf 0.7.0 with python and facing a similar issue with the local provider.
I am trying to write to a file like this:
```python3
File(
            self,
            f"{cluster_name}-kubeconfig",
            content=self.gke_auth.kubeconfig_raw_output,
            filename=cluster_name,
            depends_on=[self.clusters[cluster_name]] # I have created an array of clusters so it's waiting for that cluster to be ready.
        )
```
But I get this error:
```
Found resources without a matching povider. Please make sure to add the following providers to your stack: local
```
So I tried to instantiate the provider like this 
```python3
LocalProvider(self, "local")
```
and got this error:
```
failed to instantiate provider "registry.terraform.io/hashicorp/local" to obtain schema: unknown provider "registry.terraform.io/hashicorp/local"
```
Any suggestions/workarounds?

Ygshvr:
> @Ygshvr did you add the local provider to your `cdktf.json` file and run `cdktf get`?

Yes, I did.

jsteinich:
@Ygshvr Can you please post your `cdktf.json` file? Also, are you behind any proxies?

Ygshvr:
@jsteinich no, I am not behind a proxy. The code was downloaded successfully. I did check it by going into the imports directory. Here is the `cdktf.json`:
```json
{
  "language": "python",
  "app": "pipenv run python main.py",
  "terraformProviders": [
    "hashicorp/google@ ~> 4.0.0"
  ],
  "terraformModules": [
    "terraform-google-modules/kubernetes-engine/google//modules/auth"
  ],
  "codeMakerOutput": "imports",
  "projectId": "286db479-f6a5-4a90-88b4-b0f8903f245a"
}

```

jsteinich:
@Ygshvr you'll either need to add the local provider to your `cdktf.json` and generate or use a pre-built version of it and then instantiate as you described earlier.

Issue #705, opened 2021-05-14
~Unable to initialize cdktf project in pycharm with venv activated


maciak-g:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### cdktf & Language Versions

0.3.0

### Affected Resource(s)

project initialize

### Debug Output

https://gist.github.com/maciak-g/2435e8595ca23c33c374c0c774a857ec

### Expected Behavior

Project should be initialized

### Actual Behavior

Error message that pip cannot perform --user install.

### Steps to Reproduce

Install pycharm, create project with pycharm, initialize venv environment in the project.

### Important Factoids

I want to initialize project from the level of pycharm not custom initialization tool and install pip packages within initialized python virtual environment. The second case is I want to add this library to existing project where I have already prepared configuration for objects I will provision to cloud. I am puzzled why there are no alternative instructions to setup project manually or some more options to direct how packages should be installed.

### References

* #0000

jsteinich:
There may be a conflict between pip and your virtualenv. There are 2 separate templates, perhaps trying the `python` one would work better with your setup.

When it comes to manual setup, there's really only a couple steps.
1. Create a [cdktf.json file](https://github.com/hashicorp/terraform-cdk/blob/main/docs/working-with-cdk-for-terraform/cdktf-json.md)
2. Install the [cdktf package](https://pypi.org/project/cdktf/0.3.0/)
3. Optionally install [prebuilt providers](https://pypi.org/user/cdktf-team/)

maciak-g:
But could you check this hardcoded --user parameter in `python-pip` template ? For example as I have found on stackoverflow it is not needed when using venv, so forcing it doesn't have sense:
https://stackoverflow.com/questions/30604952/pip-default-behavior-conflicts-with-virtualenv

jsteinich:
I think right now the `python-pip` template is assuming you aren't using a virtual environment while the `python` template creates one for you.

@cmclaughlin Do you have any thoughts on if it makes sense for the pip version to better support virtual environments?

tl24:
[...]

PR #912, opened 2021-08-17
fix(cli): Don't specify --user flag when running in an pip activated virtualenv

tl24:
This fixes an issue when using the python-pip template running in an activated virtualenv.  Issue #705


packages/cdktf-cli/templates/python-pip/.hooks.sscaff.js
```diff
@@ -32,7 +32,11 @@ exports.post = options => {
   }
 
   writeFileSync('requirements.txt', pypi_cdktf, 'utf-8');
-  execSync('pip3 install --user -r requirements.txt', { stdio: 'inherit' });
+  installArgs = '';
```
ansgarm, line 35:
```suggestion
  let installArgs = '';
```
Let's keep it a local variable, JavaScript works a bit different than Python here 😅

ansgarm (changes requested):

ansgarm:
Thank you for the PR, @tl24 💯 Looks good besides a small thing 🙂

packages/cdktf-cli/templates/python-pip/.hooks.sscaff.js
tl24: Sure thing.  Working in too many different languages, hard to remember all the syntax differences...

ansgarm: Yeah, I thought so 😁 Been there as well

ansgarm (approved):

Issue #881, opened 2021-08-02
Bundle CLI for distribution


skorfmann:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

It might be worthwhile to bundle the CLI for distribution and include all dependencies via esbuild 

- We have a known, tested state in terms of dependencies which we're distributing
- It's probably faster to install
- Using the CLI as a dependency in other projects becomes less of a headache when thinking about conflicting package versions (react, graphql, express, ...)

skorfmann:
An example of what I mean is pretty well illustrated by this diff of `yarn.lock` over here https://github.com/terraform-cdk-providers/cdktf-provider-external/pull/149/files#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2de

ansgarm:
This would probably have prevented #1113

skorfmann:
> So this will also do some tree-shaking to only bundle the code that is used? Similar to how webpack does it in frontend projects?

Yes, exactly

DanielMSchmidt:
Closed by #913

PR #913, opened 2021-08-17
feat(cli): Bundle CLI with esbuild

skorfmann:
Briefly looked into #881 and it doesn't seem to be that hard to bundle it. Played around with it a bit locally, and it seems to be working. Would be great if some other people could test drive this as well :) 

When installing the unbundled CLI as it's published right now:

Screenshot 2021-08-17 at 21 08 52

Screenshot 2021-08-17 at 21 09 41

The bundled & minified version is ~ 15 MB but due to the remained of dependencies (mostly `cdktf`) there's still ~ 120 MB of node modules. Still, that's 50% less.

Thoughts?


packages/cdktf-cli/package.json
```diff
@@ -27,53 +28,8 @@
...
-    "sscaff": "^1.2.0",
-    "stream-buffers": "^3.0.2",
-    "strip-ansi": "^6.0.0",
-    "subscriptions-transport-ws": "^0.9.19",
-    "utility-types": "^3.10.0",
-    "uuid": "^8.3.0",
-    "ws": "^7.5.3",
-    "yargs": "^15.1.0",
-    "zod": "^1.11.7"
+    "cdktf": "0.0.0"
```
skorfmann, line 32: We should check if this dependency is really necessary

ansgarm:
I accidentally tried this with node v12 and got an error, which is fine, but we should mention that in the upgrade notes. We currently already require node v14 but it also happened to work with v12 so far 😅

ansgarm:
Watch fails with
```
demo-office-hours-july-serverless /Users/ansgar/projects/hashicorp/terraform-cdk/packages/cdktf-cli/bin-dist/cdktf watch --auto-approve                                                                                                           11:19:01
Starting server...
[2021-08-19T11:19:06.748] [ERROR] cli-server - internal/modules/cjs/loader.js:905
  throw err;
  ^

Error: Cannot find module '/Users/ansgar/projects/hashicorp/terraform-cdk/packages/cdktf-cli/bin-dist/server.js'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:902:15)
    at Function.Module._load (internal/modules/cjs/loader.js:746:27)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:76:12)
    at internal/main/run_main_module.js:17:47 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}
```

ansgarm (dismissed):
Code looks good, only watch seems to fail (at least the way I invoked it for testing, please assist if that was the case)

packages/cdktf-cli/bin/cmds/helper/init.ts
```diff
@@ -18,7 +18,7 @@ import { sendTelemetry } from "../../../lib/checkpoint";
 
 const chalkColour = new chalk.Instance();
 
-const templatesDir = path.join(__dirname, "..", "..", "..", "templates");
+const templatesDir = path.join(__dirname, "..", "templates");
```
jsteinich, line 21: Will this cause issues when running the non-bundled version during local testing?

skorfmann: If you mean actual usage of the locally build cli, then this shouldn't cause issues, since it's gonna bundle on `postbuild` (takes just one or two seconds)

skorfmann: For actual test cases in terms of unit tests, this would likely cause issues

skorfmann:
> I accidentally tried this with node v12 and got an error, which is fine, but we should mention that in the upgrade notes. We currently already require node v14 but it also happened to work with v12 so far 😅

Hm, v12 should work as well. Gonna look into that

packages/cdktf-cli/package.json
skorfmann: As long as we have a global installation of `cdktf-cli` (e.g. via homebrew), it's a bit weird to have this dependency. This could lead to weird behaviour.

packages/@cdktf/provider-generator/lib/config.ts
```diff
@@ -2,10 +2,11 @@ import * as fs from "fs-extra";
 import * as path from "path";
 import { Language } from "./get/constructs-maker";
 import { env } from "process";
-import { CONTEXT_ENV } from "cdktf";
 import { isRegistryModule } from "./get/module";
 import { toPascalCase } from "codemaker";
 
+const CONTEXT_ENV = "CDKTF_CONTEXT_JSON";
```
DanielMSchmidt, line 8: Good catch 👍

DanielMSchmidt (changes requested):
Looks good, but every command I run just throws this:

```
node:internal/url:1417
    throw new ERR_INVALID_ARG_TYPE('path', ['string', 'URL'], path);
    ^

TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string or an instance of URL. Received undefined
    at new NodeError (node:internal/errors:371:5)
    at fileURLToPath (node:internal/url:1417:11)
    at /Users/danielschmidt/work/terraform-cdk/packages/cdktf-cli/bin-dist/cdktf.js:6:20658
    at /Users/danielschmidt/work/terraform-cdk/packages/cdktf-cli/bin-dist/cdktf.js:1:801
    at /Users/danielschmidt/work/terraform-cdk/packages/cdktf-cli/bin-dist/cdktf.js:64:9341
    at /Users/danielschmidt/work/terraform-cdk/packages/cdktf-cli/bin-dist/cdktf.js:1:801
    at Object.<anonymous> (/Users/danielschmidt/work/terraform-cdk/packages/cdktf-cli/bin-dist/cdktf.js:2632:2010)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:981:32) {
  code: 'ERR_INVALID_ARG_TYPE'
}
```

skorfmann:
Well, not quite where I would like to get with that, but still an improvement:

```
// before
added 596 packages from 385 contributors in 37.248s

// bundled
added 269 packages from 202 contributors in 10.863s
```

Still a few tests failing, continuing tomorrow.

jsteinich:
> Well, not quite where I would like to get with that, but still an improvement:
> 
> ```
> // before
> added 596 packages from 385 contributors in 37.248s
> 
> // bundled
> added 269 packages from 202 contributors in 10.863s
> ```
> 
> Still a few tests failing, continuing tomorrow.

Since there are still a lot of un-bundled packages, would it make sense to pin versions in the release flow as well (or instead of bundling). That way what users install matches what was tested.

skorfmann:
> Since there are still a lot of un-bundled packages, would it make sense to pin versions in the release flow as well (or instead of bundling). That way what users install matches what was tested.

This works for direct dependencies only. Ignoring the recent jsii issues, transient dependencies caused most of the issues, I believe.

jsteinich:
https://docs.npmjs.com/cli/v8/configuring-npm/npm-shrinkwrap-json seems like it would lock the dependencies of the published cli package. I haven't found a way to generate directly from yarn.lock though.

skorfmann:
> https://docs.npmjs.com/cli/v8/configuring-npm/npm-shrinkwrap-json seems like it would lock the dependencies of the published cli package. I haven't found a way to generate directly from yarn.lock though.

Thanks, that might be worth looking into. I believe this works in the context of NPM only, but probably better than nothing. However, I'd not piggyback on this PR.

DanielMSchmidt (approved):
Looks good, I'd suggest to squash the commits though :D

ansgarm (commented):
Still need to test windows somehow (vagrant is currently downloading an image 😅) but looks good so far and works on OS X 👍

ansgarm:
Seems to work on Windows – I don't have Terraform on that machine, but the CLI can be invoked at all, which hints at the rest working (and we also have CI tests running on windows 👍)
image
image

ansgarm (approved):

Issue #916, opened 2021-08-18
Automatically Format PRs 


skorfmann:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

I'm imagining a dedicated PR action, which checks for formatting issues and automatically creates a PR to fix these via `yarn format`

DanielMSchmidt:
I think out pre-commit hook is not properly working, maybe we can fix it instead: https://prettier.io/docs/en/precommit.html

PR #917, opened 2021-08-18
chore(test): Dedicated lint worfkflow

skorfmann:
Allow other tests to proceed when linting fails. First step towards #916

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "examples:integration:python": "test/run-against-dist tools/build-examples.sh python",
     "examples:integration:typescript": "test/run-against-dist tools/build-examples-sequential.sh typescript",
     "examples:integration:go": "test/run-against-dist tools/build-examples-sequential.sh go",
-    "pretest": "prettier --check .",
+    "lint": "prettier --check .",
     "test": "lerna run --scope cdktf* --scope @cdktf* test",
     "watch": "lerna run --parallel  --stream --scope @cdktf/* --scope cdktf* watch-preserve-output",
     "link-packages": "lerna exec --scope cdktf* --scope @cdktf* yarn link",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -18,7 +18,7 @@
     "examples:integration:python": "test/run-against-dist tools/build-examples.sh python",
     "examples:integration:typescript": "test/run-against-dist tools/build-examples-sequential.sh typescript",
     "examples:integration:go": "test/run-against-dist tools/build-examples-sequential.sh go",
-    "pretest": "prettier --check .",
+    "lint": "prettier --check .",
```
DanielMSchmidt, line 21: We should also execute eslint here, shouldn't we?

ansgarm:
Will this add something like over here: https://github.com/hashicorp/vscode-terraform/pull/741/files ?
image

skorfmann:
> Will this add something like over here: https://github.com/hashicorp/vscode-terraform/pull/741/files ?

For things which are auto-fixable it should be automatically fixed. 

The original intention of this PR was, to create a dedicated workflow and then expand on it later. However, just looked for GithubActions and there are plenty. While they aren't creating PRs, a commit on the same branch should probably be ok as well.

e.g. this one here https://github.com/marketplace/actions/lint-action would support multiple linters

package.json
skorfmann: Yeah, good point - Probably gonna try this Action https://github.com/marketplace/actions/lint-action

skorfmann:
Right - so, auto-committing and apparently even commenting wouldn't work with these actions. I guess that the `terraform-vscode` thing works similar to this https://github.com/cschleiden/jest-github-actions-reporter

skorfmann:
So, what are your thoughts on just splitting the linting in a dedicated workflow, without any further enhancements?

PR #921, opened 2021-08-18
chore(lib): BREAKING CHANGE upgrade constructs lib to v10

ansgarm:


DanielMSchmidt:
@ansgarm I think we don't release through semantic-release yet, but just so you are warned for the future: a BREAKING CHANGE anywhere in the commit message will trigger a major release in semantic-release. Ran into this too often :D

packages/cdktf/lib/annotations.ts
```diff
@@ -0,0 +1,79 @@
+// Originally from aws-cdk v2 because with constructs v10 addMessage was removed
+// in favor of Annotations in the AWS CDK
+// https://github.com/aws/aws-cdk/blob/dcae3eead0dbf9acb1ed80ba95bb104c64cb1bd7/packages/@aws-cdk/core/lib/annotations.ts
+import { IConstruct } from "constructs";
+
+export enum AnnotationMetadataEntryType {
```
jsteinich, line 6: I wonder if it would make sense to use the same values as AWS CDK for this and other pieces in an attempt for future interop support.

ansgarm: I thought about that too. In general these are the same levels (info, warn, error) and I think a mapping would be easy enough for a future interop support. I would not want to pull in the AWS CDK as a dependency just for this now.
While using the `Aspect` from the AWS CDK should work, the `Annotations` class contains code that is dependent on other AWS CDK constants which would open a can of worms I think.
And having a hard dependency on the AWS CDK for CDKTF internals might be bad in the future if we want to support a range of AWS CDK constructs which themselves could need different AWS CDK core versions (just a suspicion at this point). Building an adapter would probably yield better results then. Similar to this poc: https://github.com/skorfmann/cfn2tf/blob/interop-poc/aws-adapter.ts

packages/cdktf/lib/terraform-asset.ts
```diff
@@ -63,12 +64,18 @@ export class TerraformAsset extends Resource {
         `TerraformAsset ${id} expects path to be a file, a directory was passed: '${config.path}'`
       );
     }
+
+    if (this.stack.node.root instanceof App) {
+      this.copyAsset(this.stack.node.root.manifest);
+    } else {
+      throw new Error("Could not copy asset because no root app was found");
+    }
```
skorfmann, line 72: I believe that's called too early? Right now, that's called when `app.synth()` is called. With this change, it would be on resource construction.

packages/cdktf/test/stack.test.ts
```diff
@@ -95,13 +94,10 @@ test("stack validation returns error when provider is missing", () => {
...
 
-  const errors = Node.of(stack).validate();
+  const errors = stack.node.validate();
 
   expect(errors).toEqual([
-    expect.objectContaining({
-      message: `Found resources without a matching povider. Please make sure to add the following providers to your stack: test-provider`,
-      source: stack,
-    }),
+    `Found resources without a matching povider. Please make sure to add the following providers to your stack: test-provider`,
```
skorfmann, line 100: Why did we loose the source here?

packages/cdktf/lib/app.ts
```diff
@@ -63,37 +73,66 @@ export class App extends Construct {
...
+            constructPath: node.node.path,
+            level: metadata.type as AnnotationMetadataEntryType,
+            message: metadata.data,
+            stacktrace: metadata.trace,
+          }))
+        )
+        .reduce((list, metadatas) => [...list, ...metadatas], []); // Array.flat()
+
+      // it is readonly but this is the place where we are allowed to write to it
+      (this.manifest.forStack(stack).annotations as any) = annotations;
```
skorfmann, line 126: Could you elaborate on why the and annotations are collected on app level, rather than having the stacks append the data to the manifest?

packages/cdktf/lib/app.ts
```diff
@@ -63,37 +73,66 @@ export class App extends Construct {
...
+    const session: ISynthesisSession = {
       outdir: this.outdir,
       skipValidation: this.skipValidation,
-      sessionContext: {
-        manifest,
-      },
+      manifest: this.manifest,
+    };
+
+    invokeAspects(this);
```
skorfmann, line 99: Why on app level?

packages/cdktf/lib/app.ts
```diff
@@ -109,3 +148,59 @@ export class App extends Construct {
...
+        nestedAspectWarning = true;
+      }
+
+      // mark as invoked for this node
+      invoked.push(aspect);
+    }
+
+    for (const child of construct.node.children) {
+      recurse(child, allAspectsHere);
+    }
```
skorfmann, line 195: I believe it's the same behaviour as https://github.com/aws/constructs/blob/ea0522e9a40ff013751734b9ae986786a9209b43/src/construct.ts#L460 - can you confirm?

packages/cdktf/lib/synthesize/synthesizer.ts
```diff
@@ -0,0 +1,38 @@
...
+    const manifest = session.manifest;
+    const stackManifest = manifest.forStack(this.stack);
+
+    const workingDirectory = path.join(
+      session.outdir,
+      stackManifest.workingDirectory
+    );
+    if (!fs.existsSync(workingDirectory)) fs.mkdirSync(workingDirectory);
+
+    const tfConfig = this.stack.toTerraform();
```
skorfmann, line 31: It just occurred to me: Since we lost `onSynthesize` and `toTerraform` is called on [Terraform Elements](https://github.com/hashicorp/terraform-cdk/blob/d904505481c7daee448b531572b2194f79b4e06f/packages/cdktf/lib/terraform-stack.ts#L157) only, we essentially don't have a way to hook into the synth event for [Resources](https://github.com/hashicorp/terraform-cdk/blob/d904505481c7daee448b531572b2194f79b4e06f/packages/cdktf/lib/resource.ts).

packages/cdktf/test/stack.test.ts
ansgarm: Because it was [dropped](https://github.com/aws/constructs/blob/10.x/src/construct.ts#L477) from `constructs`.

But thinking about it, it was probably dropped because validations now are bound to a single construct, so you must have found a bug here 🤔 I might need to walk the tree to validate all childs as well and by doing so, I can re-add the source..

packages/cdktf/lib/terraform-asset.ts
ansgarm: But wouldn't you have an app in place whenever you instantiate a stack (passing the app) which instantiates the resources (passing the stack which has an app?)

packages/cdktf/lib/synthesize/synthesizer.ts
ansgarm: Do we need to?

packages/cdktf/lib/app.ts
ansgarm: To also support Annotations on the App (which might not make sense / might not be useful at all)

packages/cdktf/lib/app.ts
ansgarm: There's no particular reason to it. But I think you're right: It might be better suited to move this to the `StackSynthesizer` instead.
At the point I wrote this, `StackManifest`s had this [weird behavior](https://github.com/hashicorp/terraform-cdk/blob/d904505481c7daee448b531572b2194f79b4e06f/packages/cdktf/lib/manifest.ts#L37) of being collected in an Array that could also end up with multiple ones of the same (which were reduced when rendering), so I had to make sure I put the annotations on the last `manifest.forStack(stack)` so it would end up as the last item to be reduced and [end up in the result](https://github.com/hashicorp/terraform-cdk/blob/d904505481c7daee448b531572b2194f79b4e06f/packages/cdktf/lib/manifest.ts#L47).

packages/cdktf/lib/app.ts
ansgarm: That's a tough question to answer 😄
It mostly should be (as in calling the Aspect on all children of the node it is registered on), but I suspect the new version is a bit more robust in not calling the same Aspect instance twice on the same resource if it has been added to a parent of said resource as well.

packages/cdktf/lib/terraform-asset.ts
skorfmann: While the app creates the [outdir](https://github.com/hashicorp/terraform-cdk/blob/a5eeb9e5901d5918407d6f0246f383c746391585/packages/cdktf/lib/app.ts#L84) in the constructor, the stack directories are created in the [StackSynthesizer](https://github.com/hashicorp/terraform-cdk/blob/a5eeb9e5901d5918407d6f0246f383c746391585/packages/cdktf/lib/synthesize/synthesizer.ts#L29).

Plus, there might be adjustments happening in Aspects (e.g. changing code - thinking of Lambda stubs).

packages/cdktf/lib/app.ts
skorfmann: Hm, probably not - since the App is mostly a container for stacks.

packages/cdktf/lib/app.ts
skorfmann: Right - was mostly asking about the order of applying https://github.com/aws/constructs/blob/ea0522e9a40ff013751734b9ae986786a9209b43/src/construct.ts#L667-L678

But seems to make sense :)

packages/cdktf/lib/synthesize/synthesizer.ts
skorfmann: It might help for https://github.com/hashicorp/terraform-cdk/blob/d904505481c7daee448b531572b2194f79b4e06f/packages/cdktf/lib/terraform-asset.ts#L26 and I've done it at least once in another project (while I'd have been able to do it differently, it was the most convenient way to achieve what I was trying to do)

ansgarm: Hm, the AWS CDK v2 does not allow users to use such an API anymore it seems. There is [something](https://github.com/aws/aws-cdk/blob/dcae3eead0dbf9acb1ed80ba95bb104c64cb1bd7/packages/@aws-cdk/core/lib/app.ts#L161) to be used internally though. It looks a bit like a workaround for enabling a certain use case.

ansgarm: Here is some background on that: https://github.com/aws/aws-cdk-rfcs/blob/master/text/0192-remove-constructs-compat.md#07-no-synthesize
It seems that we could re-add it ourselves, but I'd like to make sure whether we really need to.

packages/cdktf/test/stack.test.ts
ansgarm: Fixed the bug (now validations are run recursively while synthesizing) and re-added the source to that part of the code.
However, this has no impact on this exact thing, as this uses `Node.validate()` which does not return a source (as it is not recursively)

packages/cdktf/lib/terraform-asset.ts
ansgarm: ✅ Changed the behavior to copy files on synth (added support for a custom synth function)

packages/cdktf/lib/synthesize/synthesizer.ts
ansgarm: Re-added that but for now only exposed it internally (and used it in the Asset). If we want to open this to external usage, we should think about use cases first and whether a custom synth function would be the right way to solve them (instead of e.g. Aspects or alike).

skorfmann (approved):
Just skimmed over it to see what changed compared to the last review. Looks pretty good from my point of view.

CHANGELOG.md
```diff
@@ -1,3 +1,21 @@
+## Unreleased Changes
+
+**Breaking Changes**
+
+### Constructs upgrade (`v3` to `v10`)
+
+The `constructs` package serves as the base for all building blocks of the CDKs (e.g. CDK for Terraform, AWS CDK and cdk8s). Recently the `constructs` package was lifted to v10 with a few but major changes. The reason behind those was to make it future proof and stable across different CDKs. In this process the API surface of the `constructs` base package was removed and some functionality has been moved directly to the AWS CDK instead. For more information and the underlying motivation for the changes to the `constructs` library refer to the [Removal of Construct Compatibility Layer RFC](https://github.com/aws/aws-cdk-rfcs/blob/master/text/0192-remove-constructs-compat.md).
```
DanielMSchmidt, line 7: > In this process the API surface of the `constructs` base package was removed and some functionality has been moved directly to the AWS CDK instead

This doesn't sound like proper english to me, but as it's also not my first language I'd like to summon @danieldreier to keep us grammatically correct :D

packages/cdktf/lib/app.ts
```diff
@@ -63,37 +68,42 @@ export class App extends Construct {
...
-        manifest,
-      },
-    });
+      manifest: this.manifest,
+    };
+
+    const stacks = this.node
+      .findAll()
+      .filter<TerraformStack>(
+        (c): c is TerraformStack => c instanceof TerraformStack
```
DanielMSchmidt, line 97: Uh that's cool TIL  🤯

packages/cdktf/lib/manifest.ts
```diff
@@ -1,54 +1,62 @@
...
+  readonly annotations: StackAnnotation[];
 }
 
 export class Manifest {
   public static readonly fileName = "manifest.json";
   public static readonly stacksFolder = "stacks";
   public static readonly stackFileName = "cdk.tf.json";
 
-  public readonly stacks: StackManifest[] = [];
+  public readonly stacks: Record<StackManifest["name"], StackManifest> = {};
```
DanielMSchmidt, line 25: nice, that gives one more info than just `Record<string, StackManifest>`

packages/cdktf/lib/synthesize/synthesizer.ts
```diff
@@ -0,0 +1,191 @@
...
+        invoked = invokedByPath[node.path] = [];
+      }
+
+      if (invoked.includes(aspect)) {
+        continue;
+      }
+
+      aspect.visit(construct);
+
+      // if an aspect was added to the node while invoking another aspect it will not be invoked, emit a warning
```
DanielMSchmidt, line 128: Cool edge case handling!

packages/cdktf/lib/testing.ts
```diff
@@ -48,18 +49,43 @@ export class Testing {
...
-      },
+      manifest,
     });
 
     manifest.writeToFile();
 
     return outdir;
   }
 
+  public static renderConstructTree(construct: IConstruct): string {
```
DanielMSchmidt, line 62: nice, that will come in handy!

DanielMSchmidt (approved):

CHANGELOG.md
ansgarm: Yeah, you're right :D And I probably meant "was reduced" instead of "was removed" 😅

packages/cdktf/lib/synthesize/synthesizer.ts
ansgarm: (full credit to AWS 😂)

PR #925, opened 2021-08-23
Complex struct attributes

DanielMSchmidt:
Moves #480 over to this origin and completes it


packages/cdktf-cli/lib/get/generator/emitter/attributes-emitter.ts
```diff
@@ -217,11 +135,15 @@ export class AttributesEmitter {
     if (type.isBoolean) {
       return `this.getBooleanAttribute('${att.terraformName}')`;
     }
+    if (type.isList) {
+      return `(this.getListAttribute('${att.terraformName}') as any) as ${att.type.name}`;
```
jsteinich, line 139: This will fail since `string[] != attribute type`

jsteinich:
I'm not sure I see the advantage of using a native array type when it can't actually be used for anything more than a reference.

packages/cdktf-cli/test/get/generator/__snapshots__/resource-types.test.ts.snap
```diff
@@ -2748,9 +2748,9 @@ export class SecurityGroup extends cdktf.TerraformResource {
   }
 
   // egress - computed: true, optional: true, required: false
-  private _egress?: SecurityGroupEgress[]
-  public get egress(): SecurityGroupEgress[] {
-    return this.interpolationForAttribute('egress') as any; // Getting the computed value is not yet implemented
```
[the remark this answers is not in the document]
DanielMSchmidt: Plan changed, will re-add the comment for now

packages/cdktf-cli/lib/get/generator/emitter/attributes-emitter.ts
DanielMSchmidt: Removed this part and created #958 to plan a proper solution

packages/cdktf-cli/lib/get/generator/emitter/attributes-emitter.ts
```diff
@@ -3,6 +3,14 @@ import { AttributeModel } from "../models";
 import { downcaseFirst } from "../../../util";
 import { CUSTOM_DEFAULTS } from "../custom-defaults";
 
+type GetterType =
+  | { _type: "plain" }
+  | {
+      _type: "args";
```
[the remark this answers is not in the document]
DanielMSchmidt: using function now, think thats better

packages/cdktf-cli/lib/get/generator/emitter/attributes-emitter.ts
```diff
@@ -12,217 +20,132 @@ export class AttributesEmitter {
...
+        `public get ${this.getInputName(att, escapeInput)}()`
+      );
+      this.code.line(`return this.${att.storageName}`);
+      this.code.closeBlock();
+    }
   }
 
   public determineGetAttCall(att: AttributeModel): string {
+    if (att.isProvider) {
+      return `this.${att.storageName}`;
```
jsteinich, line 127:
```suggestion
      return `this.${att.storageName};`;
```

packages/cdktf-cli/lib/get/generator/emitter/attributes-emitter.ts
```diff
@@ -12,217 +20,132 @@ export class AttributesEmitter {
...
+        break;
+    }
 
-    this.code.openBlock(`public ${this.getResetName(att.name, escapeReset)}()`);
-    this.code.line(`this.${att.storageName} = undefined;`);
-    this.code.closeBlock();
+    if (hasSetter) {
+      this.code.openBlock(
+        `public set ${att.name}(value: ${att.type.name}${
+          hasResetMethod ? " | undefined" : ""
```
jsteinich, line 100: This doesn't match the existing behavior which only allowed `undefined` for optional attributes of a provider. See https://github.com/hashicorp/terraform-cdk/pull/310 for a discussion on the reasoning behind what was implemented.

jsteinich (changes requested):
The biggest piece missing is actually using the generated computed custom types. That's where the setter removal +/- put method come into place. The config struct (and put method) should use the non-computed version while the getter should use the computed version.

Issue #910, opened 2021-08-17
bug(lib): Fix Token Resolve


skorfmann:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### cdktf & Language Versions

- all

### Affected Resource(s)

- Tokens

### Description

I took some time to compare the synth process of AWS CDK and our implementation. At a first glance, it works similar

#### CfnStack

The AWS CDK stack calls `_toCloudformation` on [all descendant](https://github.com/aws/aws-cdk/blob/ab228a648fa08e8189c010d475426f1ff003ddf7/packages/@aws-cdk/core/lib/stack.ts#L982-L983) descendants and resolves each element individually.

```ts
    const elements = cfnElements(this);
    const fragments = elements.map(e => this.resolve(e._toCloudFormation()));
```

#### TerraformStack

That's more or less what we're doing as well:

https://github.com/hashicorp/terraform-cdk/blob/720d6a782f553d7f1fefb3b22fb84ed43e0cd5a9/packages/cdktf/lib/terraform-stack.ts#L164-L167

#### CfnResource

However, the interesting part happens in [_toCloudFormation](https://github.com/aws/aws-cdk/blob/ebac8bc08885d6862f75b1133752b639dcf54b1c/packages/@aws-cdk/core/lib/cfn-resource.ts#L313) of the `CfnResource`. There we come across the [PostResolveToken](), which essentially postpones the rendering of the properties to a [post processor](https://github.com/aws/aws-cdk/blob/ebac8bc08885d6862f75b1133752b639dcf54b1c/packages/@aws-cdk/core/lib/cfn-resource.ts#L325). 

```ts
          [this.logicalId]: new PostResolveToken({
            Type: this.cfnResourceType,
            Properties: ignoreEmpty(this.cfnProperties),
            DependsOn: ignoreEmpty(renderDependsOn(this.dependsOn)),
            CreationPolicy: capitalizePropertyNames(this, renderCreationPolicy(this.cfnOptions.creationPolicy)),
            UpdatePolicy: capitalizePropertyNames(this, this.cfnOptions.updatePolicy),
[...]

PR #935, opened 2021-08-26
feat(lib): use objects instead of objects in arrays if possible

### LINKED ISSUE #258: Improve UX For List Block Types with Max 1 Element
### Community Note

### Description

There are more then 3000 occasions in the AWS schema where a list type can have at most one element or exactly one element. See this example of t he `AwsProvider` schema and note the `max_items` at the bottom:

```json
      "assume_role": {
        "nesting_mode": "list",
        "block": {
          "attributes": {
            "external_id": {
              "type": "string",
              "description": "The external ID to use when assuming the role. If omitted, no external ID is passed to the AssumeRole call.",
              "optional": true
            },
            "policy": {
              "type": "string",
              "description": "The permissions applied when assuming a role. You cannot use, this policy to grant further permissions that are in excess to those of the,  role that is being assumed.",
              "optional": true
            },
            "role_arn": {
              "type": "string",
              "description": "The ARN of an IAM role to assume prior to making API calls.",
              "optional": true
            },
            "session_name": {
              "type": "string",
              "description": "The session name to use when assuming the role. If omitted, no session name is passed to the AssumeRole call.",
              "optional": true
            }
          }
        },
        "max_items": 1
      },
```

I'm wondering if it makes sense, to not render these schema types as Lists, but as objects.

Rather than:

```ts
    new AwsProvider(this, 'aws', {
      region: 'eu-central-1',
      assumeRole: [{
        roleArn: 'abc',
        sessionName: 'foo',
        externalId: 'bar'
      }]
    })

```

I'd expect this, which also aligns with the Terraform [docs](https://registry.terraform.io/providers/hashicorp/aws/latest/docs#assume-role)

```ts
    new AwsProvider(this, 'aws', {
      region: 'eu-central-1',
      assumeRole: {
        roleArn: 'abc',
        sessionName: 'foo',
        externalId: 'bar'
      }
    })

```

### References

https://github.com/hashicorp/terraform-cdk/issues/253

DanielMSchmidt:
The goal of this PR is to improve the UI surface so that less arrays are being used and direct access of properties is way easier

BREAKING CHANGE: This changes the API surface of every module / provider

Closes #258

### DIFF
--- a/packages/@cdktf/provider-generator/lib/get/__tests__/generator/__snapshots__/complex-computed-types.test.ts.snap
+++ b/packages/@cdktf/provider-generator/lib/get/__tests__/generator/__snapshots__/complex-computed-types.test.ts.snap
@@ -79,7 +79,7 @@ export namespace ACM {
     readonly certificateTransparencyLoggingPreference?: string;
   }
 
-  function acmCertificateOptionsToTerraform(struct?: AcmCertificateOptionsOutput | AcmCertificateOptions): any {
+  function acmCertificateOptionsToTerraform(struct?: AcmCertificateOptionsOutputReference | AcmCertificateOptions): any {
     if (!cdktf.canInspect(struct)) { return struct; }
     if (cdktf.isComplexElement(struct)) {
       throw new Error(\\"A complex element was used as configuration, this is not supported: https://cdk.tf/complex-object-as-configuration\\");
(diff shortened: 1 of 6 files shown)


packages/cdktf-cli/lib/get/generator/resource-parser.ts
```diff
@@ -291,6 +291,7 @@ class Parser {
               struct,
               isOptional: optional,
               isRequired: required,
+              isSingleItem: true,
```
jsteinich, line 294: This isn't actually a list, so this value won't be used. Doesn't hurt anything though.

jsteinich:
Tried to take another look at things. Sorry no in-line comments.

- a few todos in code
- mapper changes may be better done as multi-stage resolve like aws cdk (see https://github.com/hashicorp/terraform-cdk/issues/910)
- init from tf project should be able to use the providers defined rather than requiring passing them in
- config ignored struct changes?
- maps no longer IResolvable?
- lost a generated complex type not supported comment

PR #939, opened 2021-08-30
feat(cli): Support `nested_type` in generator

ansgarm:
Support for tfproto v6

### Open Tasks
- [ ] Write Test(s) for various `nested_type` configurations
- [ ] Write test for minItems & maxItems behaviour!
- [ ] Address all `FIXME` comments


packages/cdktf-cli/lib/get/generator/emitter/struct-emitter.ts
```diff
@@ -65,6 +65,10 @@ export class StructEmitter {
       );
     }
     this.code.closeBlock();
+
+    if (!(struct instanceof ConfigStruct)) {
+      this.emitToTerraformFuction(struct);
+    }
   }
```
jsteinich, line 72: I believe classes are only emitted for computed types.

packages/cdktf-cli/lib/get/generator/models/attribute-model.ts
```diff
@@ -75,6 +75,8 @@ export class AttributeModel {
     if (this._name === "equals") return "equalTo";
     // `node` is already used by the Constructs base class
     if (this._name === "node") return "nodeAttribute";
+    // FIXME: this is only temporary and will not work when this attribute is actually used.
+    if (this._name === "provider") return "providerAttribute";
```
jsteinich, line 79: Seems unrelated. Did this come up?

packages/cdktf-cli/lib/get/generator/models/attribute-type-model.ts
```diff
@@ -30,6 +30,7 @@ export class AttributeTypeModel {
   public isMap: boolean;
   public struct?: Struct;
   public level?: number;
+  // FIXME: do we need an isNested here? maybe for creating different outcomes for the TF JSON?
```
jsteinich, line 33: Other than the actual proto definition file, I'm having trouble finding documentation on the changes to know if anything is different.

packages/cdktf-cli/lib/get/generator/provider-schema.ts
```diff
@@ -24,15 +28,68 @@ export interface Schema {
...
 
-export interface Attribute {
-  type: AttributeType;
+type AttributeNestedTypeNesting =
+  | "invalid"
+  | "single"
+  | "list"
+  | "set"
+  | "map"
+  | "group";
```
jsteinich, line 37: `group` isn't valid for nested types, just regular types. I'm also not sure what it does is that case.

packages/cdktf-cli/lib/get/generator/provider-schema.ts
```diff
@@ -43,7 +100,7 @@ export type AttributeType =
   | ["list", AttributeType]
   | ["object", { [attribute: string]: AttributeType }];
 
-export type BlockType =
+export type BlockType =  // FIXME: use such a type for NestedType Attributes as well?
```
jsteinich, line 103: Might not hurt, but seems to be over specifying compared to the proto definition.

packages/cdktf-cli/lib/get/generator/resource-parser.ts
```diff
@@ -187,7 +190,48 @@ class Parser {
...
+        case "set": {
+          isList = true; // FIXME: check required / optional based on min_items
+          break;
+        }
+        case "map":
+          isMap = true;
+          break;
+        case "single":
+      }
+      const struct = this.addAnonymousStruct(scope, attributeType.attributes); // FIXME: don't make anonymous?
```
jsteinich, line 219: Depends on if a block type exists elsewhere.

jsteinich (commented):
Is there another source of documentation besides the proto definition file?

packages/cdktf-cli/lib/get/generator/models/attribute-model.ts
ansgarm: The latest commit is a "wip" commit 😁 So yeah, that'll go again, but currently I need this for the provider I'm testing with (some private internal one which is also currently wip, so that's why it had this attribute I guess).

packages/cdktf-cli/lib/get/generator/models/attribute-type-model.ts
ansgarm: I have not found anything yet. There is an internal RFC describing the protocol v6 (which seems to be only adding the nested type) but that is a bit broader.

packages/cdktf-cli/lib/get/generator/provider-schema.ts
ansgarm: You're right – I took that from the internal RFC but that does not seem to have made the cut to the implementation / or didn't make sense to implement 👍

packages/cdktf-cli/lib/get/generator/provider-schema.ts
ansgarm: yeah, right.

skorfmann:
@ansgarm is there a path how we can merge this soonish, or should we close this for now and revisit when the need surfaces again?

jsteinich:
> @ansgarm is there a path how we can merge this soonish, or should we close this for now and revisit when the need surfaces again?

I have this on my near-term todo list, but I was planning to just make a new PR since there are a few other pieces mixed in here.

skorfmann:
Closing as suggested by @ansgarm

Issue #676, opened 2021-05-03
bash/zsh completions for cdktf-cli commands


ansgarm:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Yargs [supports specifying completions](http://yargs.js.org/docs/#api-reference-completioncmd-description-fn) for commands and arguments. This would be really useful to autocomplete the stacks when doing e.g. a `cdktf deploy [stack]`.

skorfmann:
That sounds great!

DanielMSchmidt:
I failed getting any autocompletion output out of yargs at `cli-autocompletion` I am not super familiar with this kind of shell interaction, so I'm probably not the ideal person for this task

ansgarm:
Support for shell completions has been introduced via #943.
Currently `yargs` does not support auto completing choices (e.g. `typescript, go, ...` for `--language`). I opened a [PR](https://github.com/yargs/yargs/pull/2018) for that and as soon as that gets merged and released we can update to that `yargs` version to improve our completion support.

PR #943, opened 2021-09-01
feat(cli): Introduce support for shell completions

ansgarm:
(also removes package-lock.json that was committed by accident previously)

Relates to #676.

### Open tasks
- [x] add docs for how to setup autocomplete


packages/cdktf-cli/bin/cdktf.ts
```diff
@@ -27,6 +28,48 @@ if (semver.lt(process.version, "10.12.0")) {
...
 
+const customCompletion = function (
+  _current: string,
+  argv: any,
+  completionsFilter: (
+    done?: (err: any, defaultCompletions: string[]) => void
+  ) => void,
+  done: (completions: string[]) => void
+) {
+  // The completion function gets called when bash completion is invoked and
```
DanielMSchmidt, line 39: Nice, thanks for leaving this here!

packages/cdktf-cli/package-lock.json
```diff
@@ -1,892 +0,0 @@
-{
```
DanielMSchmidt, line 1: Is this by accident?

DanielMSchmidt (approved):

packages/cdktf-cli/bin/cdktf.ts
```diff
@@ -27,6 +28,48 @@ if (semver.lt(process.version, "10.12.0")) {
...
+  if (
+    argv._.includes("diff") ||
+    argv._.includes("deploy") ||
+    argv._.includes("plan") ||
+    argv._.includes("apply")
+  ) {
+    completionsFilter(async (_err, defaults) => {
+      const stacks: string[] = [];
+      try {
+        const manifest = await readCDKTFManifest();
```
jsteinich, line 53: The end result would be the same, but would it be better to check for the existence of the manifest instead?

packages/cdktf-cli/package-lock.json
ansgarm: It had been committed by accident by me, when I built the watch feature 😅

packages/cdktf-cli/bin/cdktf.ts
ansgarm: Left it for my future self as well 😁

packages/cdktf-cli/bin/cdktf.ts
ansgarm: It could also fail if the Manifest would not be readable somehow (e.g. file permissions) or if it would be invalid json (aka be corrupted), so I opted for the safe way here.

Issue #944, opened 2021-09-01
Not able to use S3Backend for terraform-cdk-go


richie-tt:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### cdktf & Language Versions

`github.com/hashicorp/terraform-cdk-go/cdktf v0.5.0`

### Affected Resource(s)

`S3Backend`

### Debug Output

```sh
panic: "No stack could be identified for the construct at path 'backend'"

goroutine 1 [running]:
github.com/aws/jsii-runtime-go/runtime.Create({0x18fd145, 0xf}, {0xc0003bfe38, 0x2, 0x2}, {0x1722260, 0xc000d02400})
        /home/rtkocz/go/pkg/mod/github.com/aws/jsii-runtime-go@v1.34.0/runtime/runtime.go:185 +0x794
github.com/aws/jsii-runtime-go.Create(...)
        /home/rtkocz/go/pkg/mod/github.com/aws/jsii-runtime-go@v1.34.0/deprecated.go:83
github.com/hashicorp/terraform-cdk-go/cdktf.NewS3Backend({0x7f51eb6a63d8, 0xc000d02180}, 0xc000d04f70)
        /home/rtkocz/go/pkg/mod/github.com/hashicorp/terraform-cdk-go/cdktf@v0.5.0/cdktf.go:9963 +0xbc
main.newMyStack({0x7f51eb6a63d8, 0xc000d02180}, {0xc000d02180, 0x1})
        /home/rtkocz/git/AboveCloudDevops/maintanance-tools/disaster-recovery-go/go-cdk/main.go:33 +0x175
main.main()
        /home/rtkocz/git/AboveCloudDevops/maintanance-tools/disaster-recovery-go/go-cdk/main.go:71 +0x8a
exit status 2
```

### Expected Behavior
I would like to use s3backend as a backend for Terraform state.  I was not able to find any example of how to use s3backend in Go, so  If I make a mistake in configuration, please let me know.

### Actual Behavior
It crashes  according to the debug output

### Steps to Reproduce

```go
package main

import (
	"fmt"
[...]

PR #945, opened 2021-09-02
feat(lib): Print a hint when the app has been passed to a backend instead of the stack

ansgarm:
Relates to reported issue #944


DanielMSchmidt (approved):
Looks good to me! I'm sure this will help!

PR #947, opened 2021-09-02
fix(lib): fix source path generated for local modules

### LINKED ISSUE #946: Incorrect module source path generated for local modules

It should be
```
```

```
  bucket = "cdktf-test-bucket"
  acl    = "private"
  force_destroy = true
}
```
```
<project folder>
  |-->modules
  |       |-->module1
  |         |- main.tf
  |- cdktf.json
  |- main.py
```
```
"terraformModules": [
    {
      "name": "module1",
      "source": "./modules/module1"
    }
  ],  
 ```
6. run `cdktf get`
7. update stack to use module
```
#!/usr/bin/env python

        super().__init__(scope, ns)

       Module1(
           self,
           "cdktf_test_bucket",
       )

app = App()

app.synth()
```
8. run `cdktf diff`

```
│


│
```

### PR DESCRIPTION
**Pull request by JonCubed**
This fixes issue #946

### DIFF
--- a/packages/cdktf/lib/terraform-module.ts
+++ b/packages/cdktf/lib/terraform-module.ts
@@ -33,7 +33,7 @@ export abstract class TerraformModule
     super(scope, id);
 
     if (options.source.startsWith("./") || options.source.startsWith("../")) {
-      this.source = path.join("..", options.source);
+      this.source = path.join("../../..", options.source);
     } else {
       this.source = options.source;
     }
--- a/packages/cdktf/test/__snapshots__/terraform-hcl-module.test.ts.snap
+++ b/packages/cdktf/test/__snapshots__/terraform-hcl-module.test.ts.snap
@@ -2,25 +2,12 @@
 
 exports[`add provider 1`] = `
 "{
-  \\"//\\": {
-    \\"metadata\\": {
-      \\"version\\": \\"stubbed\\",
-      \\"stackName\\": \\"test\\",
-      \\"backend\\": \\"local\\"
-    }
-  },
   \\"module\\": {
     \\"test\\": {
-      \\"source\\": \\"../../../foo\\",
       \\"providers\\": {
         \\"test\\": \\"test.provider1\\"
       },
-      \\"//\\": {
-        \\"metadata\\": {
-          \\"path\\": \\"test/test\\",
-          \\"uniqueId\\": \\"test\\"
-        }
-      }
+      \\"source\\": \\"../../../foo\\"
     }
   },
   \\"provider\\": {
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by jsteinich** [comment]
@JonCubed Very much appreciate the fix; however, at least longer term I think that local modules need to be treated like an asset (see #648). The reason being is so that they work when using Terraform Cloud.
Since we haven't actually gotten to putting that in place yet, it may still make sense to use this in the short term though.

**Comment by JonCubed** [comment]
@jsteinich what is proposed for Assets was what I initial thought would happen with modules, so looks good to me long term. Short term I would appreciate if this could be accepted as use local modules is currently broken.

**Comment by javiortizmol** [comment]
> @jsteinich what is proposed for Assets was what I initial thought would happen with modules, so looks good to me long term. Short term I would appreciate if this could be accepted as use local modules is currently broken.

Agree with you @JonCubed, I tested the TerraformAsset option and it didn't work @jsteinich because in the cdk.tf.json file, the synt command is puting an absolute path instead of a relative path and also you can't set a relative path on the TerraformAsset, it should be absolute.
```

        tfEc2Module = TerraformAsset(self, "MyEc2Module",
            path = "/Users/javierortiz/Documents/MyGitHub/tf-cdk/modules/compute"
        )
        my_ec2 = TerraformHclModule(self, "My-ec2", 
            source = tfEc2Module.path,
            variables = {
                "ami": Locals.get("ami").get(config.environment),
                "instance_type": Locals.get("instance_type").get(config.environment),
                "vpc_security_group_ids": Locals.get("vpc_security_group_ids").get(config.environment),
                "subnet_id": Locals.get("vpc").get("public_subnets_ids").get(config.environment),
                "key_name": Locals.get("key_name").get(config.environment),
                "additional_tags": { "Name": f"python-{getWorkspace()}" }                
            }
        )

```
```
╷
│ Error: Module not found
│ 
│ The module address "assets/MyEc2Module" could not be resolved.
│ 
│ If you intended this as a path relative to the current module, use "./assets/MyEc2Module" instead. The
│ "./" prefix indicates that the address is a relative filesystem path.
```
I reviewed the cdk.tf.json file generated and I found `"source": "assets/MyEc2Module"` this is an absolute path, it is the reason behind of the error message.

I modified the cdk.tf.json file manually, changing it for `"source": "./assets/MyEc2Module"`, then on the `cdktf.out/stacks/myStack` folder I ran the `terraform init` command and works.

None of the solutions you mention work @jsteinich, why don't you just accept the solution from @JonCubed ?

jsteinich:
> None of the solutions you mention work @jsteinich, why don't you just accept the solution from @JonCubed ?

I'm not sure the exact order of operations, but @DanielMSchmidt also started a PR for doing the "longer term" solution. Given the asset relative directory complications, I certainly agree with merging this approach. If the other PR finishes in time for the next release, great, but if not then the situation is still improved.

@JonCubed apologizes for not responding sooner. I don't have the ability to merge, but I can certainly poke the rest of the team.

JonCubed:
no worries @jsteinich , I've worked around it by patching my local and docker image. 

I've played around with this some more and one thing that might make this fix undesirable is that it only works for the default outputs i.e. cdktf.out/stacks/<stack name>. In order to make it work for any output folder structure we would need to get the value from the output flag or cdktf.json to determine how far back we need to go. I wasn't sure how I could get this value and if this is a deal breaker or not for the short term.

DanielMSchmidt:
@JonCubed Could you rebase this?

JonCubed:
@DanielMSchmidt rebased

JonCubed:
@jsteinich @DanielMSchmidt I've rebased this again. Any chance we can merge this or co-ordinate so that this can be merged?

DanielMSchmidt (approved):

JonCubed:
@DanielMSchmidt thank you

PR #948, opened 2021-09-02
fix(lib): fix source path generated for local modules

### LINKED ISSUE #946: Incorrect module source path generated for local modules

DanielMSchmidt:
Closes #946


packages/cdktf/lib/terraform-module.ts
```diff
@@ -31,13 +32,20 @@ export abstract class TerraformModule
     super(scope, id);
 
     if (options.source.startsWith("./") || options.source.startsWith("../")) {
-      this.source = path.join("..", options.source);
+      throw new Error(
```
jsteinich, line 35: I don't think this is going to work out very well. Users will expect to be able to use a relative path and it seems like something we should be able to support.

PR #955, opened 2021-09-06
fix(tests): disable terraform cloud test for now

DanielMSchmidt:
It seems like we fail ungracefully if there is a network error and that on CI network errors only happen on windows

### DIFF
--- a/test/test-helper.ts
+++ b/test/test-helper.ts
@@ -208,3 +208,6 @@ export class TestDriver {
     }
   };
 }
+
+export const onWindows = process.platform === "win32" ? it : it.skip;
+export const onPosix = process.platform !== "win32" ? it : it.skip;
--- a/test/typescript/terraform-cloud/test.ts
+++ b/test/typescript/terraform-cloud/test.ts
@@ -6,17 +6,18 @@
  * @group asset
  */
 
-import { TestDriver } from "../../test-helper";
+import { TestDriver, onPosix } from "../../test-helper";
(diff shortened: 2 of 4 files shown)


test/test-helper.ts
```diff
@@ -208,3 +208,6 @@ export class TestDriver {
     }
   };
 }
+
+export const onWindows = process.platform === "win32" ? it : it.skip;
+export const onPosix = process.platform !== "win32" ? it : it.skip;
```
ansgarm, line 213: nice 👍

test/typescript/terraform-cloud/test.ts
```diff
@@ -6,14 +6,14 @@
...
  */
 
-import { TestDriver } from "../../test-helper";
+import { TestDriver, onPosix } from "../../test-helper";
 import { TerraformCloud } from "@skorfmann/terraform-cloud";
 import * as crypto from "crypto";
 const { TERRAFORM_CLOUD_TOKEN, GITHUB_RUN_NUMBER, TERRAFORM_VERSION } =
   process.env;
-const withAuth = TERRAFORM_CLOUD_TOKEN ? it : it.skip;
+const withAuth = TERRAFORM_CLOUD_TOKEN ? onPosix : onPosix.skip;
```
ansgarm, line 14: nit: Maybe call this `onPosixWithAuth` now? Or add a comment above the testcase that this test is currently disabled for windows?

ansgarm (approved):

PR #961, opened 2021-09-09
fix(lib): Fixing path for terraformhclmodule asset approach

javier-caylent:


jsteinich:
@javier-caylent Thank you for taking the time to really dig into the issue and come up with a fix. I think it makes sense to hold off on merging this until either #947 or #948 is merged as they would result in some updates being needed.

For additional context, conceptually it shouldn't matter whether a local module is referenced absolutely or relatively; however, Terraform internally parses them differently. See https://github.com/hashicorp/terraform/blob/e09b831f6ee35d37b11b8dcccd3a6d6f6db5e5ff/internal/addrs/module_source.go#L51

ansgarm:
@jsteinich do you know what we need to do to get this merged?

skorfmann:
Seems like this was addressed in https://github.com/hashicorp/terraform-cdk/pull/1476 - can you confirm @jsteinich ? Close?

jsteinich:
> Seems like this was addressed in #1476 - can you confirm @jsteinich ? Close?

Yep.

PR #1006, opened 2021-09-11
fix(lib): expose testing matchers for other languages properly

DanielMSchmidt:


packages/cdktf/lib/testing/matchers.ts
```diff
@@ -15,6 +15,23 @@ export type SynthesizedStack = {
 };
 export type MatcherReturn = { message: () => string; pass: boolean };
```
jsteinich, line 16: I'm pretty sure this won't work with jsii

packages/cdktf/lib/testing/matchers.ts
```diff
@@ -15,6 +15,23 @@ export type SynthesizedStack = {
...
 
+export interface IMatchers {
+  toHaveDataSourceWithProperties(
+    received: string,
+    resourceType: TerraformConstructor,
+    properties?: Record<string, any>
```
jsteinich, line 22: I think I came across any issue related to and optional record type in jsii. Just using a string keyed map may be needed instead.

jsteinich (changes requested):
It's getting closer, but still some issues.
I only checked Java, but problems that I saw:

- `TerraformConstructor` doesn't work as intended. In native Java, the parameter would end up being `Class<? extends ITerraformResource>` or `Class<T> T extends ITerraformResource`. Neither of those are representable with jsii to my knowledge. C# would be similar.
- `IScopeCallback` also doesn't translate via jsii. [jsii callbacks](https://aws.github.io/jsii/user-guides/language-support/callbacks/) might work, but it might be better to just expose a method to get a testing stack.

I expect the jsii issues will cause issues across all the other non ts/js languages as well.

ansgarm (approved):

PR #1027, opened 2021-09-16
fix(cli): jest config needs to use ts-jest

DanielMSchmidt:


packages/cdktf-cli/templates/typescript/__tests__/main-test.ts
```diff
@@ -1,3 +1,4 @@
+import "cdktf/lib/testing/adapters/jest"; // Load expect autocomplete
```
ansgarm, line 1:
```suggestion
import "cdktf/lib/testing/adapters/jest"; // Load types for expect matchers
```

ansgarm: What do you think, @DanielMSchmidt ? (just an idea)

ansgarm (approved):

Issue #1033, opened 2021-09-16
cli: cdktf cli references manifest from cdktf


skorfmann:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### cdktf & Language Versions

- node v14
- cdktf-cli 0.6.1 (globally installed)

### Affected Resource(s)

- cdktf get

### Debug Output

```
< cdktf get
internal/modules/cjs/loader.js:892
  throw err;
  ^

Error: Cannot find module '../../cdktf/lib/manifest'
Require stack:
```

can be workaround by installing cdktf globally as well, but that's not expected

ASteinheiser:
I appreciate the workaround!

sushantdhiman:
_Workaorund:_ you will need to install `cdktf` globally as well, use `npm install --global cdktf`

ansgarm:
I think we should release the fix for this before 0.7, @skorfmann what do you think?

PR #1040, opened 2021-09-17
chore: lint monorepo for relative imports between packages

ansgarm:
This would have prevented #1033

It would've failed like this:
```
# yarn lint:workspace
yarn run v1.22.10
$ eslint .

/Users/ansgar/projects/hashicorp/terraform-cdk/packages/cdktf-cli/lib/util.ts
  7:37  error  Import for monorepo package 'cdktf' should be absolute  monorepo/no-relative-import

✖ 1 problem (1 error, 0 warnings)

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```


ansgarm:
This currently fails because #1037 is not merged yet 😁 So this works also in CI 🥇

skorfmann (approved):

PR #1042, opened 2021-09-17
chore: remove duplicated jest interface members

hasantayyar:
This duplication is harmless so this change is more like a linting.


ansgarm:
Thanks @hasantayyar!
Just had a look at it and I think it is supposed to be `toHaveDataSourceWithProperties` instead. So it's probably a copy & paste error.
Or have I overlooked something?
Greetings to Berlin!

hasantayyar:
> Thanks @hasantayyar!
> Just had a look at it and I think it is supposed to be `toHaveDataSourceWithProperties` instead. So it's probably a copy & paste error.
> Or have I overlooked something?

Yes it's probably a copy paste mistake but I didn't take the risk for fixing that :) instead just removed the duplication to avoid confusion

ansgarm:
Understandably!
I just added the missing definition and rebased your branch 👍

skorfmann (approved):

Issue #405, opened 2020-10-05
Test Run Times


skorfmann:
With the merged Java support (#360), the examples integration test runtime increased by ~ 30% from 14 to 18 minutes. I'm wondering if we can do something to make this faster (except for disabling the java tests)

![Screenshot 2020-10-05 at 21 57 45](https://user-images.githubusercontent.com/136789/95125907-eb866480-0755-11eb-88d6-b801e534deb6.png)

jsteinich:
I think for examples integration we could switch to using pre-built providers.
For the regular integration tests, parallelizing them would probably make a big difference. Some of them could probably start without providers and then add in what makes sense for the test.

skorfmann:
> test runtime increased by ~ 30% from 14 to 18 minutes.

This gets more pressing. We're now at ~ 35 minutes.
 
![Screenshot 2021-01-14 at 23 42 58](https://user-images.githubusercontent.com/136789/104658083-639fe700-56c2-11eb-85cc-aa71f8b77f48.png)

skorfmann:
> I think for examples integration we could switch to using pre-built providers.

Yes, that could be an option. Examples might end up in a dedicated repository anyway.

> For the regular integration tests, parallelizing them would probably make a big difference. Some of them could probably start without providers and then add in what makes sense for the test.

Yes, the integration tests are running sequentially. One potential optimization could be to run them in parallel. However, the [runners](https://docs.github.com/en/free-pro-team@latest/actions/reference/specifications-for-github-hosted-runners#supported-runners-and-hardware-resources) have 2 cores and 7 GB of memory, so there are tight limits on that.

One thing we can certainly do would be to merge a few of the integration tests, or even refactor one or two of them into a functional / unit test. Other than that, I've been wondering if there's an easy way to automatically split the integration tests and run each of them on its own, either in Github Actions or somewhere else. 

Personally, I'd want to see the total test run times at significantly less than 10 minutes.

jsteinich:
Just some notes:

- We should add timing to each test so that we can narrow down what is taking the most time
- Many of the tests effectively generate providers twice. Can definitely cut it down to once
- I know I came across inefficiencies in jsii in some early testing. Could make contributions there to speed things up
- There certainly are a handful of tests that could be combined
- It is very important to have at least one integration test for each language that does provider generation
- There is a lot of value in validating additional providers generate correctly as well

skorfmann:
[...]

PR #1043, opened 2021-09-17
chore(tests): Build examples as matrix - 1 build per example

skorfmann:
This PR is a first step towards drastically reduced build times - Related to #405 

It tries the concept of building a dynamic build matrix for each example, in order to parallelize these builds


tools/build-examples-sequential.sh
```diff
@@ -9,7 +9,7 @@ export CI='1'
 
 BUILD_TARGET=$1
 
-npx lerna run --stream --concurrency 1 --no-bail --scope @examples/${BUILD_TARGET}* reinstall
-npx lerna run --stream --concurrency 1 --no-bail --scope @examples/${BUILD_TARGET}* build
-npx lerna run --stream --concurrency 1 --no-bail --scope @examples/${BUILD_TARGET}* synth
+npx lerna run --stream --concurrency 1 --no-bail --scope ${BUILD_TARGET}* reinstall
+npx lerna run --stream --concurrency 1 --no-bail --scope ${BUILD_TARGET}* build
+npx lerna run --stream --concurrency 1 --no-bail --scope ${BUILD_TARGET}* synth
```
ansgarm, line 15: If we now target only one test in the `BUILD_TARGET` we can remove this script, can't we? Or does you change only apply to PR builds and we still need this elsewhere (e.g. when running builds on `main`)?

.github/workflows/build.yml
```diff
@@ -51,15 +51,28 @@ jobs:
...
+
   linux_integration:
+    needs: build-test-matrix
     runs-on: ubuntu-latest
     strategy:
-      matrix:
-        terraform: ["0.14.11", "0.15.4"]
-        target: ["typescript", "python", "java", "csharp", "go"]
+      fail-fast: false
+      matrix: ${{fromJSON(needs.build-test-matrix.outputs.examples)}}
```
jsteinich, line 73: Should this be ...outputs.tests instead of ...outputs.examples?

skorfmann: Yes, thanks!

tools/build-examples-sequential.sh
skorfmann: Yep, cleaned it up a bit

.github/workflows/build.yml
```diff
@@ -76,43 +65,66 @@ jobs:
...
+          tests=$(cd test && npx jest --listTests | jq  -R -s --arg cwd "$(pwd)/" -c 'split("\n") | map(select(length > 0)) | map(sub($cwd; "")) | { target: values, terraform: ["0.14.11", "0.15.4"]}')
+          echo $tests
+          echo "::set-output name=tests::$tests"
+
+  linux_integration:
+    needs: prepare-integration-tests
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix: ${{fromJSON(needs.prepare-integration-tests.outputs.tests)}}
```
DanielMSchmidt, line 93: fancy 🚀

DanielMSchmidt (approved):

examples/typescript/backends/remote/package.json
```diff
@@ -1,10 +1,11 @@
 {
-  "name": "@deactivated-examples/typescript-backends-remote",
+  "//": "This example test is disabled via the 'private' attribute since it needs credentials (I assume at least)",
```
jsteinich, line 2: Actually using it would need credentials, but just synthing doesn't need anything.

jsteinich (commented):
Worth noting that there is functionality (aws provider gen, module gen) that is now only tested in examples for non-typescript languages. This isn't a problem now, but should be thought about if example tests are ever removed / moved out of PR flow.

PR #1054, opened 2021-09-21
fix(cli): don't create config.json files if none was present before

DanielMSchmidt:
When running cdktf in a directory without a cdktf.json it would create one in an effort to create a project id, leaving stray files around when doing convert or init. This PR makes it only write to the cdktf.json if there was one detected

### DIFF
--- a/packages/cdktf-cli/lib/checkpoint.ts
+++ b/packages/cdktf-cli/lib/checkpoint.ts
@@ -85,27 +85,31 @@ export async function sendTelemetry(
 function getId(
   filePath: string,
   key: string,
+  forceCreation = false,
   explanatoryComment?: string
 ): string {
-  // If the file doesn't exist, we don't have an ID. So we create a file with the ID for next time
-  const _uuid = uuidv4();
-  const _idFile = {} as Record<string, string>;
-  if (explanatoryComment) {
-    _idFile["//"] = explanatoryComment.replace(/\n/g, " ");
-  }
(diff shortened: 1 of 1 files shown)


examples/csharp/aws/cdktf.json
```diff
@@ -1,8 +1,11 @@
...
-  "terraformProviders": ["aws@~> 2.0"],
+  "terraformProviders": [
+    "aws@~> 2.0"
+  ],
   "context": {
     "excludeStackIdFromLogicalIds": "true"
-  }
-}
+  },
+  "projectId": "87752e30-8860-4e59-98a4-86d2603d31fc"
```
ansgarm, line 10: What happens if folks copy config from examples including the `projectId`?
I think it might make sense to have no `projectId` in our examples – is there a need for them?

DanielMSchmidt: Haven't thought about that. I found it a bit annoying when testing against examples that there is always a diff, but then again if you have a strong feeling about this I can also disable telemetry on my machine

ansgarm: Maybe we can disable telemetry for all tests for all developers working on the CDKTF instead?

DanielMSchmidt: Yeah, I have set the `CHECKPOINT_DISABLE` env var now on my system, you are right that is easier

ansgarm (approved):

PR #1058, opened 2021-09-21
Feedback and edits to project setup PR :-)

laurapacilio:
Hi @skorfmann! 

So, I was reviewing your project setup PR, and I wanted to 1) address your comments and 2) provide feedback. And, it became such that I think it would have been absolutely awful for you to review all of that through the github UI as comments. So instead, I made a branch from your branch where I made some edits (described below). I thought you could review and we could either merge this into your branch and then merge your branch into `terraform-io-docs` OR merge this branch directly into 'terraform-io-docs' and close your initial PR. I hope this is okay, but please let me know if you want me to handle this differently in the future.

In this PR I:
- Moved the cdktf.json page to be underneath the Create and Deploy section, rather than under fundamentals. Thank you so much for adding this - I agree that it should be its own page (and I initially tried to put it somewhere else that it didn't belong). 
- Edited the cdktf.json page. I know you just ported over the original content, so I made some edits that I'm hoping will help give users more context and make the page easier to read. Please feel free to call out any places where I misunderstood the text, and we can adjust! 
- Renamed the cdktf.json page to "Configuration". The reason why is that I want to help new users immediately understand that this is going to be about configuration - they may not know that from the name of the file. Let me know what you think about that.
- Copy-edited the Project Setup page - you did a great job here, I just reworded some things and added 1-2 TODOs.
- Removed the Stacks config stuff from Project Setup and put it on its own page called Stacks in the  "Create and Deploy" section. What do you think about that? 
- Added the Configuration page (cdktf.json), and Stacks page to the sidebar navigation (.erb file) in the "Create and Deploy" Section. 

Thank you so much for your work on this, and thank you so much for the feedback about the navigation and page structure! I really appreciate it and we can continue to make changes as is needed as you work on the docs :-)

### DIFF
--- a/website/docs/cdktf/concepts/fundamentals/cdktf-json.md
+++ b/website/docs/cdktf/concepts/fundamentals/cdktf-json.md
@@ -1,6 +1,13 @@
-# cdktf.json Configuration File
+---
+layout: "docs"
+page_title: "Configuration"
+sidebar_current: "cdktf"
+description: "Learn how to build a CDKTF application from a template, project structure, and configuration. "
+---
 
-With the `cdktf.json` in your project root directory you can configure the behaviour of the Terraform CDK CLI:
+# Configuration
+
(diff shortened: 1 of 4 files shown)


skorfmann (commented):
Awesome, thanks for editing it. Gonna merge and continue in the other PR

PR #1063, opened 2021-09-22
feat(example): add example azure service bus

speedfl:
Hello Hashicorp team

I would like to propose this example for azure service bus with queue trigger 

Best regards

Geoffrey


examples/typescript/azure-service-bus-queue-trigger/README.md
```diff
@@ -0,0 +1,47 @@
...
+The above command will create a folder called `cdktf.out` that contains all Terraform JSON configuration that was generated.
+
+Run cdktf-cli commands
+
+```bash
+cdktf diff
+cdktf deploy
+```
+
+You need then to deploy a function with a [service bus trigger](https://docs.microsoft.com/fr-fr/azure/azure-functions/functions-bindings-service-bus-trigger?tabs=javascript).
```
skorfmann, line 42:
```suggestion
You need then to deploy a function with a [service bus trigger](https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-service-bus-trigger?tabs=javascript).
```

examples/typescript/azure-service-bus-queue-trigger/main.ts
```diff
@@ -0,0 +1,141 @@
...
+      name: `myappsa${uid()}`,
+      resourceGroupName: rg.name,
+      location,
+      accountReplicationType: "LRS",
+      accountTier: "Standard",
+    });
+
+    // Creates the Function
+    const azureFunction = new FunctionApp(this, "myapp-function", {
+      name: "myapp-function",
```
skorfmann, line 112: since I don't have a lot of context in Azure and I'm curious: How is the actual code of the function deployed?

skorfmann (changes requested):
Cool, thanks for the example 👍  Left a suggestion and a question

speedfl:
@skorfmann seems that there is a bug. I already signed the CLA in previous PR and it is not taken into account for this one. And I cannot validate for this one

unabletosign

speedfl:
@skorfmann it seems that I cannot sign again the CLA. 
I already signed it for previous PR and now it is not taken into account for this one
unabletosign

examples/typescript/azure-service-bus-queue-trigger/main.ts
speedfl: When you deploy the function-app you deploy the "container" but not the "application code"

In fact the goal of this approach is to deploy the infrastructure and the "application code" independently as they may have different lifecycle / ownership.

The function must be deployed independently (according to the link I showed in the README.md)
I wrote an article here:
https://www.linkedin.com/pulse/embrace-gitops-cdktf-geoffrey-muselli

Do you want me to do two separate folder in this example ?

I could do:
- infrastructure: the actual code for this example
- function: a sample function and the way to deploy it

Issue #967, opened 2021-09-09
[Docs page] Examples


schersh:
**Description**

:tickets: [Asana ticket](https://app.asana.com/0/1200893566367079/1200893566782686)

:mag: File path: /website/docs/cdktf/examples-index.html.md 

Page listing all available reference architectures + use cases (already available on the GitHub readme). Only suggestion is that we give a bit more context for each example so folks know what they’re getting when they click.
Youtube Playlist

Consider (eventually) using `cdktf convert` to generate examples in many languages

https://github.com/hashicorp/terraform-cdk/tree/main/examples is the starting point for this

laurapacilio:
I have:
- Transferred content from the original outputs GitHub page
- Copyedited the content
- Added some TODOs on the page that need engineering attention

Once engineering reviews the content and makes any necessary additions, I will review/copyedit again and we can mark this ready for publishing.

laurapacilio:
@danieldreier put up an inital PR here: https://github.com/hashicorp/terraform-cdk/pull/1062#pullrequestreview-761351957

I have reviewed!

danieldreier:
@laurapacilio I've merged that PR - I think the next (last?) step is for you to add brief descriptions of the linked examples, so I'm going to assign this over to you. Please re-assign back to me if you want me to write any more on here, or mark me for review on PRs for review.

laurapacilio:
@danieldreier Could you take a look at this PR: https://github.com/hashicorp/terraform-cdk/pull/1073. It's got my first pass at the descriptions with a few todos for you :-)

PR #1073, opened 2021-09-23
Laura add descriptions example drafts

laurapacilio:
Closes https://github.com/hashicorp/terraform-cdk/issues/967.

As promised, I took a stab at building out descriptions for the Examples we provide for various languages. In this PR I:
- Created tables for the examples for each language so that we could have a link to the example + a tiny description of what the app in the example does. 
- Updated all of the example link paths so that they're no longer relative links but full-on links to our github examples. 
- Took a pass at drafting descriptions of examples where I could. Please check these for accuracy - I'm still new to infrastructure, and I'm not familiar with all of these providers. The description should be 1-2 lines maximum - the goal is to 1) help users understand what the application is going to do/show them before they click it and 2) briefly explain any jargon that users will need to know in order to understand the value of the example (e.g. what a module is). Where appropriate, I've added in-text links to other areas of our docs for users who may need more information about a particular concept. 

@danieldreier Could you please just review for accuracy and leave me comments with any missing descriptions? Thank you :-)   @schersh Always value your feedback - let me know what you think/if you have any ideas to make this even more helpful for our users.

### DIFF
--- a/website/docs/cdktf/examples.html.md
+++ b/website/docs/cdktf/examples.html.md
@@ -32,11 +32,14 @@ Follow these hands-on tutorials from HashiCorp Learn: [Write CDK for Terraform C
 |[vault](./examples/typescript/vault) | TBD - one line about what the app does. |
 
 #### Backends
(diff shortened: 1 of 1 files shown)


schersh:
@danieldreier would it be more time-effective to ask the eng team to review this list for examples they built/are familiar with and add the one-line description?

PR #1078, opened 2021-09-24
chore(docs): Terraform Outputs

skorfmann:
First pass for Terraform Outputs


schersh (approved):
What we have here looks good to me, @skorfmann was there other content that you wanted to add?

laurapacilio:
It's just responding to the few TODOS :-) Otherwise, I agree it looks awesome!

PR #1080, opened 2021-09-24
Laura edits 987

laurapacilio:
Hi @skorfmann!

I was talking to @schersh, and she was mentioning how you both discussed how it's odd to talk about outputs in isolation. She and I discussed the idea of combining a few of the related concepts pages into a single page....so what do you both think about this?  It puts Input Variables, Local Values, and Output Values onto one page called "Variables and Outputs." The [Terraform documentation](https://www.terraform.io/docs/language/values/index.html) groups these topics this way as well, so I thought maybe it'd be nice parallelism? 

I made the new page, adjusted the sidebar, and did a bit of work on structuring the page. But I didn't do a super thorough copyedit yet because I wanted to see what you both think of this approach before moving forward. If we like it, maybe you can merge it into your [original PR](https://github.com/hashicorp/terraform-cdk/pull/1078#event-5356242451) @skorfmann, you can work on the copy a bit more, and then Sarah and I can do the copyedit on that one?

Thank you both :-)


skorfmann (approved):
Cool, I like the idea 👍

PR #1083, opened 2021-09-27
feat(cli): generate big providers as submodules

### LINKED ISSUE #572: Generate Providers as Submodules
### Community Note

### Description

`cdktf` generates provider classes without namespaces and exports all generated classes and interfaces. Here's an example from the [AWS provider](https://github.com/terraform-cdk-providers/cdktf-provider-aws/blob/99b4f9e148cbce91076daf0ad48034971ad9fe44/src/index.ts):

```ts
// src/index.ts

export * from './accessanalyzer-analyzer';
export * from './acm-certificate';
export * from './acm-certificate-validation';
// ... a lot more exports
```
which can be used like the following in Typescript:

```ts
import { AccessanalyzerAnalyzer } from '@cdktf/aws-provider'
const new AccessanalyzerAnalyzer(this, 'foo', {})
```

One major problem with this approach are naming conflicts. If `acm-certificate` and `acm-certificate-validation` would export a type with the same name, the provider wouldn't compile anymore. (see https://github.com/hashicorp/terraform-cdk/issues/568 for a specific occurrence of this)

Another issue which popped up quite frequently recently: In particular for the AWS provider, but potentially for all providers which consist of lots data sources / resources, some intellisense like analziers are not working out of the box due to code size limitations (the AWS provider is a single file with about 20 MB in Python). (see https://github.com/hashicorp/terraform-cdk/issues/464)

#### Submodules 

jsii supports [submodules](https://github.com/aws/jsii/blob/4731aeb6ad85c04160b30232e85e3f8a43c712a6/gh-pages/content/specification/2-type-system.md#submodules), which are essentially Typescript namespaces. The example above could look like this:

```ts
export * as AccessanalyzerAnalyzer from './accessanalyzer-analyzer';
export * as AcmCertificate from './acm-certificate';
export * as AcmCertificateValidation from './acm-certificate-validation';
// ... a lot more exports
```

which can be used like the following in Typescript:

```ts
import { AccessanalyzerAnalyzer } from '@cdktf/aws-provider'

const new AccessanalyzerAnalyzer.AccessanalyzerAnalyzer(this, 'foo', {})
```

I don't particularly like the namespace / name duplication here, but I don't really have a better idea right now. 

The implementation seems to be pretty straightforward, it'll be a breaking change though.

### References

- https://github.com/hashicorp/terraform-cdk/issues/464
- https://github.com/hashicorp/terraform-cdk/issues/568

DanielMSchmidt:
This change splits up big providers into submodules, which we expect to speed up compile time on other languages

Closes #572

BREAKING CHANGE: This changes the API surface of big providers and will require manual fixes for the imports e.g. aws.Route53Record becomes aws.Route53.Record


jsteinich:
While convenient, I think the auto splitting may be a bit a too dangerous. As is, if a provider crosses over the threshold, it will suddenly have a breaking change. I think we need to have more control over the process.

Does the change to unique class name scopes only kick in if the provider is being split up? Worth double checking that it does only change in that case.

I think we want to put data sources into the same submodule as their non-data source counterparts. There could perhaps be a nested submodule for each, but the seems likely to be overkill.

PR #1101, opened 2021-09-29
feat(cli): namespace aws provider

### LINKED ISSUE #572: Generate Providers as Submodules

DanielMSchmidt:
This change splits up the AWS providers into submodules, which we expect to speed up compile time on other languages and help with compilation issues when adding more classes / function exports per file

Closes #572

BREAKING CHANGE: This changes the API surface of big providers and will require manual fixes for the imports e.g. aws.Route53Record becomes aws.Route53.Record

This is an alternative approach to #1083

### Follow up work

- [ ] Upgrade all aws provider usages to the latest version
- [ ] Move the config parts into a separate package?


packages/cdktf-cli/lib/get/generator/constants/provider-namespaces.ts
```diff
@@ -0,0 +1,42 @@
...
+const awsNamespaceComments: { [key in keyof typeof aws]: string } = {
+  ACM: "AWS Certificate Manager",
+};
+const PROVIDER_NAMESPACES_COMMENT: {
+  [key in keyof typeof PROVIDER_NAMESPACES]: Record<string, string>;
+} = {
+  aws: awsNamespaceComments,
+};
+
+export function getResourceNamespace(
```
skorfmann, line 22: That would be usable in other contexts as well, right? Like in cdktf convert

packages/cdktf-cli/lib/get/generator/constants/provider-namespaces.ts
```diff
@@ -0,0 +1,42 @@
+export type ResourceNamespace = { name: string; comment: string };
+
+// Namespace to Resource mapping
+const aws = {
+  ACM: ["acm_certificate", "acm_certificate_validation"],
+};
```
skorfmann, line 6: What if there's no mapping, e.g. when a new resource / data source is added?

examples/typescript/aws-cloudfront-proxy/main.ts
```diff
@@ -46,7 +45,7 @@ class MyStack extends TerraformStack {
       allowOverwrite: true,
     });
 
-    new AcmCertificateValidation(this, "certvalidation", {
+    new ACM.AcmCertificateValidation(this, "certvalidation", {
```
skorfmann, line 48: What do you think about the duplicated prefix?

packages/cdktf-cli/lib/get/generator/provider-generator.ts
```diff
@@ -77,32 +78,41 @@ export class TerraformProviderGenerator {
...
+        namespacedResources[namespace].push(resourceModel);
+      } else {
+        files.push(this.emitResourceFile(resourceModel));
+      }
+    });
+
+    for (const [namespace, resources] of Object.entries(namespacedResources)) {
+      files.push(this.emitNamespacedResourceFile(namespace, resources));
+    }
+
```
skorfmann, line 115: That's generating a single file for the namespace?

packages/cdktf-cli/lib/get/generator/constants/provider-namespaces.ts
DanielMSchmidt: Yeah, I'd pull it out into convert (or maybe another shared package) once we settled that this is the right way

packages/cdktf-cli/lib/get/generator/constants/provider-namespaces.ts
DanielMSchmidt: @jsteinich Suggested we could use the prefix in general, I'll resort to that 👍 

When there is no prefix I'd say we treat it as unnamespaced, e.g. if AWS adds a new service

packages/cdktf-cli/lib/get/generator/provider-generator.ts
DanielMSchmidt: Yeah as far as I could tell you can not reopen a namespace, it has to be in one block (I had multiple files with the same namespace before and the last one won)

skorfmann: Cool 👍

packages/cdktf-cli/lib/get/generator/constants/provider-namespaces.ts
skorfmann: How do you mean "prefix by default"?

jsteinich: I was thinking we could specify a prefix string rather than an explicit list of resources. That would allow new resources within a "service" to get automatically added to the namespace.

If no specified prefixes matches, it would just go into the generic aws namespace (like everything is currently). We could add more services/prefixes in the future, but those would be breaking changes (unless we added some type of pointer).

examples/typescript/aws-cloudfront-proxy/main.ts
jsteinich: I'm a bit torn. As the example is currently setup I'd prefer to remove the prefix, but thinking about Java/C# where the import/using is more specific would lead to just seeing `new CertificateValidation`. That's not so bad, but other resources get awfully generic names once the prefix is stripped away. Also increases the likelihood of having a name that conflicts with something else.

DanielMSchmidt: I'll leave the complete part for now. There are some groups that have a lot of resources not prefixed (e.g. VPC) where removing the prefix would not do anything / remove the resource entirely. This leads to pretty unpredictable behaviour. If we want to go that route we should plan it more carefully I believe

packages/cdktf-cli/lib/get/generator/constants/provider-namespaces.ts
DanielMSchmidt: What @jsteinich said is implemented. I also had to add a list of additional prefixes since some groups have resources without a common prefix in them (e.g. VPC)

packages/cdktf-cli/lib/get/generator/constants/provider-namespaces.ts
```diff
@@ -0,0 +1,614 @@
...
+  },
+  workspaces: {
+    name: "Workspaces",
+    comment: "AWS WorkSpaces",
+  },
+  xray: {
+    name: "XRay",
+    comment: "AWS X-Ray",
+  },
+  _datasources: {
```
jsteinich, line 564: This group is a bit strange. Some entries seem like rather generic non-service data sources, but others seem like they would fit within a service.

jsteinich (commented):
@DanielMSchmidt How did you get the list and all of the prefixes? At a glance they look correct, but would take awhile to verify all of them.

We should try to confirm if this fixes the python ide issues.
May also want to check the impact on provider generation time.

packages/cdktf-cli/lib/get/generator/constants/provider-namespaces.ts
DanielMSchmidt: I took the ones in the DataSources side bar entry on the provider

DanielMSchmidt:
@jsteinich I went through the AWS provider by hand trying to be very diligent

skorfmann:
> We should try to confirm if this fixes the python ide issues. 
> May also want to check the impact on provider generation time.

Yes, good points - we should confirm this 👍

packages/@cdktf/hcl2cdk/package.json
```diff
@@ -31,11 +31,11 @@
     "@babel/template": "^7.14.5",
     "@babel/types": "^7.14.5",
     "@cdktf/hcl2json": "*",
+    "@cdktf/provider-generator": "*",
```
ansgarm, line 34: I know this change is not new, but other packages depend on `0.0.0` instead of `*` and I think that's then replaced by the current version somehow.

packages/@cdktf/provider-generator/lib/get/generator/constants/provider-namespaces.ts
```diff
@@ -0,0 +1,614 @@
+type Prefix = string;
+export type ResourceNamespace = {
+  name: string;
+  comment: string;
+  additionalPrefix?: Prefix[];
+};
+
+// We need to order these so that the longest prefixes come first (e.g. acmpa before acm)
```
ansgarm, line 8: I wouldn't want to only rely on key order here – maybe sorting the keys in reverse alphabetical order before lookup could ensure this behaviour as well?

packages/@cdktf/provider-generator/lib/get/generator/constants/provider-namespaces.ts
```diff
@@ -0,0 +1,614 @@
...
+  },
+};
+
+const PROVIDER_NAMESPACES: Record<string, Record<Prefix, ResourceNamespace>> = {
+  aws,
+};
+
+// Comments per namespace mapping
+export function getResourceNamespace(
+  provider: string,
```
ansgarm, line 590: Will provider already be "normalized" here? Reason: Passing "hashicorp/aws" should be treated as "aws" (which could happen when a provider is specified as `hashicorp/aws` in the `cdktf.json` file.

packages/@cdktf/provider-generator/lib/get/generator/provider-generator.ts
```diff
@@ -152,6 +160,48 @@ export class TerraformProviderGenerator {
...
+    const ns = resources[0].namespace;
+    const comment = ns?.comment;
+    const name = ns?.name || "";
+    const baseFilePath = resources[0].filePath
+      .split("/")
+      .slice(0, -1)
+      .join("/");
+    const filePath = `${baseFilePath}/${name}.ts`;
+
+    // TODO: find better file path
```
ansgarm, line 176: Is this TODO obsolete?

packages/@cdktf/provider-generator/lib/get/generator/provider-generator.ts
```diff
@@ -152,6 +160,48 @@ export class TerraformProviderGenerator {
...
+    const importStatements = [
+      ...new Set( // make list unique
```
ansgarm, line 178: nice!

packages/@cdktf/provider-generator/lib/config.ts
```diff
@@ -3,7 +3,7 @@ import * as path from "path";
 import { Language } from "./get/constructs-maker";
 import { env } from "process";
 import { CONTEXT_ENV } from "cdktf";
-import { isRegistryModule } from "@cdktf/hcl2cdk";
+import { isRegistryModule } from "./get/module";
 
 const CONFIG_FILE = "cdktf.json";
 const DEFAULTS = {
```
ansgarm, line 9: As Github does not allow commenting on moving files I comment here:
I'm a bit unsure about moving `readConfig` to `@cdktf/provider-generator` – as `readConfig` concerns the cdktf config. For example:
```
export interface Config {
  readonly app?: string;
  readonly language?: Language;
  readonly output: string;
  readonly codeMakerOutput: string;
  terraformProviders?: TerraformProviderConstraint[];
  terraformModules?: TerraformModuleConstraint[];
  checkCodeMakerOutput?: boolean;
  readonly context?: { [key: string]: any };
}
```

I can understand that parsing the config is required here and a circular dependency would be worse – but I'm still unsure about the naming of this package (which I like for what it does) and having things like `readonly app?: string;` in here. No blocker for me, just wanted to surface this.

ansgarm: Ah, just saw your "Move the config parts into a separate package?" follow up task :D

packages/@cdktf/hcl2cdk/package.json
DanielMSchmidt: Yeah sounds good

packages/@cdktf/provider-generator/lib/get/generator/constants/provider-namespaces.ts
DanielMSchmidt: Are you sure this is going to make it clearer? I feel like while it's not great either way, but this way it's at least obvious

ansgarm: To me a comment like "the longest matching prefix will determine the namespace" would make that clear while being safer than relying on order (which I'm unsure if it is guaranteed at all across Node.js versions 🤔).

packages/@cdktf/provider-generator/lib/get/generator/constants/provider-namespaces.ts
DanielMSchmidt: Yes, it's normalized: https://github.com/hashicorp/terraform-cdk/blob/5bf105cb1a8cf3477722c93d8869d431d75f7888/packages/@cdktf/provider-generator/lib/get/generator/provider-generator.ts#L74

packages/@cdktf/provider-generator/lib/get/generator/provider-generator.ts
DanielMSchmidt: yes, thank you

packages/@cdktf/provider-generator/lib/config.ts
DanielMSchmidt: Yeah, this is meant to unblock our two workstreams, not as the final package architecture I'd envision for the generation parts

ansgarm (approved):

Issue #998, opened 2021-09-10
[Docs page] Configuration


schersh:
**Description**

:tickets: [Asana ticket](https://app.asana.com/0/1200893566367079/1200893566782679)

:mag: File path: /website/docs/cdktf/cli-reference/configuration.html.md

Talk about the [json config file](https://github.com/hashicorp/terraform-cdk/blob/main/docs/working-with-cdk-for-terraform/cdktf-json.md)
- How to invoke terraform cli to work around limitations
- Explain that you can use CDKTF to plan/deploy your application (you don’t have to just run Terraform on the output)
- Note that some upcoming features will only be possible when running this through the CLI

schersh:
@schersh update ticket, remove empty file path

schersh:
@laurapacilio the current /configuration file actually has the Telemetry and other info included. I think I will just open a PR that adds a link to the cdktf-json file, and if that PR is approved, we can mark this "Ready to publish"

PR #1111, opened 2021-09-30
chore(docs): Update cli configuration doc

schersh:
Closes https://github.com/hashicorp/terraform-cdk/issues/998 

This PR adds a link to the cdktf.json documentation.

### DIFF
--- a/website/docs/cdktf/cli-reference/configuration.html.md
+++ b/website/docs/cdktf/cli-reference/configuration.html.md
@@ -2,22 +2,10 @@
 layout: "docs"
 page_title: "CLI Configuration"
 sidebar_current: "cdktf"
-description: "Learn to install and configure the CDKTF Command Line Interface."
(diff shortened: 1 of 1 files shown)


laurapacilio (approved):
Thank you! :-) 

My only question is - I originally felt like the telemetry stuff belonged here because it's really applying to the CLI but now we have our own telemetry page. Should we shorten that here and link to the larger telemetry page that Daniel wrote?

schersh:
> Thank you! :-)
> 
> My only question is - I originally felt like the telemetry stuff belonged here because it's really applying to the CLI but now we have our own telemetry page. Should we shorten that here and link to the larger telemetry page that Daniel wrote?

I'll add a link to the primary telemetry page, thanks!

laurapacilio (approved):
Thank you! Ship it! 🚀

Issue #972, opened 2021-09-09
[Docs page] CDKTF Architecture


schersh:
**Description**

:tickets: [Asana ticket](https://app.asana.com/0/1200893566367079/1200893566367087)

:mag: File path: /website/docs/cdktf/concepts/architecture.html.md

Explain the key components of CDKTF itself and how they fit together | include a diagram 

- Library vs. CLI
- ([Apps & Stacks](https://github.com/hashicorp/terraform-cdk/blob/main/docs/working-with-cdk-for-terraform/app-stacks-concept.md)) An Application is the cdktf code, so the program that defines the terraform configuration. A Stack is a part of this application, think of stacks as environments (in HCL you would use terragrunt or have multiple folders for each environment). An environment could be staging / dev / prod or a separate application (e.g. you could define a stack for your CMS application and one for you E-Commerce and invoke the stack constructor with different parameters to create different versions of these environments)
- [Constructs](https://github.com/aws/constructs) in applications
- Tokens

schersh:
@skorfmann will prioritize this issue this week

PR #1112, opened 2021-10-01
chore(docs): CDK for Terraform Architecture

skorfmann:
Fixes: https://github.com/hashicorp/terraform-cdk/issues/972


skorfmann:
FYI: That's heavily WIP

PR #1116, opened 2021-10-04
fix(cli): directly depend on cli-spinners v2.6.0

ansgarm:
This tries to fix the build & might fix installation of `cdktf-cli` when using yarn (`resolutions` is a yarn feature, the npm equivalent is not available yet).


DanielMSchmidt (approved):

RomeroGaliza:
Thanks for the patch! 💛

PR #938, opened 2021-08-29
fix(hcl2cdk): Always using module binding for convert command

jsteinich:
Fixes #859 
Fixes #860 

Did a bit of refactoring in order to be able to reuse the code for parsing module sources.

I'll rebase once #929 has been merged in.

PR #1118, opened 2021-10-04
test(hcl2cdk): re-enable test case for convert

ansgarm:


DanielMSchmidt (approved):

jsteinich:
This looks very similar to what I was working on with #938. I hope to have that updated shortly.

jsteinich:
I ended up creating #1120 since the refactorings made updating #938 more trouble than it was worth.

PR #1122, opened 2021-10-05
chore(deps): update aws provider

DanielMSchmidt:
We use inconsistent versions across our docs and in the tests, I aligned them to the latest version

### DIFF
--- a/packages/@cdktf/provider-generator/lib/get/__tests__/provider.test.ts
+++ b/packages/@cdktf/provider-generator/lib/get/__tests__/provider.test.ts
@@ -7,7 +7,7 @@ import {
 const getProvider = (constraint: TerraformDependencyConstraint) =>
   expectImportMatchSnapshot(constraint);
 
-getProvider(new TerraformProviderConstraint("aws@= 2.60.0"));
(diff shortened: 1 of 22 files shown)


packages/@cdktf/provider-generator/lib/get/__tests__/provider.test.ts
```diff
@@ -7,7 +7,7 @@ import {
 const getProvider = (constraint: TerraformDependencyConstraint) =>
   expectImportMatchSnapshot(constraint);
 
-getProvider(new TerraformProviderConstraint("aws@= 2.60.0"));
+getProvider(new TerraformProviderConstraint("aws@= 3.61.0"));
```
jsteinich, line 10: For everything besides this test, I think it may make sense to use `~> 3.0` instead. That better matches the provider documentation.

There is an argument to be made that since the examples are used as extended integration test, that they shouldn't change arbitrarily; however, I'd rather know sooner than later about a new AWS provider breaking something.

skorfmann:
This is not relevant anymore - docs and examples were updated in other PRs

Issue #1084, opened 2021-09-27
Design improved provider get experience


danieldreier:
This is a meta-issue tracking a set of possible improvements around using providers. This is a research and design task. 

There are several issues to be solved:

1. Users do not reliably understand how to use pre-built providers [Now]
https://github.com/hashicorp/terraform-cdk/issues/831
https://github.com/hashicorp/terraform-cdk/issues/953
https://github.com/hashicorp/terraform-cdk/issues/318

2. Interacting with cdktf.json is a point of friction while getting started [Now]
https://github.com/hashicorp/terraform-cdk/issues/228
https://github.com/hashicorp/terraform-cdk/issues/160
https://github.com/hashicorp/terraform-cdk/issues/228

3. Synthesizing CDKTF provider bindings is relatively slow [Later] 
https://github.com/hashicorp/terraform-cdk/issues/791
https://github.com/hashicorp/terraform-cdk/issues/578

4. Finding provider documentation is hard on people [Later]
https://github.com/hashicorp/terraform-cdk/issues/748

**Goal**
The purpose of this task is to reserve design time for designing a simplified UX that address as many of these inter-connected issues as can be feasibly built within a one-month iteration. The outcome of this should be scoped, ticketed work in the next release (and subsequent ones, if it won't fit) that references back to this issue.

**Proposed Design Direction**
My sense is that this is a case where the CLI workflow should be improved such that people don't need to understand (or read documentation on) provider generation as a pre-requisite to using cdktf in typical, basic use cases. I'm open to alternative solutions.

I believe there are two parts to solving this:
1. Improve UX for interacting with the provider configuration and generation workflow, and
2. Fix performance issues, metadata information, odds and ends of small papercut issues that are technically not related but are thematically connected to this UX

My proposal (to solve items 1 and 2) is for a [CLI subcommand](https://github.com/hashicorp/terraform-cdk/issues/877) that is the primary interface to adding, removing, and updating providers. Ideally, this would also help people generate/find provider documentation, but that might be too much scope expansion.

See also:
* https://github.com/hashicorp/terraform-cdk/issues/218
* [local providers](https://www.terraform.io/docs/cli/config/config-file.html#development-overrides-for-provider-developers) are notoriously hard for people to figure out how to install. Ideally, we should provide a simple workflow that allows people to install, generate bindings for, and use a local provider without having to figure out what Terraform folder to put them in.

chiayuan:
So excited about this improvement!
[...]

PR #1125, opened 2021-10-05
feat(cli): Enable to override 'checkCodeMakerOutput'

mypmc:
To synthesize using [Bazel](https://bazel.build/) and [rules_nodejs](https://github.com/bazelbuild/rules_nodejs).

Here is an example, snipped unnecessary parts, of `BUILD.bazel`.

```python
ts_project(
    name = "compile",
    srcs = ["main.ts"],
    deps = [
        ":code_maker_output",
        "@npm//cdktf",
        "@npm//constructs",
    ],
)

cdktf(
    name = "synth",
    outs = [
        "cdktf.out/manifest.json",
        "cdktf.out/stacks/a/cdk.tf.json",
        "cdktf.out/stacks/b/cdk.tf.json",
        "cdktf.out/stacks/c/cdk.tf.json",
    ],
    args = [
        "synth",
        "--output",
        "$(@D)/cdktf.out",
    ],
    data = [
        ":compile",
        ":main.js",
    ],
    env = {
        "CDKTF_APP": "node $(execpath :main.js)",
    },
)
```

The build `:synth` above fails by `fs.pathExists(config.codeMakerOutput)`.
Because, from the bazel's point of view, it is `:compile` not `:synth` that depends on `:code_maker_output`.
Disabling this check allow us to synthesize easily with Bazel.


jsteinich:
@mypmc at what point are you running `cdktf get`? The check exists as a means to improve the error messaging when it has been ran but should have been.

mypmc:
> at what point are you running `cdktf get`?

Just before `:compile` like this.

```python
ts_project(
    name = "code_maker_output",
    srcs = glob([".gen/**/*.ts"]),
    tsconfig = ":tsconfig",
    deps = [
        "@npm//cdktf",
        "@npm//constructs",
    ],
)
```

> The check exists as a means to improve the error messaging when it has been ran but should have been.

Make sense.
The problem is that the check, which is for improving error messages, prevents synthesizing in our case.

mypmc:
From 63225d14e8bcd5f7f10a735fa62a480da4e041c8, it turns out that the above workaround does not work.
If you have decided which option to go(or both?), let me know.

jsteinich:
@mypmc apologies for the delay. I'll make sure to poke the team on this once they are back in the new year.

skorfmann (changes requested):
> I'd also lean towards either option 1 or 3. However, I'd keep in mind, that we might touch this when we [improve the provider get experience](https://github.com/hashicorp/terraform-cdk/issues/1084). For example further checks could be introduced and might call for renaming the flag to something like `disableGeneratedProvidersCheck` or similar.

Since this flag is essentially doing what `checkCodeMakerOutput` is checking, it should be somehow streamlined. So,  leaning towards option 3 and could make it a dedicated flag on top of this if necessary. No strong opinion on the naming.

packages/@cdktf/provider-generator/lib/config.test.ts
```diff
@@ -23,6 +23,19 @@ describe("parseConfig", () => {
     `);
   });
 
+  it("parses 'checkCodeMakerOutput'", async () => {
```
jsteinich, line 26: Maybe rename this test since it's not obvious that it is parsing correctly (even though the output is correct).

packages/@cdktf/provider-generator/lib/config.ts
```diff
@@ -245,8 +245,11 @@ export const parseConfig = (configJSON?: string) => {
     ...JSON.parse(configJSON || "{}"),
   };
 
-  config.checkCodeMakerOutput =
-    isPresent(config.terraformModules) || isPresent(config.terraformProviders);
+  if (config.checkCodeMakerOutput == undefined || config.checkCodeMakerOutput) {
```
jsteinich, line 248:
```suggestion
  if (config.checkCodeMakerOutput === undefined || config.checkCodeMakerOutput) {
```

packages/cdktf-cli/bin/cmds/synth.ts
```diff
@@ -22,6 +22,11 @@ class Command implements yargs.CommandModule {
         desc: "Stack to output when using --json flag",
         type: "string",
       })
+      .option("check-code-maker-output", {
```
jsteinich, line 25: I think this isn't needed since the value specified in `cdktf.json` will be honored.

jsteinich (changes requested):
A couple nit things and I think we no longer need the cli option. Happy discuss further though.

packages/cdktf-cli/bin/cmds/synth.ts
mypmc: It seems a bit unnatural to me that such a simple flag is an option that can only be specified in a config file. 

The field `checkCodeMakerOutput` seems to be used only in `synth` so far.
So, IMAO, it seems most explicit to remove the field from the Config interface and implement it as a method.
WDYT??

mypmc: I tried to implement it, WDYT??

jsteinich: There currently aren't very many arguments to the `synth` command. I believe there is value in keeping it that way, especially since there is already a configuration file.

Also, `codeMakerOutput` is already only present in the configuration file, so I think it makes sense to keep an option corresponding to its use next to it.

All that said, it probably doesn't really matter. Either way it is functionality that needs supporting.

packages/cdktf-cli/bin/cmds/synth.ts
```diff
@@ -37,6 +37,11 @@ class Command implements yargs.CommandModule {
         desc: "Provide JSON output for the generated Terraform configuration.",
         default: false,
       })
+      .option("check-code-maker-output", {
+        type: "boolean",
+        desc: "Whether if check `codeMakerOutput` existence",
```
jsteinich, line 42:
```suggestion
        desc: "Should `codeMakerOutput` existence check be performed? By default it will be checked if providers or modules are configured.",
```

packages/cdktf-cli/bin/cmds/synth.ts
skorfmann: We have a pretty similar pattern on other commands (e.g. https://github.com/hashicorp/terraform-cdk/blob/a9879d4726c18138ac556a8387934bf786825c52/packages/cdktf-cli/bin/cmds/deploy.ts#L23-L34). Don't have a strong opinion, I'd be ok with having that flag.

mypmc: I also don't have a strong opinion too. Please decide which way should I go. @jsteinich 

1. revert 3950895 and then remove the flag
2. revert 3950895 and then leave the flag
3. as is ( leave only the flag

My preference is 3, next is 2.

jsteinich (approved):
Given that no one has a super strong opinion on this, let's just keep it moving.
Thanks again @mypmc for working on this.

jsteinich:
@mypmc would you be up for rebasing this to resolve conflicts?

PR #1148, opened 2021-10-15
feat(lib): change the namespace of encoded string, list and number tokens

ansgarm:
To make sure CDK for Terraform tokens don't conflict with AWS CDK
tokens we change the namespace of string, list and number tokens.
For string tokens we use e.g. `${TFToken[1]}` instead of `${Token[1]}`.
Likewise for list tokens (they will become `#{TFToken[1]}`).
For number tokens we change the left part of the bitmask from `0xfbff`
to `0xfdff` (which results in flipping bit 25 instead of bit 26).

This change makes sure that tokens don't get accidentally resolved
when e.g. a CDKTF token gets passed to an AWS CDK construct. However,
it won't achieve full interoperability between the two token systems.

### DIFF
--- a/website/docs/cdktf/concepts/data-sources.html.md
+++ b/website/docs/cdktf/concepts/data-sources.html.md
@@ -57,7 +57,7 @@ export class HelloTerraform extends TerraformStack {
             }
         });
 
-        new AwsInstance(this, 'foo', {
+        new aws.EC2.Instance(this, 'foo', {
             ....
             subnetId: remoteState.get('subnet_id')
         });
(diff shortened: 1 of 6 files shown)


skorfmann:
One build was cancelled due to Docker rate limits

```
  Warning: Docker pull failed with exit code 1, back off 2.751 seconds before retry.
  /usr/bin/docker pull docker.mirror.hashicorp.services/hashicorp/jsii-terraform
  Error: The operation was canceled.
```

Gonna merge anyway.

PR #1149, opened 2021-10-15
feat(docs): Aws Adapter

ansgarm:
Adds docs for the new AWS Adapter (originally #1123, now moved to separate repository [cdktf-aws-cdk](https://github.com/hashicorp/cdktf-aws-cdk))


ansgarm:
Thanks @laurapacilio 💯
Added all suggestions as they were ✨

PR #1179, opened 2021-10-21
feat(lib): cross-stack references

DanielMSchmidt:
This is a POC on Cross Stack References, the docker example works: if one runs `cdktf deploy images && cdktf deploy demo-cdktf-ts-docker` the nginx image defined in the images stack gets used in the demo-cdktf-ts-docker stack to provision a resource.

## Todos

- [x] Pass backend configuration into terraform remote state definition
- [x] Use Terraform Remote State based on originStacks backend (e.g. by finding the backend instance and calling a "getRemoteState" method)
- [x] Handle local remote state through asset to ensure compatibility with TFC
- [x] Add stack to all `ref()` calls
- [x] Create a dependency graph between stacks
- [x] Write dependencies into metadata / manifest
- [x] Ensure there can be no loops in the dependency graph
- [x] Handle same TFExpressions object being referenced in different stacks / contexts
- [ ] Use TerraformStack path instead of name
- [ ] Check if we can throw a good error if not all of a stacks dependencies has been run
- [ ] Add tests
  - [ ] Check if x-stack refs work within TF Functions
  - [ ] Check if x-stack refs work passing the ref through multiple stacks
  - [ ] Check if x-stack refs work when the same variable is used in different stacks (`const x = foo.bar` since this is only one ref)
  - [ ] Check if x-stack refs work through JSII
- [ ] Add CLI support for deploying all dependant stacks
- [ ] Optional: Add CLI flag to deploy all stacks?
- [ ] Ensure compatiblity with watch, ideally ensure users specify the all dependent provider flag if cross stack refs are used


skorfmann:
On a highlevel, this is a test how this works in AWS CDK

- https://github.com/aws/aws-cdk/blob/5e8e7abc16ba392f325ee4bbae05ebfee20f043f/packages/%40aws-cdk/aws-iam/test/auto-cross-stack-refs.test.ts

On a more detailed level, here is how it's implemented:

- https://github.com/aws/aws-cdk/blob/ab228a648fa08e8189c010d475426f1ff003ddf7/packages/@aws-cdk/core/lib/private/synthesis.ts#L24
- https://github.com/aws/aws-cdk/blob/ab228a648fa08e8189c010d475426f1ff003ddf7/packages/%40aws-cdk/core/lib/private/prepare-app.ts#L18
- https://github.com/aws/aws-cdk/blob/ab228a648fa08e8189c010d475426f1ff003ddf7/packages/%40aws-cdk/core/lib/private/refs.ts#L39
- https://github.com/aws/aws-cdk/blob/ab228a648fa08e8189c010d475426f1ff003ddf7/packages/%40aws-cdk/core/lib/private/refs.ts#L109
- https://github.com/aws/aws-cdk/blob/ab228a648fa08e8189c010d475426f1ff003ddf7/packages/%40aws-cdk/core/lib/private/refs.ts#L164

There it's done before synthesizing. There's also an explicitly defined dependency as part of this process (see https://github.com/aws/aws-cdk/blob/ab228a648fa08e8189c010d475426f1ff003ddf7/packages/%40aws-cdk/core/lib/deps.ts#L27) which is likely used for figuring out the synth, deployment order. 

The export / import functionality seems to be more or less just adding the correct resources to the producing and consuming Stack. 

Is there a reason why we wouldn't be able to do it in a similar fashion?

DanielMSchmidt:
> Is there a reason why we wouldn't be able to do it in a similar fashion?

Not sure if I got back to you on this one somewhere else: This is exactly the plan, I spared the dependency parts in the POC since they involved fewer unknowns :)

packages/cdktf/lib/backends/remote-backend.ts
```diff
@@ -14,6 +14,18 @@ export class RemoteBackend extends TerraformBackend {
...
+  public getRemoteStateDataSource(
+    scope: Construct,
+    name: string,
+    _fromStack: string
+  ): TerraformRemoteState {
+    return new DataTerraformRemoteState(scope, name, {
+      organization: this.props.organization,
+      workspaces: this.props.workspaces,
+      hostname: this.props.hostname,
+      token: this.props.token,
```
jsteinich, line 26: Would be nice to just use the entire existing props here.

packages/cdktf/lib/terraform-backend.ts
```diff
@@ -1,17 +1,36 @@
...
 
+  /**
+   * Creates a TerraformRemoteState resource that accesses this backend.
+   */
+  public getRemoteStateDataSource(
+    _scope: Construct,
+    _name: string,
+    _fromStack: string
+  ): TerraformRemoteState {
+    throw new Error("This Backend is not implemented yet");
```
jsteinich, line 31: If we implement for all backends to start with, probably change this to an abstract method. If we don't, we'll want to make this error a bit more helpful.

packages/cdktf/lib/terraform-stack.ts
```diff
@@ -136,12 +141,34 @@ export class TerraformStack extends Construct {
       : "";
   }
 
+  private findAll<T>(ClassConstructor: Constructor<T>): T[] {
+    const items: T[] = [];
+
+    const visit = async (node: IConstruct) => {
```
jsteinich, line 147: Does this need to be async? I maybe thinking too much about c#, but seems like this is synchronous.

packages/cdktf/lib/terraform-stack.ts
```diff
@@ -136,12 +141,34 @@ export class TerraformStack extends Construct {
...
+
+    return resolve(this, items);
+  }
+
   public allProviders(): TerraformProvider[] {
-    const providers: TerraformProvider[] = [];
+    return this.findAll(TerraformProvider);
+  }
+
+  public get backend(): TerraformBackend {
```
jsteinich, line 166: Why not use the `findAll` method?

packages/cdktf/lib/terraform-stack.ts
```diff
@@ -184,6 +220,59 @@ export class TerraformStack extends Construct {
 
     return resolve(this, tf);
   }
+
+  public registerOutgoingCrossStackReference(identifier: string) {
+    if (this.crossStackOutputs[identifier]) {
+      return this.crossStackOutputs[identifier];
+    }
+
+    const stack = TerraformStack.of(this);
```
jsteinich, line 229: Why not just use `this`?

jsteinich (commented):
Looks like this is coming along pretty well.

PR #1210, opened 2021-10-25
fix(provider-generator): use put setter so that getter returns required value

### LINKED ISSUE #1200: Prebuilt AWS Typing Error on Generated Resource Properties
### Community Note

### cdktf & Language Versions
```
cdktf --version
0.7.0

"typescript": "4.4.4"
"@cdktf/provider-aws": "2.0.13"
```

### Affected Resource(s)
```
IAM.IamRole
S3.S3Bucket
```
and maybe more...

### Debug Output
https://gist.github.com/vu-bui/7001cd3001792b9d447b1874e9e5e0dc

### Expected Behavior
The properties that are specified as optional inputs and are auto-generated by terraform should not be `undefined` as they're Resolvable

### Actual Behavior
`tsc` throws compile error

DanielMSchmidt:
Fixes #1200

@skorfmann and I talked about this and we found out that we have to make some decisions here:

## Decision 1: Mutation via Setter or Put Method
If we decide to have setters we need to have the same type for the getter. If a value is computed or not, as long as it's optional it's typed as optional, making deep property access harder.

```ts
interface MyPropertyType {
  foo?: string; 
  bar: string;
}
class MyPropertyOutputReference {
  public get foo(): string {
    return "str"
  }

  public get bar(): string {
    return "str"
  }
}
class MyResource {
  // The return type if a setter is present is MyPropertyType, when it's absent it's MyPropertyOutputReference
  public get myProp() {
    return new MyPropertyOutputReference();  
  }

  public set myProp(value: MyPropertyType){}
}
```

## Decision 2: Output values are always present or not
Regarding output values, we are also unsure if we want to make only optional & computed types always present or all. Having only optional & computed ones always present could allow users to catch errors earlier, but would make setters harder.

```ts
// Access with always present type
nginxKubernetesDeployment.spec.selector.matchLabels

// Access with optional values
nginx.spec.selector ? nginx.spec.selector.matchLabels : {}
```

## Decision 3: Mutating primitives via set / put method 
If we want to handle these cases better we'd need to use putXYZ methods to mutate values, we are unsure here if we also want to use these for primitive types.

```ts
// This would in both cases be the way since it's a complex object
nginxKubernetesDeployment.putSpec({ selector: { matchLabels: { foo: "bar" } } });

// Only put
nginxKubernetesDeployment.spec.selector.putMatchLabels({ foo: "bar" })

// Put and set mixed
nginxKubernetesDeployment.spec.selector.matchLabels = { foo: "bar" }
```

### DIFF
--- a/packages/@cdktf/provider-generator/lib/get/__tests__/generator/__snapshots__/complex-computed-types.test.ts.snap
+++ b/packages/@cdktf/provider-generator/lib/get/__tests__/generator/__snapshots__/complex-computed-types.test.ts.snap
@@ -53,22 +53,22 @@ export namespace ACM {
   export class AcmCertificateDomainValidationOptions extends cdktf.ComplexComputedList {
 
     // domain_name - computed: true, optional: false, required: false
-    public get domainName() {
+    public get domainName(): string {
       return this.getStringAttribute('domain_name');
     }
 
     // resource_record_name - computed: true, optional: false, required: false
-    public get resourceRecordName() {
+    public get resourceRecordName(): string {
       return this.getStringAttribute('resource_record_name');
     }
 
     // resource_record_type - computed: true, optional: false, required: false
-    public get resourceRecordType() {
+    public get resourceRecordType(): string {
       return this.getStringAttribute('resource_record_type');
     }
 
     // resource_record_value - computed: true, optional: false, required: false
-    public get resourceRecordValue() {
+    public get resourceRecordValue(): string {
       return this.getStringAttribute('resource_record_value');
     }
   }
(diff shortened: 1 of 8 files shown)


jsteinich:
1. I definitely think that we should use a put method. I've run into issues with wanting different types many times. Fundamentally they are different types, but jsii won't allow them to be different (TS actually does now). We've also already introduced put methods and the benefits to better representing the types outweighs drawbacks to setting (especially since constructor is unchanged).
2. I believe HCL always allows for referencing a value. I haven't encountered any issues there, but I haven't extensively tested. Perhaps an `optional: true, computed: false` property would cause an issue; in which case, we could make the getter still return optional type. Anything that is computed or required will have a value.
3. I see the question being more about should we remove the setters universally (also for required properties). I have a slight preference towards keeping them, but having consistency is appealing.

jsteinich:
Looks like part of this was actually just a regression that is being fixed directly with #1310.

There are still some parts from #480 where this is applicable. We are already using an `<X>OutputReference` type for single item complex attributes, but there will most likely be some instances of non-single item complex attributes. Once #1310 is merged in I can take a shot at updating that (hopefully) last case.

DanielMSchmidt:
Thank you for your help! We are planning to brain storm a bit on our long-term vision of the API async during the next weeks, I think after that we will have a good enough vision that we can resolve the uncertainties around the decisions to be made in this PR.

PR #1234, opened 2021-11-01
chore(docs): move references to website

DanielMSchmidt:
I forgot to add this while writing the docs :)


website/docs/cdktf/concepts/providers-and-resources.html.md
```diff
@@ -264,6 +264,30 @@ export class HelloTerra extends TerraformStack {
 
 The [examples page](/docs/cdktf/examples.html) contains multiple example projects for every supported programming language.
 
+### Using References
+
+When working with resources created from providers you can use properties on the instances from the generated bindings to reference in other resources. This allows us to track logical dependencies and access computed values.
```
laurapacilio, line 269: Okay I have a couple questions before I rework this a bit :-) 

- Can you give a real-world example of when this is useful (beyond the example below)?
- Can we give an explanation of what's happening in the example below?
- How can folks find the properties - should they just check the provider documentation for available ones?

laurapacilio (changes requested):
Thank you for adding this!! A couple of questions for you before I make some suggested edits :-)

website/docs/cdktf/concepts/providers-and-resources.html.md
DanielMSchmidt: > Can you give a real-world example of when this is useful (beyond the example below)?

Every time they would normally write a reference in HCL, so if there is a value or logical dependency between two items. E.g. a new Role being created and being referenced by ARN in some assignment. A logical dependency would be in the Kubernetes space, resource there live in namespaces, so the namespace needs to be there before the resource. If I don't use a reference TF might want to provision resources before the namespace is there. When using a reference the order is clear, a dependency has been established

> Can we give an explanation of what's happening in the example below?

In the example we create a namespace and us its name as the namespace property in the Deployment. This creates an implicit dependency between the two and the value of the namespace's name is used in the Deployment. If we used generateName that generates a random name in kubernetes we could still reference the name and it's computed value would be used through the reference. I hope this sort of makes sense 🙈 


> How can folks find the properties - should they just check the provider documentation for available ones?

Yeah the provider documentation is the best place to look at

laurapacilio (changes requested):
More comments - but I think we're getting closer! Let me know what you think :)

website/docs/cdktf/concepts/providers-and-resources.html.md
DanielMSchmidt: Ah I think I see your confusion now. Metadata is a resource attribute / block, it has nothing to do with references. Using a property on a resource (e.g., `myResource.whateverProperty`) inside of the configuration of another construct is what creates the reference

website/docs/cdktf/concepts/providers-and-resources.html.md
```diff
@@ -264,6 +264,34 @@ export class HelloTerra extends TerraformStack {
 
 The [examples page](/docs/cdktf/examples.html) contains multiple example projects for every supported programming language.
 
+### References
+
+You can reference resource properties throughout your configuration. For example, you may want to use the name of a parent resource when assigning names to related child resources. Refer to your provider's documentation for a full list of available properties for each resource type.
+
+To create references, call `myResource.<propertyName>` on the resource instance. For example, you could use `myResource.name` to retrieve the `name` property from `myResource`. Terraform does not support passing an entire block (e.g. `exampleNamespace.metadata`) into a resource or data source, so you must create a reference for each individual property.
+
+References are also useful when you need to track logical dependencies. For example, Kubernetes resources live in a namespace, so a namespace must exist before Terraform can provision the associated resources. The TypeScript example below uses a reference for the namespace property in the the deployment. This reference tells Terraform that it needs to create the namespace before creating the resources.
```
laurapacilio, line 273: okay I like this a lot but now I'm feeling like people may think that if they DONT create a reference, Terraform won't know which order to create things in which isn't true...is it? I thought Terraform always knows about dependencies and creates resources accordingly. 

What am I missing here? Would we get an error if we didn't do this as a reference?

website/docs/cdktf/concepts/providers-and-resources.html.md
```diff
@@ -264,6 +264,34 @@ export class HelloTerra extends TerraformStack {
...
+  metadata: {
+    name: "nginx",
+    namespace: exampleNamespace.metadata.name, // This let's you reference the name
+    labels: {
+      app,
+    },
+  });
+```
+
+Please be aware that you can not pass an entire block (e.g. `exampleNamespace.metadata`) into a resource or data source as this is not supported by Terraform. You need to specify all values you want to pass.
```
laurapacilio, line 293: I _think_ we can delete this now, as we incorporated it into the paragraphs above (yes?). Let me know if you disagree :-)

laurapacilio (changes requested):
Thank you for your patience with me as we work on this :-) We're almost there!

website/docs/cdktf/concepts/providers-and-resources.html.md
DanielMSchmidt: It's true, if things are not connected through a reference it's not deterministic in which order the rest gets deployed. If one thing depends on another there needs to be a reference, otherwise it might error on deploy in some occasions (could work if you are lucky)

In HCL you use references all the time, the language is designed in a way that hints a bit more to that.

website/docs/cdktf/concepts/providers-and-resources.html.md
DanielMSchmidt: yes, I'll remove it

website/docs/cdktf/concepts/providers-and-resources.html.md
```diff
@@ -270,6 +270,32 @@ You can instantiate the same resource multiple times throughout your infrastruct
...
+const exampleNamespace = new Namespace(this, "tf-cdk-example", {
+  metadata: {
+    name: "tf-cdk-example",
+  },
+});
+
+new Deployment(this, "nginx-deployment", {
+  metadata: {
+    name: "nginx",
+    namespace: exampleNamespace.metadata.name, // This let's you reference the name
```
laurapacilio, line 292:
```suggestion
    namespace: exampleNamespace.metadata.name, // Reference the namespace name propery
```

laurapacilio (approved):
last nitpick and then I approve!! :)

PR #1235, opened 2021-11-01
test(cli): test generated providers

DanielMSchmidt:
This PR adds a couple of usage tests to the providers integration tests. These are aimed to find behaviour that should be working but is broken in regards to using the generated resources. We want to test mutation, references, and namespaces here.


test/test-helper.ts
```diff
@@ -211,3 +211,32 @@ export class TestDriver {
...
+
+    return constructs[id];
+  }
+
+  public output(id: string): string {
+    return this.stack.output[id].value;
+  }
+}
+
+export function $(stack: string): Query {
```
ansgarm, line 240: Getting nostalgic, aren't we 😁

test/test-helper.ts
```diff
@@ -211,3 +211,32 @@ export class TestDriver {
...
+  public byId(id: string): Record<string, any> {
+    const constructs = (
+      [
+        ...Object.values(this.stack.resource || {}),
+        ...Object.values(this.stack.data || {}),
+      ] as Record<string, any>[]
+    ).reduce(
+      (carry, item) => ({ ...carry, ...item }),
+      {} as Record<string, any>
+    );
```
ansgarm, line 230: Maybe add a comment that this "jumps over" the resource type and combines all resources by id, ignoring their type? I had a hard time figuring out how this worked using `Object.values()` 😅 An alternative could be adding stronger types, which I'd probably go for.
I'm thinking about something like (with better names 😄):
```ts
type RenderedTerraformResource = Record<string, any>;
type TerraformProvider = string;
type Id = string;
type RenderedStack = {
  resource: Record<TerraformProvider, Record<Id, RenderedTerraformResource>>,
  data: Record<TerraformProvider, Record<Id, RenderedTerraformResource>>,
  output: Record<Id, {value: any}>
};
const constructs: RenderedTerraformResource[];
```
(This is probably too verbose 😂 But I hope I could convey what I meant)

test/test-helper.ts
```diff
@@ -211,3 +211,32 @@ export class TestDriver {
 
 export const onWindows = process.platform === "win32" ? it : it.skip;
 export const onPosix = process.platform !== "win32" ? it : it.skip;
+
+class Query {
```
ansgarm, line 215: Just an idea: You could also add a `toString` method (which returns the stack string value) and always return a Query from `driver.synthesizedStack()` calls – that could remove the need for `$`. Would be a breaking change however (type-wise at least)

test/typescript/providers/test.ts
```diff
@@ -9,6 +9,115 @@ describe("full integration test", () => {
...
+        });
+
+        test("put method mutation", () => {
+          expect(instance.metadata_options).toEqual({
+            http_endpoint: "127.0.0.1",
+          });
+        });
+      });
+
+      describe.skip("single-item nesting", () => {
```
ansgarm, line 90: skipped

test/test-helper.ts
DanielMSchmidt: That's a great idea!

test/typescript/providers/test.ts
DanielMSchmidt: Yeah, it's not supported, I'll add a comment 👍

test/test-helper.ts
```diff
@@ -8,6 +8,39 @@ const path = require("path");
 const fs = require("fs");
 const fse = require("fs-extra");
 
+class Query {
```
ansgarm, line 11: Maybe now rename to `QueryableStack` or `TestStack` or similar as this is not the result of a `$()` call anymore?

test/typescript/providers/test.ts
```diff
@@ -9,6 +9,114 @@ describe("full integration test", () => {
...
+        });
+
+        test("put method mutation", () => {
+          expect(instance.metadata_options).toEqual({
+            http_endpoint: "127.0.0.1",
+          });
+        });
+      });
+
+      // The necessary functions are exposed, but currently do not work
```
ansgarm, line 90: maybe link to an issue? I wouldn't know why they don't work or where to look whether this comment is still valid.

ansgarm (approved):

ansgarm (dismissed):
Just came to my mind: A draft for release notes would be nice. Some that mention `.toString()` being required when using Snapshot tests.

ansgarm (approved):

PR #1245, opened 2021-11-03
chore(docs): Fix docs nits like images, headings, and contractions

laurapacilio:
Fixing a few images on the architecture page + a few little nits here and there like inconsistent headings and note format.


schersh (approved):
Looks good to me, thanks!

Issue #1203, opened 2021-10-25
Exported APIs cannot use un-exported type "@pahud/cdktf-aws-eks.EKS.EksCluster" 


skorfmann:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### cdktf & Language Versions

- 0.7

### Affected Resource(s)

- aws provider

### References

> Hi  @skorfmann 
> 
> I ran into this error when I upgrade cdktf 0.6.0 to 0.7.0 from this  https://github.com/pahud/cdktf-aws-eks/pull/61
> 
> <img width="1783" alt="圖片" src="https://user-images.githubusercontent.com/278432/138653293-0f021b79-2d3d-4447-8b45-fa5d9ca94207.png">
> 
> The complete CI log https://github.com/pahud/cdktf-aws-eks/runs/3993893529?check_suite_focus=true
> 
> Looks like the `EksCluster` is an exported class but somehow it complained not an exported class:
> https://github.com/hashicorp/cdktf-provider-aws/blob/2c441c164977d7124602468bc5232dee7e03e8a9/src/EKS.ts#L659
> 
> Do you have any idea about this error?

See https://github.com/hashicorp/cdktf-provider-aws/issues/465 for more details

skorfmann:
@pahud could you give the latest pre-release `0.8.0-pre.1` a try? it should solve the issue you saw around types

ansgarm:
@skorfmann I think we need to first release the aws pre-built provider using cdktf 0.8 😅 Otherwise a lot of local linking (and building that pre-built provider locally) is needed.

ansgarm:
Seems to work, I created a [PR](https://github.com/pahud/cdktf-aws-eks/pull/63) that updates to `v0.8.1`.

ansgarm:
I'm going to close this issue as the PR I posted compiles without the error originally described in this issue. Feel free to open a new issue if any further errors might come up 👍

Issue #1170, opened 2021-10-19
Enable Barrel Imports for Namespaced Providers


skorfmann:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

With `0.7` the AWS provider is generated in a namespaced fashion, where all resources of a service are part of one namespace. 

```ts
import { LambdaFunction } from './.gen/providers/aws'

new LambdaFunction.LambdaFunction(this, 'foo', {
      functionName: 'f',
      role: ''
    })
```

It would be very handy to enable barrel imports as well

```ts
import { LambdaFunction } from './.gen/providers/aws/LambdaFunction'

new LambdaFunction(this, 'foo', {
      functionName: 'f',
      role: ''
    })
```

### References

The [aws-cdk-lib](https://github.com/aws/aws-cdk/tree/master/packages/aws-cdk-lib#barrel-import) is doing something similar.

skorfmann:
Fixed with https://github.com/hashicorp/terraform-cdk/pull/1248

PR #1248, opened 2021-11-03
fix(lib): Rework AWS Provider Namespacing

skorfmann:
This changes the provider generation from explicit namespaces in a single file, to implicit namespaces when exporting from a subfolder.

### Before

```ts
// file providers/aws/index.ts
export * from './CloudFront'
// ... a bunch of other namespace exports

// file providers/aws/CloudfFront.ts
export namespace CloudFront {
// all resource / data source classes in this namespace
}
```

### After

```ts
// file providers/aws/index.ts
export * as cloudfront from './cloudfront'
// ... a bunch of other namespace exports

// file providers/aws/cloudfront/cloudfront-distribution.ts
export class CloudfrontDistribution {
// single resource class
}

// file providers/aws/cloudfront/index.ts
export * from './cloudfront-distribution'
}
```

### Benefits 

- Allows direct imports `import { CloudfrontDistribution } from '.gen/providers/aws/cloudfront`
- Allows namespace import `import { cloudfront } from '.gen/providers/aws`
- Should have less issues with jsii, since that's pretty much how the AWS CDK does it at the moment
- By splitting the resources to dedicated files again, we work around the Typescript < 4 issues where > 1200 exports per file throw an error 
- This should address issues like [this](https://github.com/hashicorp/terraform-cdk/issues/1203)

The only slightly annoying thing is yet another breaking change around the imports for the AWS provider. This is due to jsii expecting the namespace to be lowercase. However, I'm rather confident that the naming should be more or less stable from thereon.

We have yet to see how this behaves in CI in terms of memory consumption. Locally, I haven't seen any noticeable spikes yet.

Created a separate [PR](https://github.com/hashicorp/terraform-cdk/pull/1249) to update the provider snapshot for better readability in this PR (in honour of the infinity PR 😄 )

### References

- https://github.com/hashicorp/terraform-cdk/issues/1203
- https://github.com/microsoft/TypeScript/issues/38691 (even though it's closed, I still can reproduce it in the latest ts < 4 version)
- https://github.com/aws/jsii/issues/3095
- #1170


skorfmann:
Here we go again :/ 

```
$ cdktf get
Generated typescript constructs in the output directory: .gen
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
lerna ERR! yarn run build stderr:
/__w/terraform-cdk/terraform-cdk/node_modules/typescript/lib/tsc.js:82931
                throw e;
                ^

RangeError: Maximum call stack size exceeded
```

Gonna look into this tomorrow - wondering if there's really a single resource file which exports more than ~ 1200 types

skorfmann:
It's the [wafv2_web_acl](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/wafv2_web_acl) with over 7000 (!) exports. Not sure if there are other resources which would cause the same problem.

skorfmann:
> It's the [wafv2_web_acl](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/wafv2_web_acl) with over 7000 (!) exports. Not sure if there are other resources which would cause the same problem.

Gonna look into either deduplication or export sharding across multiple files tomorrow.

skorfmann:
For reference, that's the provider schema of the `aws_wafv2_web_acl` resource https://gist.github.com/skorfmann/71865435f2b62b8f3279a7e07cc35fea with more than 40k lines of JSON

skorfmann:
Also, I've checked a few other resources and they are usually way below 100 exports. But anyway, gonna implement some sharding for files with more than 1000 exports

packages/@cdktf/provider-generator/lib/get/generator/provider-generator.ts
```diff
@@ -161,44 +166,59 @@ export class TerraformProviderGenerator {
...
-    this.code.line(`/**`);
-    this.code.line(`* ${comment}`);
-    this.code.line(`*/`);
-    this.code.openBlock(`export namespace ${namespace}`);
+
+    if (!ns?.name) throw new Error("namespace name is missing");
+
+    const generatedFiles = [];
+
+    const lowerCasedNamespace = toSnakeCase(ns.name, "");
```
ansgarm, line 179: Does this produce weird names like for example `f_sx`, `io_t` and `e_c2`? I encountered such behaviour when debugging (and using `toSnakeCase`).

Maybe it might be better to change `ns.name` to the snake case version [in the source](https://github.com/hashicorp/terraform-cdk/blob/3058eb7e1c397f8fb76a5a72ac48b4b39eb162ab/packages/%40cdktf/provider-generator/lib/get/generator/constants/provider-namespaces.ts#L193)?

skorfmann:
> Also, I've checked a few other resources and they are usually way below 100 exports. But anyway, gonna implement some sharding for files with more than 1000 exports

Have a rough prototype working locally which compiles with TS 3.9 and jsii can build it. Fun fact: The jsii assembly just for this resource is 19 MB 🤯

skorfmann:
This is still WIP, some cleanup to do. However, the provider for the `wafv2-web-acl` resource compiles properly with TS 3.9.10 - The existing code generation is currently broken, gonna fix that tomorrow.

packages/@cdktf/provider-generator/lib/get/generator/provider-generator.ts
skorfmann: Yes, good point - updated it in bd82746

skorfmann:
Still have to fix a Java and c# example, but other than that it's starting to look good.

ansgarm (commented):
I think drafting upgrade instructions (i.e. `EC2 -> ec2`, `data_sources -> datasources`) would be useful. I'd add them to an "unreleased" section in `CHANGELOG.md`.

skorfmann:
> I think drafting upgrade instructions (i.e. EC2 -> ec2, data_sources -> datasources) would be useful. I'd add them to an "unreleased" section in CHANGELOG.md.

See a3c60f0

ansgarm (approved):

PR #1252, opened 2021-11-04
fix(provider-generator): escape hyphens in provider names for Go package identifiers

### LINKED ISSUE #1250: golang: how to include providers with dashes?
### Community Note

### cdktf & Language Versions

cdktf: 0.6.4

### Affected Resource(s)

google-beta.*

### Debug Output

```
generated/hashicorp/google-beta/google-beta.go:2:15: syntax error: unexpected -, expecting semicolon or newline
cdktf encountered an error while synthesizing
```

### Expected Behavior

google-beta should be converted to google_beta or googlebeta? or a way to alias providers should exist?

### Actual Behavior

google-beta.go starts with

```
package google-beta
```

which is invalid go

### Steps to Reproduce
Run `cdktf init --template="go" --local`

Add the following to cdktf.json
```json
  "terraformProviders": [
    "hashicorp/google@~> 3.90",
    "hashicorp/google-beta@~> 3.90",
  ],
```

Run `cdktf get`

Run `cdktf synth`

ansgarm:
resolves #1250


DanielMSchmidt (approved):

packages/@cdktf/provider-generator/lib/get/__tests__/constructs-maker.test.ts
```diff
@@ -79,4 +83,18 @@ describe("constructsMaker", () => {
...
+      const target = new ConstructsMakerProviderTarget(
+        {
+          name: "google-beta",
+          fqn: "google-beta",
+          source: "google-beta",
+          version: "~> 4.0",
+        },
+        Language.GO
+      );
+      expect(target.srcMakName).toEqual("google_beta");
```
figadore, line 97: From the Go documentation

> Good package names are short and clear. They are lower case, with no under_scores or mixedCaps

I  wasn't sure how important the underscores part was, whether it would be better to convert `google-beta` to `google_beta` or `googlebeta`. Maybe there are already conventions for this elsewhere within cdk?

ansgarm: Good catch!
Looks like it might make sense to remove any hyphens instead of replacing them. The only reason in favor of underscores would be to prevent collisions between names like e.g. `google-beta` and `googlebeta` 🤔

skorfmann: > to prevent collisions between names like e.g. google-beta and googlebeta

Seems like a good point for underscores. 

Checked the AWS [docs](https://docs.aws.amazon.com/cdk/latest/guide/work-with-cdk-go.html) and they're going with something like `github.com/aws/aws-cdk-go/awscdk/awss3` where the typescript version is something like `awscdk/aws-s3`. However, they have a known set of resources which we don't have.

What would be the workaround if there would be a name collision?

figadore: I'm not sure where the cdktf.json schema is documented, but I tried a few variations of the block syntax
```
    "terraformProviders": [
      {
        "name": "google-beta",
        "source": "google-beta",
        "version": "4.0.0"
      }
    ],
```

and nothing helped with collisions. I hoped that something similar to terraform's [alias field](https://www.terraform.io/docs/language/providers/configuration.html#alias-multiple-provider-configurations) would work, since collisions need to be handled somehow (e.g. how do I use both `Runelab/bitbucket` and `aeirola/bitbucket`?), but maybe this is a separate issue.

ansgarm: Yes, sounds like a separate issue to file :) But still something we should support 👍
Going to resolve this discussion and keep the `_` in the Go package then.

ansgarm: Follow-up issue: https://github.com/hashicorp/terraform-cdk/issues/1258

PR #1253, opened 2021-11-04
feat(cli): use ts-node to compile main.ts

### LINKED ISSUE #1012: Stale javascript from typescript compilation
### Community Note

### cdktf & Language Versions

```
❯ yarn list | grep -e cdk -e typescript
├─ @cdktf/provider-google@0.3.7
├─ cdktf@0.5.0
├─ typescript@4.4.3
```

### Affected Resource(s)

Any

### Debug Output

N/A

### Expected Behavior

Any change to `*.ts` should be reflected in the next deploy.

### Actual Behavior

Usually it is, but not always.

### Steps to Reproduce

Still unclear. Tends to happen after a number of edits and deploys. I tend to run this to fix it.

```
./node_modules/.bin/tsc --build --clean
```

### Important Factoids

Another way I've found to work around this is to add `ts-node` as a dependency and update my `cdktf.json` to have `"app": "npx ts-node main.ts",`

### References

None

---

I'm posting this partly for discussion to see if others have encountered this. 

Using ts-node directly seems to work well so far and avoids having `.d.js,.js` files peppered around my working directory.

Would it make sense to maybe have the default be to run via ts-node?

DanielMSchmidt:
This should help with stale changes being deployed in certain scenarios, Closes #1012


examples/typescript/ucloud/package.json
```diff
@@ -27,6 +27,7 @@
     "@types/node": "^14.0.26",
     "cdktf-cli": "0.0.0",
     "jest": "27.1.0",
+    "ts-node": "10.4.0",
```
skorfmann, line 30: why this specific version?

DanielMSchmidt: Because it's the latest version, using the latest tag was a bit too risky for me

skorfmann: I'm wondering if this will ever be updated when it's fixed. It might also suggest to users that there's a reason why it's a fixed version, even though it's just there since it happened to be the current version at some point.

DanielMSchmidt: Would adding a ^ solve your concern?

skorfmann: Yes, I think that would be good 👍

DanielMSchmidt: 👍

skorfmann (approved):
Cool, thanks!

Issue #561, opened 2021-02-11
Package Pre Release Versioning


skorfmann:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Each commit on the `main` branch gets automatically released as pre-release version to all targets: npm / pypi / nuget / maven

Currently, we take the current version number (e.g. `0.1.0` as of today) and append a `pre-<BUILD_NUMBER` in all cases but pypi, where it's `dev.<BUILD_NUMBER>`. This results in something like `0.1.0-pre-178`, which is a bit misleading from my point of view. I think I'd expect the pre-release to indicate the next version. So, something like `0.2.0-pre-178`.

### Overview

| Target | release version | current pre-release | actually expected pre-release      |
|--------|-----------------|---------------------|---------------|
| [NPM](https://www.npmjs.com/package/cdktf)    | 0.1.0           | 0.1.0-pre.188       | 0.2.0-pre.188 |
| [pypi](https://pypi.org/project/cdktf/0.1.0/#history)   | 0.1.0           | 0.1.0.dev188        | 0.2.0.dev188  |
| [nuget](https://www.nuget.org/packages/HashiCorp.Cdktf)  | 0.1.0           | 0.1.0-pre.188       | 0.2.0-pre.188  |
| [maven](https://search.maven.org/artifact/com.hashicorp/cdktf)  | 0.1.0           | 0.1.0-pre.188       | 0.2.0-pre.188  |

### Goals

The goals I see:

- A pre-release version should indicate that it's a snapshot of unreleased contents of the `main` branch
- A pre-release version should track the next version compared to the current "official" release
- For the normal workflow, I'd expect the next minor version as pre-release target. I think we can ignore patch versions for next releases. 
- Major version pre-releases will be handled manually somehow (we'll deal with that when we have the problem)
- Package managers should ignore pre-releases by default and only install them by explicit request
[...]

PR #1257, opened 2021-11-05
chore(release): Pre-releases should use next version

jsteinich:
This takes care of the main part of #561. Also includes some tools to help with #656.

Using [standard-version](https://github.com/conventional-changelog/standard-version) to handle bumping the version number in a manner similar to semantic-release but without the release.

Can run `yarn prepare-release` or `yarn prepare-release --release-as minor` when preparing for an actual release to bump the version and generate the changelog.

Follow up:
We also discussed changing the qualifier. For at least Java that would probably be a good idea; however, there isn't an easy way to change for just 1 language. Probably makes sense to hold off on any changes until we get actual user feedback requesting changes.


skorfmann:
Awesome, thanks! So, this would work like this when merging the PR?

| Target | current release version | pre-release | 
|--------|-----------------|---------------------|
| [NPM](https://www.npmjs.com/package/cdktf)    | 0.7.0           | 0.8.0-pre.500       |
| [pypi](https://pypi.org/project/cdktf/0.1.0/#history)   | 0.7.0           | 0.8.0.dev500      |
| [nuget](https://www.nuget.org/packages/HashiCorp.Cdktf)  | 0.7.0           | 0.8.0-pre.500 |
| [maven](https://search.maven.org/artifact/com.hashicorp/cdktf)  | 0.7.0           | 0.8.0-pre.500 |

jsteinich:
> this would work like this when merging the PR?

Yep.
It increments the pre-release version each time rather than using the github build number, so technically it will start with `0.8.0-pre.0`, but we should have a tag to be able to track the exact commit if needed.

jsteinich:
I've been doing some testing here: https://github.com/jsteinich/standard-version-test

I think that I've run into https://github.com/conventional-changelog/standard-version/issues/724 as everything worked for the 1st pre-release, but failed on the next one.

jsteinich:
I did some more investigation and managed to get things working.

I'm not going to update the release instructions quite yet since the release changelog generation hasn't been fully tested yet. I figure we can try it out with the next release and compare to the existing script. Since that is a secondary goal, if it doesn't work we can just not use it and continue iterating or remove.

skorfmann:
Seems to work as expected:

```
✗ yarn prepare-next-release
yarn run v1.22.5
$ standard-version --prerelease pre --release-as minor --skip.changelog=true --skip.commit=true
✔ bumping version in package.json from 0.7.0 to 0.8.0-pre.1
✔ bumping version in packages/cdktf-cli/package.json from 0.0.0 to 0.8.0-pre.1
✔ bumping version in packages/cdktf/package.json from 0.0.0 to 0.8.0-pre.1
✔ bumping version in packages/@cdktf/hcl2cdk/package.json from 0.0.0 to 0.8.0-pre.1
✔ bumping version in packages/@cdktf/hcl2json/package.json from 0.0.0 to 0.8.0-pre.1
✔ bumping version in packages/@cdktf/provider-generator/package.json from 0.0.0 to 0.8.0-pre.1
✔ tagging release v0.8.0-pre.1
ℹ Run `git push --follow-tags origin pr/jsteinich/1257 && npm publish --tag pre` to publish
```

Is there anything else missing or would you consider this as good to go @jsteinich ?

jsteinich:
> Is there anything else missing or would you consider this as good to go

@skorfmann I think it is good to go.

Issue #1264, opened 2021-11-05
Generating azurerm provider for go exceeds 'hard' memory limit


waxb:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### cdktf & Language Versions

```
waxb@mbp:cdktf-azure-go$ cdktf --version
0.7.0
waxb@mbp:cdktf-azure-go$ cat cdktf.json 
{
    "language": "go",
    "app": "go run main.go",
    "codeMakerOutput": "generated",
    "terraformProviders": ["hashicorp/azurerm@~> 2.74"],
    "terraformModules": [],
    "context": {
        "excludeStackIdFromLogicalIds": "true",
	"allowSepCharsInLogicalIds": "true"
    }
}
waxb@mbp:cdktf-azure-go$ go version
go version go1.17.2 darwin/amd64
```

### Expected Behavior
```
waxb@mbp:cdktf-azure-go$ cdktf get
⠧ downloading and generating modules and providers...
Generated go constructs in the output directory: generated

The generated code depends on jsii-runtime-go. If you haven't yet installed it, you can run go mod tidy to automatically install it.
waxb@mbp:cdktf-azure-go$ 
```
### Actual Behavior

https://gist.github.com/waxb/78ecff27cd265525e290043da7db6c71

```
  | <--- Last few GCs --->
  | [49924:0x7f8761900000]   197321 ms: Mark-sweep 10085.5 (10275.2) -> 10071.7 (10276.9) MB, 12795.4 / 0.0 ms  (average mu = 0.142, current mu = 0.002) allocation failure scavenge might not succeed
  | [49924:0x7f8761900000]   210142 ms: Mark-sweep 10087.5 (10276.9) -> 10073.8 (10279.2) MB, 12795.8 / 0.0 ms  (average mu = 0.077, current mu = 0.002) allocation failure scavenge might not succeed
  | <--- JS stacktrace --->
  | FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
```

`cdktf get` fails as the azurerm provider generation is bigger than the default
[...]

PR #1265, opened 2021-11-05
fix(lib): Increased --max-old-space-size value

waxb:
Azurerm provider generation for Go [fails](https://github.com/hashicorp/terraform-cdk/issues/1264) because the 'hard' limit is 10240 MB and azurerm provider needs ~13 GBs of memory to get generated.

⚠️ Note: This is just further delaying a fix of a workaround rather than a solution.


ansgarm:
(fyi: I changed your commit message to match the enforced schema for single commit PRs)

ansgarm (approved):
Thank you @waxb!
Sidenote: I checked locally what happens if we specify more memory than the machine has to offer and Node.js does not seem to check that – So this change still allows machines with less than e.g. 16 GB memory to run `cdktf get`. Which (as I now think about it) already happened in Github Actions as they offer less than 10 GB of memory 😄

PR #1278, opened 2021-11-08
chore(release): Revert "Merge pull request #1277 from hashicorp/rever…

skorfmann:
This reverts commit 7838993b88a3c2d96a7b8a2815bd0dbc0f14489b, reversing
changes made to c0a13c56b50f0925714146df2548100f6aaea789.

See 

- https://github.com/hashicorp/terraform-cdk/pull/1277
- https://github.com/hashicorp/terraform-cdk/pull/1257

It turned out that the `standard-version` command is only taking the `version` field in `package.json` into account. `lerna version` is updating the cross dependencies between the packages in `dependencies` / `devDependencies` / `peerDependencies`.

Had a brief look at the options of [standard-version](https://github.com/conventional-changelog/standard-version) and it seems to be tailored to npm. 

I'm wondering if we could achieve the same with [lerna version](https://github.com/lerna/lerna/blob/main/commands/version/README.md#options) but also open for other fixes.

### DIFF
--- a/.github/workflows/release_next.yml
+++ b/.github/workflows/release_next.yml
@@ -20,9 +20,15 @@ jobs:
       CHECKPOINT_DISABLE: "1"
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0 # gives standard-version access to all previous commits
       - name: installing dependencies
         run: yarn install
-      - run: tools/align-version.sh "-pre.${{ github.run_number }}"
+      - name: Configure git user
+        run: |
+          git config user.name github-team-tf-cdk
+          git config user.email github-team-tf-cdk@hashicorp.com
+      - run: yarn prepare-next-release
       - run: yarn build
       - run: yarn test
       - run: yarn package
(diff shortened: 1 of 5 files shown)


jsteinich:
Sorry about this. I made sure to check that the package version was updating correctly for all the sub-packages, but I didn't think to check the dependencies.

I pushed a commit that goes back to relying on `tools/align-version.sh` which were using before and that uses `lerna version` internally.

It doesn't look like just `lerna version` could do everything that we want, but the other idea I had was just to have a `vNext` version that we update manually or part of a release script.

skorfmann:
> Sorry about this. I made sure to check that the package version was updating correctly for all the sub-packages, but I didn't think to check the dependencies.

No worries, that's exactly what I missed as well when trying this locally 🤷

skorfmann:
> I pushed a commit that goes back to relying on `tools/align-version.sh` which were using before and that uses `lerna version` internally.

Ok, let's go with that for now.

Issue #1152, opened 2021-10-17
Python cdktf imports throwing error


raghavmri:
I am running `cdktf get` and all the modules are being loaded onto my imports folder but after that, if I try to run my app I get the following error

> Process SpawnProcess-1:
> Traceback (most recent call last):
>   File "/usr/local/lib/python3.9/multiprocessing/process.py", line 315, in _bootstrap
>     self.run()
>   File "/usr/local/lib/python3.9/multiprocessing/process.py", line 108, in run
>     self._target(*self._args, **self._kwargs)
>   File "/home/mriblog9/gogamic-managed-cloud/backend/api-python/env/lib/python3.9/site-packages/uvicorn/subprocess.py", line 76, in subprocess_started
>     target(sockets=sockets)
>   File "/home/mriblog9/gogamic-managed-cloud/backend/api-python/env/lib/python3.9/site-packages/uvicorn/server.py", line 68, in run
>     return asyncio.run(self.serve(sockets=sockets))
>   File "/usr/local/lib/python3.9/asyncio/runners.py", line 44, in run
>     return loop.run_until_complete(main)
>   File "/usr/local/lib/python3.9/asyncio/base_events.py", line 642, in run_until_complete
>     return future.result()
>   File "/home/mriblog9/gogamic-managed-cloud/backend/api-python/env/lib/python3.9/site-packages/uvicorn/server.py", line 76, in serve
>     config.load()
>   File "/home/mriblog9/gogamic-managed-cloud/backend/api-python/env/lib/python3.9/site-packages/uvicorn/config.py", line 448, in load
>     self.loaded_app = import_from_string(self.app)
>   File "/home/mriblog9/gogamic-managed-cloud/backend/api-python/env/lib/python3.9/site-packages/uvicorn/importer.py", line 21, in import_from_string
>     module = importlib.import_module(module_str)
>   File "/usr/local/lib/python3.9/importlib/__init__.py", line 127, in import_module
>     return _bootstrap._gcd_import(name[level:], package, level)
>   File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
>   File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
>   File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
>   File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
>   File "<frozen importlib._bootstrap_external>", line 850, in exec_module
[...]

PR #1288, opened 2021-11-11
chore(deps): relax jsii dependencies

skorfmann:
With the 0.7 release we fixed jsii to `1.37.0` because of https://github.com/hashicorp/terraform-cdk/issues/1152 and https://github.com/aws/jsii/issues/3078

This pull request aims to relax that restriction again. I'm not exactly sure if our changes to the namespacing fixed the issues already or if we have to push for https://github.com/aws/jsii/issues/3078 - Tests will tell :)


skorfmann:
Looked a bit into aws/jsii#3078 and yes, it's something which we should somehow address upstream.

I see two options, both around the way module imports are generated in jsii.

1. what @jsteinich posted already, something like `from ..aws import access_analyzer`
2. create fully qualifed import, in this case `import imports.aws.accessanalyzer`

There's no clean way to just configure this on our end, so we'll work on this upstream.

skorfmann:
The upstream fix in https://github.com/aws/jsii/pull/3181 got merged and is likely gonna be released early next week

ansgarm:
This one was superseded by https://github.com/hashicorp/terraform-cdk/pull/1336, wasn't it?

Issue #960, opened 2021-09-08
csharp: Amazon.JSII.Runtime.JsiiException Expected a boolean, got "a" (string)


rgl:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### cdktf & Language Versions

cdktf 0.5.0
csharp 3.1/5.0

### Affected Resource(s)

vsphere_virtual_machine

### Expected Behavior

It should have used a bool.

### Actual Behavior

It converted the bool to a string.

### Steps to Reproduce

Using the code from https://github.com/rgl/dotnet-terraform-cdk-vsphere-ubuntu-example/blob/main/Main.cs#L134-L141 which is something alike;

```csharp
            new VirtualMachine(this, name, new VirtualMachineConfig
            {
                Disk = new[]
                {
                    new VirtualMachineDisk
                    {
                        UnitNumber = 0,
                        Label = "os",
                        Size = Math.Max(ubuntuTemplate.Disks("0").Size, vmDiskOsSize),
                        EagerlyScrub = ubuntuTemplate.Disks("0").EagerlyScrub,
                        ThinProvisioned = ubuntuTemplate.Disks("0").ThinProvisioned,
                    },
```

Execute the stack with `cdktf diff`.

This is the error I get:

```console
vagrant@ubuntu:~/dotnet-terraform-cdk-vsphere-ubuntu-example$ npx cdktf diff⠸ synthesizing...
[2021-09-08T18:25:11.944] [ERROR] default - Unhandled exception.
[2021-09-08T18:25:11.946] [ERROR] default -  
[2021-09-08T18:25:11.961] [ERROR] default - Amazon.JSII.Runtime.JsiiException: Expected a boolean, got "${data.vsphere_virtual_machine.ubuntu_template.disks.0.eagerly_scrub}" (string)
   at Amazon.JSII.Runtime.Services.Client.TryDeserialize[TResponse](String responseJson)
   at Amazon.JSII.Runtime.Services.Client.ReceiveResponse[TResponse]()
   at Amazon.JSII.Runtime.Services.Client.Send[TRequest,TResponse](TRequest requestObject)
[...]

PR #1299, opened 2021-11-14
chore(lib): Standardize IResolvable usage

jsteinich:
The goal of this PR is to establish consistent usage of `IReolvable` when working with attributes of Terraform elements by looking to the AWS CDK for guidance and making improvements where appropriate.

Work already done:

- Update TerraformLocal to use refs similar to other types
- Split interpolation and getString within TerraformModule for future refinements and possible manual escapes.
- Stop casting booleans. Will fix additional cases of #960 

IResolvable usage:
| TF Type                 | CDKTF CURRENT                             | AWS CDK                                                          | CDKTF PROPOSED                             |
|-------------------------|-------------------------------------------|------------------------------------------------------------------|--------------------------------------------|
| string                  | string                                    | string                                                           | string                                     |
| number                  | number                                    | number                                                           | number                                     |
| boolean                 | boolean \| IResolvable                    | boolean \| IResolvable                                           | boolean \| IResolvable                     |
| object                  | object                                    | object \| IResolvable                                            | object                                     |
| string[]                | string[]                                  | string[]                                                         | string[]                                   |
| number[]                | number[]                                  | number[] \| IResolvable                                          | number[]                                   |
| boolean[]               | boolean[]                                 | Array<boolean \| IResolvable> \| IResolvable                     | Array<boolean \| IResolvable> \| IResolvable                                   |
| object[]                | object[]                                  | Array<object \| IResolvable> \| IResolvable                      | object[] \| IResolvable                                   |
| Record<string, string>  | { [key: string]: string } \| IResolvable  | { [key: string]: (string) } \| IResolvable                       | { [key: string]: string }                  |
| Record<string, number>  | { [key: string]: number } \| IResolvable  | { [key: string]: (number) } \| IResolvable                       | { [key: string]: number }                  |
| Record<string, boolean> | { [key: string]: boolean } \| IResolvable | { [key: string]: (boolean \| IResolvable) } \| IResolvable       | { [key: string]: (boolean \| IResolvable) }                 |
| Record<string, object>  | { [key: string]: object } \| IResolvable  | { [key: string]: (object \| IResolvable) } \| IResolvable        | { [key: string]: object } \| IResolvable                  |
| Set<string>             | string[]                                  | string[]                                                         | string[]                                   |
| Set<number>             | number[]                                  | number[] \| IResolvable                                          | number[]                                   |
| Set<boolean>            | boolean[]                                 | Array<boolean \| IResolvable> \| IResolvable                     | Array<boolean \| IResolvable> \| IResolvable                                  |
| Set<object>             | object[]                                  | Array<object \| IResolvable> \| IResolvable                      | object[] \| IResolvable                                   |
| dynamic                 | { [key: string]: any }                    | { [key: string]: (any \| IResolvable) } \| IResolvable           | { [key: string]: any }                     |
| list<map<string>>       | may be broken                             | Array<{ [key: string]: (string) } \| IResolvable> \| IResolvable | { [key: string]: string }[] \| IResolvable |
| set<map<string>>        | may be broken                             | Array<{ [key: string]: (string) } \| IResolvable> \| IResolvable | { [key: string]: string }[] \| IResolvable |
| list<list<string>>      | may be broken                             | string[][] \| IResolvable                                        | broken                  |
| list<set<string>>       | may be broken                             | string[][] \| IResolvable                                        | broken                  |
| map<list<string>        | may be broken                             | { [key: string]: (string[]) } \| IResolvable                     | broken               |

The general rule is include `IResolvable` when no token type is available and when not using custom structs.

Changes from AWS CDK:

- ~~Don't use `IResolvable` for custom structs (labeled as 'object'). Since Terraform doesn't allow passing entire blocks, I'm not sure why we would need these to be resolvable~~ ---Use `IResolvable` when only an interface is generated
- Create a token type for `number[]`. They are more common in Terraform and can combine `string[]` and `number` token techniques
- ~~Create a token type for `boolean[]`. This one is debatable, but a token could be created using binary encoding~~
- Create a token type for `Record<string, *>`. Can use a special string token key in the map to indicate the entire map is a token. Only make for core types and expand others to use `IResolvable` similar to AWS CDK since jsii requires exact typing.

Other areas of consideration:

- Should `count` on resources being just a number? Probably yes since we can use a token and number translates to other languages better.
- Should functions use `IResolvable`? The AWS CDK would say no as it hurts other language support. Since chaining functions is more common in Terraform, we probably want to use it. Might be nice to make available for all, but could also do case by case analysis of which functions are likely to be chained. --- Match with other rules as much as possible.


packages/cdktf/lib/terraform-local.ts
```diff
@@ -22,31 +22,31 @@ export class TerraformLocal
...
   }
 
   private interpolation(): any {
-    return `local.${this.friendlyUniqueId}`;
+    return ref(`local.${this.friendlyUniqueId}`);
   }
 
   public get fqn() {
-    return `local.${this.friendlyUniqueId}`;
+    return Token.asString(`local.${this.friendlyUniqueId}`);
```
DanielMSchmidt, line 49: I wonder if we should wrap this with ref, otherwise there could be an extra "${}" if this is used within a Terraform function

jsteinich: I could see us wanting to have some more information when working on cross-stack references, but at this time there is no `$` present in the string, so I don't there will be any problems there.

DanielMSchmidt: Oh yeah, I missed that, makes sense 👍

jsteinich:
Thanks for taking a look @DanielMSchmidt. @skorfmann @ansgarm any thoughts on the proposed changes? This may not make the next release, but I'd definitely like to see it in the following one.

ansgarm:
Thanks for the draft, @jsteinich 👍

> Should `count` on resources being just a number? Probably yes since we can use a token and number translates to other languages better.

The AWS adapter uses the output of a Terraform function here:
https://github.com/hashicorp/cdktf-aws-cdk/blob/c76da087c38d55551f73faa8934e4457f778dfc8/src/aws-adapter.ts#L165-L169
However (as you noted) this output could be wrapped in a Token.

> Create a token type for boolean[]. This one is debatable, but a token could be created using binary encoding

Hehe, nice idea. But, yeah, debatable 😄

Maybe there are other specifics of the current implementation used in the AWS adapter, but I've not found further ones and I could always resort to overrides should there be some 😄

packages/@cdktf/provider-generator/lib/get/generator/emitter/attributes-emitter.ts
```diff
@@ -295,6 +298,13 @@ export class AttributesEmitter {
           )}ToTerraform(${varReference}.internalValue),`
         );
         break;
+      case type.isComplex && !type.struct?.isClass && !type.isSingleItem:
+        this.code.line(
+          `${att.terraformName}: ${defaultCheck}${downcaseFirst(
+            type.struct!.name
+          )}ToTerraform(${varReference}),`
+        );
+        break;
```
ansgarm, line 307: What is this new case handling? (Just out of curiosity)

test/csharp/edge/Main.cs
```diff
@@ -110,12 +131,16 @@ public ProviderStack(Construct scope, string id) : base(scope, id)
                 Alias = "full"
             });
 
-            // Not supported .ReqXXX is optional in CSharp
-            // new RequiredAttributeResource(this, "reqOpt", new RequiredAttributeResourceConfig {
-            //     Bool = providerOpt.Reqbool,
-            //     Num = providerOpt.Reqnum,
-            //     Str = providerOpt.Reqstr
-            // });
+            // TODO: this currently does not compile because provider.reqbool may be undefined
```
ansgarm, line 134: Is this TODO still up to date?

test/java/edge/Main.java
```diff
@@ -113,8 +138,7 @@ public ProviderStack(Construct scope, String id) {
         // TODO: this currently does not compile because provider.reqbool may be undefined
```
ansgarm, line 138: Isn't this TODO outdated now?

ansgarm (approved):
This is some stellar work @jsteinich 💯

packages/@cdktf/provider-generator/lib/get/generator/emitter/attributes-emitter.ts
jsteinich: It's where a non-computed attribute is specified as an object. From what I've seen it is quite rare, but not having it did break the [Kubernetes provider manifest resource wait_for attribute](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/resources/manifest#using-wait_for-to-block-create-and-update-calls).

test/csharp/edge/Main.cs
jsteinich: Was copied from the TS test. I'll update these to have a note, but drop the TODO.

packages/@cdktf/provider-generator/lib/get/generator/emitter/attributes-emitter.ts
ansgarm: Ah, nice. Thanks!

DanielMSchmidt (approved):

skorfmann (approved):

PR #1310, opened 2021-11-17
fix(provider-generator): attributes on resources should not be undefined

### LINKED ISSUE #1200: Prebuilt AWS Typing Error on Generated Resource Properties

DanielMSchmidt:
While this is true in general also on structs we can not implement it there without removing the ability to use setters since setter and getter types need to be the same. This is a short term fix to get ridof the regression we introduced.

Closes #1200

### DIFF
--- a/packages/@cdktf/provider-generator/lib/get/generator/emitter/attributes-emitter.ts
+++ b/packages/@cdktf/provider-generator/lib/get/generator/emitter/attributes-emitter.ts
@@ -21,7 +21,7 @@ export class AttributesEmitter {
     const hasInputMethod = isStored;
 
     if (isStored) {
-      this.code.line(`private ${att.storageName}?: ${att.type.storedName}; `);
+      this.code.line(`private ${att.storageName}?: ${att.type.name}; `);
     }
 
     switch (att.getterType._type) {
@@ -51,15 +51,15 @@ export class AttributesEmitter {
         break;
     }
 
-    if (att.setterType._type === "set") {
-      this.code.openBlock(
-        `public set ${att.name}(value: ${att.setterType.type})`
-      );
+    const setterType = att.setterType;
+
+    if (setterType._type === "set") {
+      this.code.openBlock(`public set ${att.name}(value: ${setterType.type})`);
       this.code.line(`this.${att.storageName} = value;`);
       this.code.closeBlock();
-    } else if (att.setterType._type === "put") {
+    } else if (setterType._type === "put") {
       this.code.openBlock(
-        `public put${titleCase(att.name)}(value: ${att.setterType.type})`
+        `public put${titleCase(att.name)}(value: ${setterType.type})`
       );
       this.code.line(`this.${att.storageName} = value;`);
       this.code.closeBlock();
--- a/packages/@cdktf/provider-generator/lib/get/generator/models/attribute-model.ts
+++ b/packages/@cdktf/provider-generator/lib/get/generator/models/attribute-model.ts
@@ -159,19 +159,21 @@ export class AttributeModel {
   }
 
   public get setterType(): SetterType {
-    return this.isStored
-      ? this.getterType._type === "stored_class"
-        ? {
-            _type: "put",
-            type: this.type.storedName,
-          }
-        : {
-            _type: "set",
-            type: `${this.type.storedName}${
-              this.isProvider ? "| undefined" : ""
-            }`,
-          }
-      : { _type: "none" };
+    if (!this.isStored) {
+      return { _type: "none" };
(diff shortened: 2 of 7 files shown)


test/typescript/providers/main.ts
```diff
@@ -59,14 +59,21 @@ export class NamespacedProviders extends TerraformStack {
...
       {}
     );
+
+    const role = new Aws.iam.IamRole(this, "role", {
+      assumeRolePolicy: "assumeRolePolicy",
+    });
+    new Aws.iam.IamRolePolicyAttachment(this, "lambda-role-vpc-att", {
+      policyArn:
+        "arn:aws:iam::aws:policy/service-role/AWSLambdaVPCAccessExecutionRole",
+      role: role.name,
```
skorfmann, line 69: that's the one which changed before, right? Without these changes it would have to be `role.name!`?

DanielMSchmidt: exactly 👍

skorfmann: top 👍

skorfmann (approved):
:shipit:

packages/@cdktf/provider-generator/lib/get/generator/emitter/attributes-emitter.ts
```diff
@@ -21,7 +24,12 @@ export class AttributesEmitter {
     const hasInputMethod = isStored;
 
     if (isStored) {
-      this.code.line(`private ${att.storageName}?: ${att.type.storedName}; `);
+      this.code.line(
+        `private ${att.storageName}?: ${
+          // TODO: this is a hack to make optional attributes required on resources but not on structs
+          this.context === "resource" ? att.type.name : att.type.storedName
```
jsteinich, line 30: What is the case where this check is needed? Since we are always are emitting as an optional field and the only difference between `type.name` and `type.storedName` is a possible `| undefined` it doesn't seem like it changes anything.

packages/@cdktf/provider-generator/lib/get/generator/models/attribute-model.ts
```diff
@@ -158,20 +158,30 @@ export class AttributeModel {
...
+        _type: "put",
+        type: this.type.storedName,
+      };
+    }
+
+    // TODO: this is a hack to make optional attributes required on resources but not on structs
+    if (context === "resource") {
+      return {
+        _type: "set",
+        type: `${this.type.name}${this.isProvider ? "| undefined" : ""}`,
```
jsteinich, line 177:
```suggestion
        type: `${this.type.name}${this.isProvider ? " | undefined" : ""}`,
```

packages/@cdktf/provider-generator/lib/get/generator/models/attribute-model.ts
```diff
@@ -158,20 +158,30 @@ export class AttributeModel {
...
+      };
+    }
+
+    return {
+      _type: "set",
+      type: `${this.type.storedName}${this.isProvider ? "| undefined" : ""}`,
```
jsteinich, line 183: I'm not sure I follow why structs should be different.

DanielMSchmidt: Because we currently support setters in structs, leading to different types for getters and setters on these. There is a bigger solution that should replace this temporary "make it the old thing in this case" extra case

packages/@cdktf/provider-generator/lib/get/generator/emitter/attributes-emitter.ts
DanielMSchmidt: I remove this, not sure why I had it in the first place. If something arises I'll add a test to catch it 👍

packages/@cdktf/provider-generator/lib/get/generator/models/attribute-model.ts
DanielMSchmidt: I removed this, not sure why I had it in the first place. If something arises I'll add a test to catch it 👍

PR #1311, opened 2021-11-17
test(provider-generator): verify generated provider emit same HCL across languages

### LINKED ISSUE #1207: Test provider generation with jsii
We had various issues around catching provider generation issues, which only popped up in one example after an provider upgrade. Hence, we'll need more explicit test coverage for this. 

This could be done by using a kitchensink provider, or we'll use the most popular ones (aws, google, azurerm) as integration targets.

DanielMSchmidt:
By providing a custom JSON schema we can build edge cases specific to our project

Closes #1207

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,8 @@
       "examples/csharp/*",
       "examples/go/*",
       ".yalc/*",
-      ".yalc/@*/*"
+      ".yalc/@*/*",
+      "test"
     ],
     "nohoist": [
       "archiver",
--- a/package.json
+++ b/package.json
@@ -50,8 +50,7 @@
       "examples/csharp/*",
       "examples/go/*",
       ".yalc/*",
-      ".yalc/@*/*",
-      "test"
+      ".yalc/@*/*"
     ],
     "nohoist": [
       "archiver",
--- a/packages/@cdktf/provider-generator/lib/__tests__/edge-provider-schema/index.ts
+++ b/packages/@cdktf/provider-generator/lib/__tests__/edge-provider-schema/index.ts
@@ -1,0 +1,32 @@
+import { ProviderSchema } from "../../";
+import { schema, SchemaBuilder as S } from "./builder";
+
+const required_attribute_resource = new S()
+  .addAllPrimitiveTypes(true, false)
+  .build();
+
+const optional_attribute_resource = new S()
+  .addAllPrimitiveTypes(false, false)
+  .build();
+
+const optional_computed_attribute_resource = new S()
+  .addAllPrimitiveTypes(false, true)
+  .build();
+
+const list_block_resource = new S()
+  .listBlock("opt", new S().addAllPrimitivePermutations().asBlock(), 0, 42)
+  .listBlock("req", new S().addAllPrimitivePermutations().asBlock(), 1, 42)
+  .listBlock("singleopt", new S().addAllPrimitivePermutations().asBlock(), 0, 1)
+  .listBlock("singlereq", new S().addAllPrimitivePermutations().asBlock(), 1, 1)
+  .build();
+
+export const edgeSchema: ProviderSchema = schema(
+  "edge",
+  new S().addAllPrimitivePermutations().build(),
+  {
+    required_attribute_resource,
+    optional_attribute_resource,
+    optional_computed_attribute_resource,
+    list_block_resource,
+  }
+);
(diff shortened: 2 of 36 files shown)


ansgarm:
Builds fail due to: https://github.com/facebook/jest/issues/12098

packages/@cdktf/provider-generator/lib/__tests__/edge-provider-schema/index.ts
```diff
@@ -0,0 +1,32 @@
...
+const optional_computed_attribute_resource = new S()
+  .addAllPrimitiveTypes(false, true)
+  .build();
+
+const list_block_resource = new S()
+  .listBlock("opt", new S().addAllPrimitivePermutations().asBlock(), 0, 42)
+  .listBlock("req", new S().addAllPrimitivePermutations().asBlock(), 1, 42)
+  .listBlock("singleopt", new S().addAllPrimitivePermutations().asBlock(), 0, 1)
+  .listBlock("singlereq", new S().addAllPrimitivePermutations().asBlock(), 1, 1)
+  .build();
```
skorfmann, line 21: I think this would be easier to read if the arguments for the builder functions were named.

test/csharp/edge/Main.cs
```diff
@@ -0,0 +1,141 @@
...
+using Providers.Null;
+using Providers.Edge;
+using Constructs;
+using HashiCorp.Cdktf;
+
+namespace MyCompany.MyApp
+{
+    // Using references to resource attributes as resource arguments
+    class ReferenceStack : TerraformStack
+    {
```
skorfmann, line 14: this file is auto generated?

test/package.json
```diff
@@ -1,8 +1,12 @@
 {
   "name": "end2end-tests",
+  "private": true,
   "version": "0.0.0",
   "description": "This directory contains integration tests for the cdktf project.",
-  "scripts": {},
+  "scripts": {
+    "edge:install": "rm -rf ./edge-provider-bindings && cp -R ../packages/@cdktf/provider-generator/edge-provider-bindings/ ./edge-provider-bindings",
```
skorfmann, line 7: could this be done via the installed packages in `./run-against-dist` ?

package.json
```diff
@@ -50,7 +50,8 @@
       "examples/csharp/*",
       "examples/go/*",
       ".yalc/*",
-      ".yalc/@*/*"
+      ".yalc/@*/*",
+      "test"
```
skorfmann, line 54: I believe this will pull in a ton of dependencies for all integration tests. Not necessarily a deal breaker, but would be nice to avoid.

skorfmann (changes requested):
Would be helpful to get a high level overview of how that's working in the PR description and / or one of the Readmes (not sure which one would be the right place).

skorfmann (changes requested):
Would be helpful to get a high level overview of how that's working in the PR description and / or one of the Readmes (not sure which one would be the right place).

test/csharp/edge/Main.cs
DanielMSchmidt: No it's hand-written, the autogenerated version does not compile, but was a starting point

test/package.json
DanielMSchmidt: Potentially 🤷  I don't feel super bold on bat so I thought I just take this approach

package.json
DanielMSchmidt: Yeah, I can get rid of that 👍

test/yarn.lock
```diff
@@ -302,6 +302,17 @@
...
 
+"@cdktf/provider-generator@*":
+  version "0.7.0"
+  resolved "https://registry.yarnpkg.com/@cdktf/provider-generator/-/provider-generator-0.7.0.tgz#87574783f78590f37755d0bb5aac56a2e060845d"
+  integrity sha512-PzL3qTxZEnXWFGB+nQOHub8xmEykJKhHLgc6CUEfu9z8PCmK9dnPLK6BKvDgvQtnTT2/aXmFjDr9GVbuOZ5Gsw==
+  dependencies:
+    "@skorfmann/jsii-srcmak" "^0.1.272"
+    cdktf "0.7.0"
+    codemaker "^0.22.0"
+    fs-extra "^8.1.0"
```
skorfmann, line 313: Why is this and other cdktf packages locked to `0.7.0`?

test/package.json
skorfmann: Ok, so it's not used as part of the CI workflow, but a helper when doing this manually. That's probably ok then

packages/@cdktf/provider-generator/lib/__tests__/edge-provider-schema/index.ts
skorfmann: 👍

skorfmann (approved):

PR #1322, opened 2021-11-19
fix(lib): Make sure empty single item lists in synth

jsteinich:
Fixes #1182

The Terraform JSON supports either array or object for single block types. This returns to using an array like before.
`deepMerge` handles arrays differently than empty objects. Maybe we could change that, but likely what's there is necessary. Rather than try to get additional context to that point, it's easier to just use an array.

PR #1329, opened 2021-11-19
fix(lib): Keep specified empty objects

### LINKED ISSUE #1182: AWS Wafv2WebAcl Missing default_action
### Community Note

### cdktf & Language Versions
cdktf: `0.7.0`
language: `typescript`
prebuilt AWS provider `@cdktf/provider-aws`: `2.0.13`

### Affected Resource(s)
WAFV2.Wafv2WebAcl

### Debug Output
https://gist.github.com/vu-bui/9b2b66eaab08794edc9ad4324ee6665d

### Expected Behavior
It should be able to create a Wafv2WebAcl

### Actual Behavior
It throws `defaut_action` is missing though I did specify an empty default action
```
      defaultAction: {
        allow: {},
      },
```
It works on the previous `cdktf` and `@cdktf/provider-aws` version

jsteinich:
Fixes #1182 

This is an alternative to #1322 that doesn't alter the output json in other ways.
I initially thought this approach would have many other side effects, but early testing hasn't revealed any (other than other things that were broken).

### DIFF
--- a/packages/@cdktf/provider-generator/lib/get/__tests__/generator/__snapshots__/complex-computed-types.test.ts.snap
+++ b/packages/@cdktf/provider-generator/lib/get/__tests__/generator/__snapshots__/complex-computed-types.test.ts.snap
@@ -89,6 +89,8 @@ export function acmCertificateOptionsToTerraform(struct?: AcmCertificateOptionsO
 }
 
 export class AcmCertificateOptionsOutputReference extends cdktf.ComplexObject {
+  private isEmptyObject = false;
+
   /**
   * @param terraformResource The parent resource
   * @param terraformAttribute The attribute on the parent resource this class is referencing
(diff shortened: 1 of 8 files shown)


jsteinich:
This is fixing a regression from 0.7, so would be nice to get into 0.8.

Thinking about this some more, it does make sense why the check isn't needed. The entire object will simply be undefined / not present in cases where it hasn't been set, so having it be an empty object means that it needed to have be explicitly set in the 1st place and we should respect that when synthesizing.

ansgarm (approved):

jsteinich:
Ran into some issues with how the `OutputReference` types were handled. I tried a couple solutions, but settled on only keeping and empty object for them when explicitly specified. I don't believe it's very common for an empty block to signify configuration, so needing to be intentional about it seems fine.

jsteinich:
Apparently the updated generated providers were enough to run out of memory when building the providers integration test. Had to alter test setup slightly in order to give node a large heap size.

jsteinich:
Since main had a different approach added to deal with test memory, I cleaned this up to use that and just have the relevant changes.

PR #1342, opened 2021-11-23
chore(docs): document list escape hatches

DanielMSchmidt:
I always have to look up how this works, figured we should have at least a sentence to this topic

### DIFF
--- a/website/docs/cdktf/concepts/providers-and-resources.html.md
+++ b/website/docs/cdktf/concepts/providers-and-resources.html.md
@@ -339,6 +339,8 @@ When you run `cdktf synth`, CDKTF generates a Terraform configuration with the v
 }
 ```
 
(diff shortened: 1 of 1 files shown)


website/docs/cdktf/concepts/providers-and-resources.html.md
```diff
@@ -339,6 +339,8 @@ When you run `cdktf synth`, CDKTF generates a Terraform configuration with the v
 }
 ```
 
+You can access elements in arrays by a dot notation: `resource.addOverride("configurations.0.https", true)`.
```
laurapacilio, line 342:
```suggestion
Use a dot notation to access elements in arrays: `resource.addOverride("configurations.0.https", true)`.
```

laurapacilio (approved):
One tiny suggestion - otherwise, looks good :-)

PR #1346, opened 2021-11-24
feat(lib): Allow relative assets

jsteinich:
This is meant to be a stepping stone towards automatically creating assets for local modules.


jsteinich:
@DanielMSchmidt do you recall why the absolute file restriction was put into place? I just removed the check and added some tests and everything passed.

test/typescript/asset/main.ts
```diff
@@ -27,17 +27,35 @@ function assets(scope: Construct, assetOverrides = {}) {
...
 
+  const relativeAsset = new TerraformAsset(scope, "relative-asset", {
+    path: "relative-asset.txt",
+    ...assetOverrides,
+  });
+
+  const relativeAssets = new TerraformAsset(scope, "relative", {
+    path: "./relative",
+    ...assetOverrides,
+  });
```
skorfmann, line 38: I think this works only by accident, since the files are all present in `process.cwd()`. When referencing from a nested folder structure, this wouldn't work without inferring the file location.  When looking at other languages than Typescript, I'm not sure how this would work.

DanielMSchmidt:
@jsteinich My reasoning was that absolute paths are clear and precise. With relative paths I thought there might be more confusion since it might be relative from the currrent file, relative from cwd, relative from cdktf.json, relative from cdktf.out/stacks/mystack. Using absolute paths removed this confusion

jsteinich:
> @jsteinich My reasoning was that absolute paths are clear and precise. With relative paths I thought there might be more confusion since it might be relative from the currrent file, relative from cwd, relative from cdktf.json, relative from cdktf.out/stacks/mystack. Using absolute paths removed this confusion

Thanks @DanielMSchmidt. That all makes sense and I may have actually fallen into that while testing.

I'll do some more testing with other languages and folder structures. While there will always be the possibility for confusion, if it turns out that there is a consistent answer (I'm leaning towards no), then we could still probably pursue further.

I'll also think about some other approaches to getting local modules copied as assets.

DanielMSchmidt:
> I'll also think about some other approaches to getting local modules copied as assets.

Why are relative paths important for that? Couldn't we require absolute ones on the HclTerraform construct as well?

jsteinich:
> > I'll also think about some other approaches to getting local modules copied as assets.
> 
> Why are relative paths important for that? Couldn't we require absolute ones on the HclTerraform construct as well?

See https://github.com/hashicorp/terraform-cdk/pull/961#issuecomment-915747755

Also it's an extra step when coming from HCl.

jsteinich:
I modified the asset handling to allows treat relative assets as relative to `cdktf.json`. If no `cdktf.json` is found, the path will be rejected with an error.

DanielMSchmidt:
That sounds great 👍 I like the idea

jsteinich:
> Absolute paths are still allowed right?

Absolutely.

packages/cdktf/lib/terraform-asset.ts
```diff
@@ -41,16 +46,27 @@ export class TerraformAsset extends Resource {
...
-    if (!path.isAbsolute(config.path)) {
-      throw new Error(
-        `TerraformAsset path needs to be absolute, got relative path: '${config.path}'`
-      );
+    if (path.isAbsolute(config.path)) {
+      this.sourcePath = config.path;
+    } else {
+      const cdktfJsonPath = findFileAboveCwd("cdktf.json");
+      if (cdktfJsonPath) {
+        // Relative paths are always considered to be relative to cdktf.json, but operations are performed relative to process.cwd
```
DanielMSchmidt, line 54: Could you add this to the docs for TerraformAsset?

DanielMSchmidt (commented):
I like it, seems very straight forward now, thanks a ton 👍

packages/cdktf/lib/terraform-asset.ts
jsteinich: I added to `assets.mdx`. I can also update `terraform-assets.md`, but it sounded like those docs would be going away.

DanielMSchmidt (approved):
Awesome 👍

PR #1348, opened 2021-11-24
chore(cli): typescript init prints link to all prebuilt providers

DanielMSchmidt:


packages/cdktf-cli/templates/typescript/help
```diff
@@ -32,7 +32,8 @@
 
  Use Prebuilt Providers:
 
-  You can add one or multiple of the prebuilt providers listed below:
+  You can find all prebuilt providers on npm: https://www.npmjs.com/search?q=keywords:cdktf
+  Installing a provider is as simple as running
```
ansgarm, line 36: Maybe change this to something along the lines of recommending commonly used providers?
"Simple" is bugging me a bit from a wording perspective – maybe removing that wording would already be enough for me :)

ansgarm (changes requested):

packages/cdktf-cli/templates/typescript/help
DanielMSchmidt: But are they commonly used just because they are prebuilt? I'll remove the simple part

ansgarm: I'm happy with "prebuilt" – I only was unhappy with "as simple as" as that might put people off, that don't know how to work with npm for example 😇

ansgarm (approved):

Issue #1374, opened 2021-12-01
bug(lib): Datadog provider Dashboard Resource doesn't build with 0.8


skorfmann:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### cdktf & Language Versions

- 0.8

### Affected Resource(s)

- DataDog/datadog@~> 3.0
- Dashboard resource

### Description

The Dashboard resource has too man exports and runs into https://github.com/microsoft/TypeScript/issues/38691

In contrast to what got fixed in https://github.com/hashicorp/terraform-cdk/pull/1248 for the AWS provider, the Datadog provider doesn't have namespaces. Tried a few quick hacks, but the struct sharding is too coupled with the namespacing.

PR #1375, opened 2021-12-01
fix(lib): Add struct sharding for providers without namespaces

skorfmann:
The Dashboard resource has too man exports and runs into microsoft/TypeScript#38691

In contrast to what got fixed in #1248 for the AWS provider, the Datadog provider doesn't have namespaces. Hence, this PR adds struct sharding for providers without namespaces.
 
Fixes https://github.com/hashicorp/terraform-cdk/issues/1374


packages/@cdktf/provider-generator/lib/get/__tests__/provider.test.ts
```diff
@@ -14,3 +14,5 @@ getProvider(
 );
 
 getProvider(new TerraformProviderConstraint("oci@= 4.13.0"));
+
+getProvider(new TerraformProviderConstraint("DataDog/datadog@~> 3.0"));
```
DanielMSchmidt, line 18: The other version constraints are pinned, I guess that would make sense for this provider as well

packages/@cdktf/provider-generator/lib/get/generator/models/resource-model.ts
```diff
@@ -178,11 +178,18 @@ export class ResourceModel {
...
 
   public get namespacedFilePath(): string {
-    return path.join(
-      this.filePath.split("/").slice(0, -1).join("/"),
-      this.namespace!.name,
-      this.structsFolderName
-    );
+    if (this.namespace) {
+      return path.join(
+        this.filePath.split("/").slice(0, -1).join("/"),
```
DanielMSchmidt, line 183: Nitpick: Moving this into a variable with an explanatory name would be nice for readability

DanielMSchmidt (approved):
Only small nit picks 👍

packages/@cdktf/provider-generator/lib/get/generator/provider-generator.ts
```diff
@@ -102,6 +102,8 @@ export class TerraformProviderGenerator {
           namespacedResources[namespace] = [];
         }
         namespacedResources[namespace].push(resourceModel);
+      } else if (resourceModel.structsRequireNamespace) {
```
ansgarm, line 105: I'm a bit confused by the naming `structsRequireNamespace` (I know it existed previously) but will this create a namespace or "just" a resource that lives in a directory instead of a single file? Will this then already be a namespace?

On a related note: What happens, if a Terraform provider gets updated and a resource then passes the limit of 400: Will this automatically namespace the resource (Can it then be imported from `/lib/some-resource`)? If this happens unexpectedly, will users experience broken imports after running `get` for a new provider resource? Or will this only add the opportunity for that single resource to now be imported from a `/lib/some-resource` path while still keeping the old way of importing that resource?

skorfmann: >On a related note: What happens, if a Terraform provider gets updated and a resource then passes the limit of 400: Will this automatically namespace the resource (Can it then be imported from /lib/some-resource)? If this happens unexpectedly, will users experience broken imports after running get for a new provider resource? Or will this only add the opportunity for that single resource to now be imported from a /lib/some-resource path while still keeping the old way of importing that resource?

This would be transparent to the user and work automatically without a breaking change.

> I'm a bit confused by the naming structsRequireNamespace (I know it existed previously) but will this create a namespace or "just" a resource that lives in a directory instead of a single file? Will this then already be a namespace?

Yeah, the name is confusing, should adapt it. This will perform the struct sharding without namespacing

packages/@cdktf/provider-generator/lib/get/generator/provider-generator.ts
```diff
@@ -253,6 +255,75 @@ export class TerraformProviderGenerator {
...
+      }
+
+      this.code.line(`export * from './${resource.structsFolderName}'`);
+
+      resource.importStatements.forEach((statement) =>
+        this.code.line(statement)
+      );
+      this.code.line();
+      this.code.line(`/**`);
+      this.code.line(`* SOME COMMENT`);
```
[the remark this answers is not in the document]
skorfmann: Heh, I'll check

PR #1401, opened 2021-12-09
fix(lib): inline references in strings should not destroy string itself

DanielMSchmidt:
Found this bug while iterating on cross stack references

### DIFF
--- a/packages/cdktf/lib/tfExpression.ts
+++ b/packages/cdktf/lib/tfExpression.ts
@@ -36,7 +36,7 @@ class TFExpression extends Intrinsic implements IResolvable {
    * It must only be used on non-token values
    */
   protected escapeString(str: string) {
-    return str // Escape double quotes
+    return str
       .replace(/\n/g, "\\n") // escape newlines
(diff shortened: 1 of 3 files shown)


skorfmann (approved):

packages/cdktf/lib/tfExpression.ts
```diff
@@ -63,12 +63,24 @@ class TFExpression extends Intrinsic implements IResolvable {
...
+        if (leftTokenList.literals.length === 1 && leftTokenCount === 0) {
+          leftValue = this.escapeString(left);
+        }
+
+        // if left is only a token, needs to be wrapped as terraform expression
+        if (leftTokenList.literals.length === 0 && leftTokenCount === 1) {
+          leftValue = `\${${leftTokens[0]}}`;
+        }
 
         const rightValue =
```
jsteinich, line 85: Could the same issue happen with the right side?

DanielMSchmidt: No, the token list is joined from left to right

PR #1411, opened 2021-12-12
feat(lib): Map Tokens

jsteinich:
Dividing up #1299, starting with tokens for map (string indexed / Record<string, any>) types as these are quite common.

Some points that came up from the original design:

- Map value type needs to be exact in order for jsii to work. Added support for all primitive types and left `IResolvable` option for types of maps
- Added an `AnyMap` computed wrapper for completions sake
- boolean maps support tokens, but since booleans don't, added `IResolvable` to the value type in that case
- Didn't fully handle `map<list>` type since current code doesn't distinguish between that and `list<map>` type. It's very rare, so that can be a different project down the road
- Added a fix for the C# edge provider Constructs version error
- I didn't add to the go edge provider test. I could take a stab at it, but probably for someone with go experience (or we can add later).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 **Breaking Changes**
 
+### Map Tokens [#1411](https://github.com/hashicorp/terraform-cdk/pull/1411)
+
+As part of an effort to use more native types, there are now tokens for maps of primitive values.
+
+As a result, there is a minor breaking change:
+
+- Map attributes have gone from `{ [key: string]: TYPE } | cdktf.IResolvable` to `{ [key: string]:TYPE }` when `TYPE` is `string, number, or boolean`.
+  - The most common impact is maps created by using Terraform functions (`Fn.(...)`) will now need to be passed to `Token.as<String/Number/Boolean>Map()` before assigning to a resource attribute.
+
 ### Number[] Tokens [#1471](https://github.com/hashicorp/terraform-cdk/pull/1471)
 
 As part of an effort to use more native types, there are now tokens for `number[]`.
--- a/packages/@cdktf/provider-generator/lib/get/generator/models/attribute-model.ts
+++ b/packages/@cdktf/provider-generator/lib/get/generator/models/attribute-model.ts
@@ -144,6 +144,9 @@ export class AttributeModel {
     if (type.isBooleanMap) {
       return `boolean`;
     }
+    if (type.isAnyMap) {
+      return `any`;
+    }
     if (process.env.DEBUG) {
       console.error(
         `The attribute ${JSON.stringify(this)} isn't implemented yet`
--- a/packages/@cdktf/provider-generator/lib/get/generator/models/attribute-model.ts
+++ b/packages/@cdktf/provider-generator/lib/get/generator/models/attribute-model.ts
@@ -135,16 +135,16 @@ export class AttributeModel {
 
   public get mapType(): string {
     const type = this.type;
-    if (type.isStringMap) {
+    if (type.isStringMap || (type.isMap && type.innerType === "string")) {
       return `string`;
     }
-    if (type.isNumberMap) {
+    if (type.isNumberMap || (type.isMap && type.innerType === "number")) {
       return `number`;
     }
-    if (type.isBooleanMap) {
+    if (type.isBooleanMap || (type.isMap && type.innerType === "boolean")) {
       return `boolean`;
     }
-    if (type.isAnyMap) {
+    if (type.isAnyMap || (type.isMap && type.innerType === "any")) {
       return `any`;
     }
     if (process.env.DEBUG) {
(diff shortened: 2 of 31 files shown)


packages/cdktf/lib/tokens/token.ts
```diff
@@ -82,6 +82,62 @@ export class Token {
...
+  }
+
+  /**
+   * Return a reversible map representation of this token
+   */
+  public static asStringMap(
+    value: any,
+    options: EncodingOptions = {}
+  ): { [key: string]: string } {
+    return this.asMap(value, "", options);
```
DanielMSchmidt, line 108: If I understand this correctly `""` is the value used in the map, e..g `{ "&[Token.1]": ""}`. I wonder if we can make this string more telling that it's part of a token value and not a real string, so that if someone accidentally accesses e.g. all values of a token and sees an empty string they have some more clues to what is going on. Long-term we should check this value when resolving outside of a map context, similar to the warning we show when accessing a value from within a list, but that is probably out of scope here

DanielMSchmidt (commented):
Really awesome work 🎉  I got some comments more in the realm of nit-picks, but I think they are easy and worthy to address :)

packages/@cdktf/provider-generator/lib/get/generator/models/attribute-model.ts
```diff
@@ -135,15 +135,18 @@ export class AttributeModel {
 
   public get mapType(): string {
     const type = this.type;
-    if (type.isStringMap) {
+    if (type.isStringMap || (type.isMap && type.innerType === "string")) {
```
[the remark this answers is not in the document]
jsteinich: It doesn't actually match that, but I agree that it should.
I've been contemplating a bigger refactor to how the types are represented, but have been holding off. I'll won't put anything big in this PR, but I'll make a couple small updates to clarify.

packages/cdktf/lib/tokens/token.ts
jsteinich: Sounds good. I initially had it as the inner token part, but changed when needing to make type specific versions.
Easy to use any string here, so I'll make a constant value that is more token like. I can also put in a dummy value for number maps.
If the check seems simple enough, I'll just add it; otherwise, I'll put in a follow up PR.

packages/@cdktf/provider-generator/lib/get/generator/models/attribute-model.ts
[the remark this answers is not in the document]
DanielMSchmidt: Looking forward to that bigger refactor <3 I hope we can get it to a state where changing a return type or so can happen in one instead of 5 places :D 

Thank you for addressing this while keeping the scope small 👍

packages/cdktf/lib/tokens/token.ts
DanielMSchmidt: Sounds great, thank you 👍

test/csharp/edge/test.ts
```diff
@@ -28,6 +28,12 @@ describe("csharp full integration test synth", () => {
       stdio: "inherit",
       env: driver.env,
     });
+
+    execSync("dotnet add package Constructs", {
+      stdio: "inherit",
+      env: driver.env,
+    });
+
```
jsteinich, line 36: @DanielMSchmidt feel free to pull this out into a PR so constructs updates stop breaking tests.

DanielMSchmidt: Done, thank you! https://github.com/hashicorp/terraform-cdk/pull/1460

DanielMSchmidt:
@jsteinich could you rebase this?

DanielMSchmidt (approved):

skorfmann:
@jsteinich Could you add a change log entry, please?

jsteinich:
> @jsteinich Could you add a change log entry, please?

I added a short entry. Let me know if that's not what you had in mind.

CHANGELOG.md
```diff
@@ -1,3 +1,16 @@
...
+**Breaking Changes**
+
+### Map Tokens [#1411](https://github.com/hashicorp/terraform-cdk/pull/1411)
+
+As part of an effort to use more native types, there are now tokens for maps of primitive values.
+
+As a result, there is a minor breaking change:
+
+- Map attributes have gone from `{ [key: string]: TYPE } | cdktf.IResolvable` to `{ [key: string]:TYPE }` when `TYPE` is `string, number, or boolean`.
+  - The most common impact is maps created by using Terraform functions (`Fn.(...)`) will now need to be passed to `Token.as<String/Number/Boolean>Map()` before assigning to a resource attribute.
```
DanielMSchmidt, line 12: So the bottom line is adding `| cdktf.IResolvable` would make TF functions easier to use, but it would obfuscate the type in Java, C#, and Go right?

DanielMSchmidt: I checked with @ansgarm and he too is okay with that API change :)

jsteinich:
Test failure seems unrelated:
```
Command failed: terraform init
    ╷
    │ Error: Failed to install provider
    │ 
    │ Error while installing kreuzwerker/docker v2.14.0: could not query provider
    │ registry for registry.terraform.io/kreuzwerker/docker: failed to retrieve
    │ authentication checksums for provider: 403 Server failed to authenticate
    │ the request. Make sure the value of Authorization header is formed
    │ correctly including the signature.
    ╵
```

PR #1415, opened 2021-12-13
feat(lib): Initial improvements for set support

### LINKED ISSUE #992: Support Sets for providers and modules
### Community Note

### Description

Currently sets either [fail the generation](https://github.com/hashicorp/terraform-cdk/issues/928) or [are treated as lists, making the access impossible](https://github.com/hashicorp/terraform-cdk/issues/424). We should implement `ComplexComputedSet`s that provides helpers for accessing individual items or all items as a list. 

In the provider generation, if a set is passed as input (e.g. datadog_monitor#renotify_statuses) Terraform will error on execution.

#### Requirements
- [ ] Add a testcase to the edge provider tests (introduced in https://github.com/hashicorp/terraform-cdk/pull/1311)

### References

- https://github.com/hashicorp/terraform-cdk/issues/928
- https://github.com/hashicorp/terraform-cdk/issues/424

ansgarm:
### Background
Before this PR the CDK for Terraform treated sets as arrays. While this works when setting data (e.g. passing `["something"]` to a resource argument of the type `set(string)`) accessing that data can have unexpected consequences. For example using `Fn.element(list, index)` does not work on set types. This was not visible from the API we generated (as we exposed all sets as "array of something"). This PR tries to improve the current state by explicitly wrapping references to set type attributes with `Fn.tolist()` which enables `Fn.element()` to work. This is only a first intermediate step until JSII has supports for sets ([issue](https://github.com/aws/jsii/issues/1)) so that we can model them properly in all target languages.

Resolves #992.

### DIFF
--- a/packages/@cdktf/provider-generator/lib/__tests__/__snapshots__/edge-provider-schema.test.ts.snap
+++ b/packages/@cdktf/provider-generator/lib/__tests__/__snapshots__/edge-provider-schema.test.ts.snap
@@ -280,6 +280,7 @@ export * from './optional-attribute-resource';
 export * from './optional-computed-attribute-resource';
 export * from './list-block-resource';
 export * from './map-resource';
+export * from './set-block-resource';
 export * from './edge-provider';
 
 "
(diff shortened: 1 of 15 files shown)


packages/@cdktf/provider-generator/lib/get/generator/emitter/attributes-emitter.ts
```diff
@@ -148,6 +151,9 @@ export class AttributesEmitter {
     }
 
     this.code.line(`// Getting the computed value is not yet implemented`);
+    if (type.isSet) {
+      return `cdktf.Fn.tolist(this.interpolationForAttribute('${att.terraformName}')) as any`;
```
jsteinich, line 155: This really doesn't work. Perhaps we can wait for a better fix since the line below this was already broken, but doesn't feel great to add more broken things.

packages/@cdktf/provider-generator/lib/get/generator/models/attribute-type-model.ts
```diff
@@ -94,8 +98,20 @@ export class AttributeTypeModel {
     return this.name === TokenizableTypes.NUMBER;
   }
 
+  public get isStringSet(): boolean {
+    return this.isSet && this.name === TokenizableTypes.STRING_LIST;
```
jsteinich, line 102: I think checking `_type` would be more clear

packages/cdktf/lib/complex-computed-list.ts
```diff
@@ -75,12 +76,26 @@ export class ComplexComputedList extends ComplexComputedAttribute {
...
+    if (this.wrapsSet) {
+      const x = propertyAccess(
+        Fn.tolist(
+          this.terraformResource.interpolationForAttribute(
+            this.terraformAttribute
+          )
+        ),
+        [this.complexComputedListIndex, property]
+      );
+      console.log(x);
```
jsteinich, line 95: I assume this line is for debugging.

packages/cdktf/lib/runtime.ts
```diff
@@ -24,6 +25,13 @@ export const booleanToTerraform: Mapper = identity;
 export const anyToTerraform: Mapper = identity;
 export const numberToTerraform: Mapper = identity;
 
+export function setMapper(elementMapper: Mapper): Mapper {
+  return (x: any) => {
+    if (!canInspect(x)) return x; // do not wrap undefined in toset()
+    return Fn.toset(listMapper(elementMapper)(x));
```
jsteinich, line 31: Does it need to be wrapped in `toset` or will terraform contextually know that it's a set?

packages/cdktf/lib/tokens/private/resolve.ts
```diff
@@ -111,8 +111,9 @@ export function resolve(obj: any, options: IResolveOptions): any {
   if (typeof obj === "string") {
     // If this is a "list element" Token, it should never occur by itself in string context
     if (TokenString.forListToken(obj).test()) {
+      console.log({ obj });
```
jsteinich, line 114: More debugging?

jsteinich (commented):
I know this is still in-progress, but wanted to give some initial feedback.
Maybe someday [#jsii/1](https://github.com/aws/jsii/issues/1) will get implemented and we provider support without the side effects, but this is infinitely better than just being broken.

packages/@cdktf/provider-generator/lib/get/generator/models/attribute-type-model.ts
ansgarm: I took that from the `isStringList()` check – I suspect that `name()` handles further cases (i.e. single item lists) but they may also not apply for sets.
But I agree that this whole class could be refactored 😅

packages/cdktf/lib/runtime.ts
ansgarm: I _think_ Terraform might imply it. I have to check that.

packages/@cdktf/provider-generator/lib/get/generator/emitter/attributes-emitter.ts
ansgarm: Yeah 😅 I just wrapped it in `Token.asAny()` now it at least works in TypeScript (added a testcase for the edge provider test). Without the `Token.asAny()` the result of `tolist` was a list token which got stuck in the `listMapper` of the receiving resource (as that Token is an Array with a single element and so the listMapper tried to map the object). But with `Token.asAny()` the behaviour is now the same (half broken 😅) one as for lists.

packages/cdktf/lib/tokens/private/resolve.ts
ansgarm: removed ✅

packages/cdktf/lib/complex-computed-list.ts
ansgarm: removed ✅

packages/@cdktf/provider-generator/lib/get/generator/emitter/attributes-emitter.ts
```diff
@@ -148,6 +151,14 @@ export class AttributesEmitter {
     }
 
     this.code.line(`// Getting the computed value is not yet implemented`);
+    if (type.isSet) {
```
DanielMSchmidt, line 154: What would be required to make it work in non-ts languages?

DanielMSchmidt (commented):
Can we document the things around sets that are not supported yet somewhere so it's clearer?

packages/@cdktf/provider-generator/lib/get/generator/emitter/attributes-emitter.ts
jsteinich: This would require moving away from native array return types in all cases or dynamically generated token types.

When working with jsii, the type of the returned object needs to actually match the type signature of the method.

packages/cdktf/lib/runtime.ts
ansgarm: I removed it 👍

packages/@cdktf/provider-generator/lib/__tests__/__snapshots__/edge-provider-schema.test.ts.snap
```diff
@@ -1885,3 +1886,147 @@ export class RequiredAttributeResource extends cdktf.TerraformResource {
...
+  // ATTRIBUTES
+  // ==========
+
+  // set - computed: false, optional: true, required: false
+  private _set?: SetBlockResourceSet[]; 
+  public get set() {
+    // Getting the computed value is not yet implemented
+    return cdktf.Token.asAny(cdktf.Fn.tolist(this.interpolationForAttribute('set'))) as any;
+  }
+  public set set(value: SetBlockResourceSet[]) {
```
DanielMSchmidt, line 2010: :D public set set

packages/@cdktf/provider-generator/lib/get/generator/emitter/attributes-emitter.ts
```diff
@@ -156,6 +159,14 @@ export class AttributesEmitter {
...
+      // Token.asAny is required because tolist returns an Array encoded token which the listMapper
+      // would try to map over when this is passed to another resource. With Token.asAny() it is left
+      // as is by the listMapper and is later properly resolved to a reference
+      // (this only works in TypeScript currently, same as for referencing lists
+      // [see "interpolationForAttribute(...) as any" further below])
```
DanielMSchmidt, line 167: If we don't get to replace complex computed lists with native lists for 0.9 we might need to add sth to the docs around set access only supported in TS, don't we?

packages/@cdktf/provider-generator/lib/get/generator/models/attribute-type-model.ts
```diff
@@ -69,26 +72,25 @@ export class AttributeTypeModel {
       return `{ [key: string]: ${this._type} } | cdktf.IResolvable`;
 
     // single item list
-    if (this.isList && !this.isComputed && this.isSingleItem)
+    if ((this.isList || this.isSet) && !this.isComputed && this.isSingleItem)
```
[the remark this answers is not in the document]
ansgarm: neat idea 👍

packages/@cdktf/provider-generator/lib/get/generator/emitter/attributes-emitter.ts
ansgarm: Hm, that's the same as it already is. Simple sets (i.e. accessing a String set with `element()`) should work now without casting the set to a list using `tolist()` in any language.

PR #1416, opened 2021-12-13
feat(lib): cross-stack references

### LINKED ISSUE #651: Cross Stack References
### Community Note

### Description

With #636 we've introduced multiple stacks. We should enable referencing resources across stacks.

### References

- Blocked on #525

DanielMSchmidt:
Closes #651

## General

This PR adds support for cross stack references (initially for the local and Terraform Cloud backend being used in the source stack).
A Cross Stack Reference is a reference that is used in a different stack than it's created, e.g.:

```ts
export class SourceStack extends TerraformStack {
  public diceRoll: Integer;
    super(scope, id);

    new RandomProvider(this, "random");

    this.diceRoll =  new Integer(this, "int", {
      min: 0,
      max: 6,
    });
  }
}

export class ConsumerStack extends TerraformStack {
  constructor(
    scope: Construct,
    id: string,
    inputs: { randomValue: number }
  ) {
    super(scope, id);

    new LocalProvider(this, "local");

    new File(this, "dice-roll-result", {
        filename: "dice-roll-result.txt",
        content: String(inputs.randomValue)
    })
  }
}

const app = new App();
const src = new SourceStack(app, "source");
new ConsumerStack(app, "destination", {
  randomValue: src.diceRoll.result, // we use the reference from the source stack here
});
app.synth();
``` 

## How to review

I'd recommend going commmit by commit, they should all make sense by themselves


jsteinich:
Does this supersede #1179? There were some comments there that may still apply here.

DanielMSchmidt:
Yes it will, I left the other PR open so that I don't forget to look through the comments

packages/cdktf-cli/bin/cmds/helper/stack-dependencies.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  public finishRun(stack: SynthesizedStack): void {
+    this.inflightStacks = this.inflightStacks.filter(
+      (item) => item.name !== stack.name
+    );
+    this.deployedStacks.push(stack);
+  }
+
+  public pendingDeployableStacks(): SynthesizedStack[] {
```
ansgarm, line 26: Nit: Should we make this a getter? We don't have a common codestyle / linting for such things (yet) – no hard feelings though 😁

packages/cdktf/lib/app.ts
```diff
@@ -81,6 +84,30 @@ export class App extends Construct {
     this.manifest = new Manifest(version, this.outdir);
   }
 
+  public static isApp(x: any): x is App {
+    return x !== null && typeof x === "object" && APP_SYMBOL in x;
+  }
+
+  public static of(construct: IConstruct): App {
```
ansgarm, line 91: love this ❤️

packages/cdktf/lib/backends/local-backend.ts
```diff
@@ -14,6 +15,19 @@ export class LocalBackend extends TerraformBackend {
...
+  public getRemoteStateDataSource(
+    scope: Construct,
+    name: string,
+    fromStack: string
+  ): TerraformRemoteState {
+    return new DataTerraformRemoteStateLocal(scope, name, {
+      workspace: this.props.workspaceDir,
+      path:
+        this.props.path ||
+        path.join(process.cwd(), `terraform.${fromStack}.tfstate`),
```
ansgarm, line 28: Theoretically a user could overwrite `terraform.${fromStack}.tfstate` by registering a local backend for their stack explicitly instead of relying on the default local backend which the cdktf-cli uses if no backend has been specified.
Maybe something to add as a future task to improve this?

packages/cdktf/lib/terraform-backend.ts
```diff
@@ -1,17 +1,34 @@
...
     return {};
   }
 
+  /**
+   * Creates a TerraformRemoteState resource that accesses this backend.
+   */
+  public abstract getRemoteStateDataSource(
+    _scope: Construct,
+    _name: string,
+    _fromStack: string
```
ansgarm, line 29: Are underscores needed for "unused" parameters of abstract methods?

packages/cdktf/lib/terraform-stack.ts
```diff
@@ -137,21 +188,14 @@ export class TerraformStack extends Construct {
...
-        providers.push(node);
-      }
-
-      for (const child of node.node.children) {
-        visit(child);
-      }
-    };
-
-    visit(this);
+    return this.findAll({ byConstructor: TerraformProvider });
```
ansgarm, line 191: 😍

packages/cdktf/lib/terraform-stack.ts
```diff
@@ -184,6 +228,64 @@ export class TerraformStack extends Construct {
 
     return resolve(this, tf);
   }
+
+  public registerOutgoingCrossStackReference(identifier: string) {
```
ansgarm, line 232: Can register accidentally be called with the same identifier for different references?

packages/cdktf/lib/terraform-stack.ts
```diff
@@ -184,6 +228,64 @@ export class TerraformStack extends Construct {
...
+  }
+
+  public registerIncomingCrossStackReference(fromStack: TerraformStack) {
+    if (this.crossStackDataSources[String(fromStack)]) {
+      return this.crossStackDataSources[String(fromStack)];
+    }
+    const originBackend = fromStack.backend;
+    const originPath = fromStack.node.path;
+
+    // TODO: stack name in construct identifier?
```
ansgarm, line 257: todo left

packages/cdktf/lib/tfExpression.ts
```diff
@@ -114,18 +116,47 @@ export function rawString(str: string): IResolvable {
 }
 
 class Reference extends TFExpression {
-  constructor(private identifier: string) {
+  private crossStackIdentifier: Record<string, string> = {};
```
ansgarm, line 119: a hint what maps to what could help here

test/python/cross-stack-references/__snapshots__/test.ts.snap
```diff
@@ -0,0 +1,57 @@
...
+      ]
+    },
+    \\"functionOutput\\": {
+      \\"name\\": \\"functionOutput\\",
+      \\"constructPath\\": \\"functionOutput\\",
+      \\"workingDirectory\\": \\"stacks/functionOutput\\",
+      \\"synthesizedStackPath\\": \\"stacks/functionOutput/cdk.tf.json\\",
+      \\"annotations\\": [],
+      \\"dependencies\\": [
+        \\"source\\"
```
ansgarm, line 52: Shouldn't `functionOutput` depend on `fns`?

ansgarm: Apparently it does not need to (because the Terraform function called in `fns` returns the same result when called in `functionOutput`). Doesn't look like expected however – maybe use an instantiated resource in `fns` instead of the TF functions?

test/typescript/cross-stack-references/__snapshots__/test.ts.snap
```diff
@@ -0,0 +1,57 @@
...
+      ]
+    },
+    \\"functionOutput\\": {
+      \\"name\\": \\"functionOutput\\",
+      \\"constructPath\\": \\"functionOutput\\",
+      \\"workingDirectory\\": \\"stacks/functionOutput\\",
+      \\"synthesizedStackPath\\": \\"stacks/functionOutput/cdk.tf.json\\",
+      \\"annotations\\": [],
+      \\"dependencies\\": [
+        \\"source\\"
```
ansgarm, line 52: Same as for the Python snapshot

test/typescript/terraform-cloud/test.ts
```diff
@@ -60,10 +60,34 @@ describe("full integration test", () => {
...
+          terraformVersion: TERRAFORM_VERSION,
+        },
+        type: "workspaces",
+      },
+    });
+
+    driver.deploy("source-stack");
+    driver.deploy("consumer-stack");
+
+    await client.Workspaces.deleteByName(orgName, workspaceName);
```
ansgarm, line 87: Maybe add a note that only source stack is configured to deploy to Terraform Cloud?
For full Terraform Cloud support (both stacks deployed in separate TFC workspaces) we need at least docs / guidance for users to inform them that they have to set the correct permissions to the TFC workspaces if they are using remote execution – else the consumer workspace can't access the state of the source workspace I think.
Ideally the cdktf-cli could read the stack dependencies and check this in TFC and offer to setup the required permissions. But that's definitely something for the future 😁

website/docs/cdktf/concepts/stacks.mdx
```diff
@@ -118,9 +118,57 @@ To run multiple stacks at once, move them into the same directory and run `cdktf
 
 ### Cross-Stack References
 
-CDKTF does not yet support referencing resources from another stack automatically. You can achieve this manually with [outputs](/cdktf/concepts/variables-and-outputs) and the [remote state data source](/language/state/remote-state-data).
+When you reference resources from one stack in another stack, you can do so by exposing the resource in the source stack and referencing it in the target stack.
 
-Please refer to this [issue](https://github.com/hashicorp/terraform-cdk/issues/651) if you are interested in using cross-stack references.
+```typescript
```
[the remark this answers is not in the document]
DanielMSchmidt: Yeah that makes sense, I'll rewrite it 👍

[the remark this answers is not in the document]
DanielMSchmidt: I pushed a new version, is that what you envisioned?

examples/typescript/aws-kubernetes/main.ts
```diff
@@ -0,0 +1,241 @@
...
+          asgDesiredCapacity: 1,
+        },
+      ],
+
+      tags: {
+        githubRepo: "hashicorp/terraform-cdk",
+      },
+    });
+
+    this.eks = new eks.DataAwsEksCluster(this, "eks-cluster", {
```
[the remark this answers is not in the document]
DanielMSchmidt: Yeah makse sense

examples/typescript/aws-kubernetes/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "synth": "cdktf synth",
+    "compile": "tsc --pretty",
+    "watch": "tsc -w",
+    "test": "jest",
+    "test:watch": "jest --watch",
+    "upgrade": "npm i cdktf@latest cdktf-cli@latest",
+    "upgrade:next": "npm i cdktf@next cdktf-cli@next"
+  },
+  "engines": {
+    "node": ">=10.12"
```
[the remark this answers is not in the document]
DanielMSchmidt: I copied another example as a start, let's do this update afterwards in one sweep

packages/cdktf/lib/backends/local-backend.ts
DanielMSchmidt: This is why we prefer to take the path configured in the origin backend here and only then default to the default value for the path. It's tested here: https://github.com/hashicorp/terraform-cdk/blob/bf4c591747b2ea5fc21a8817e5ea7d8efe28139c/packages/cdktf/test/app.test.ts#L287

packages/cdktf/lib/terraform-backend.ts
DanielMSchmidt: Huh, apparently not. I think I had it non-abstract before and didn't change the args

packages/cdktf/lib/backends/local-backend.ts
ansgarm: Oh, I missed that 👍
However, if the user explicitly specifies the LocalBackend without a path, that [path might default to](https://www.terraform.io/language/settings/backends/local#path) `terraform.tfstate` (as that backend could then take precedence over the legacy `-state` option we pass). But yeah, quite an edge case 🙈

examples/typescript/aws-kubernetes/package.json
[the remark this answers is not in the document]
ansgarm: yeah, fine with me 👍

packages/cdktf/lib/terraform-stack.ts
DanielMSchmidt: Yes it can, even intentionally, e.g. if you take the VPC example and reference the VPC in two different application stacks. This would call the function twice, creating only one (since caching is in place) output. 

This is no problem though, we just need the data in an output, there is no need to have one output per referencing stack.

packages/cdktf/lib/terraform-stack.ts
DanielMSchmidt: ✅

test/python/cross-stack-references/__snapshots__/test.ts.snap
DanielMSchmidt: The origin of the data is pinned down. I'd argue that this is correct, we should only automatically create data-level dependencies between stacks, but I'll make sure to document this.

packages/cdktf/lib/backends/local-backend.ts
DanielMSchmidt: Ah related to https://github.com/hashicorp/terraform-cdk/issues/1430 right?

packages/cdktf/lib/terraform-stack.ts
ansgarm: Ah, got it

ansgarm:
Just came across this when reading a former proposal for cross stack references I did a while back: https://github.com/aws/aws-cdk/pull/12778
We should at least create a follow-up task to check whether this could happen to us as well.

ansgarm (approved):

skorfmann:
Did some testing with one the examples, here a few findings and questions: 

could you elaborate on why `sensitive` true?

```
// automatically generated output
"typescriptawscloudfrontproxy_crossstackoutputawsroute53recordtypescriptawscloudfrontproxyCertValidationRecord8AA53AEAfqdn_14831FB7": {
  "value": "${aws_route53_record.typescriptawscloudfrontproxy_CertValidationRecord_8AA53AEA.fqdn}",
  "sensitive": true
}
```

---

when passing a full resource across a stack, it seems to work when synthing. When looking at the producing stack there's no output - which is probably correct. However, the consuming stack still tries to reference the resource. Would be throwing an explicit error an option here?

```
  "output": {
    "consuming_fqdn_783793B2": {
      "value": "${aws_route53_record.typescriptawscloudfrontproxy_CertValidationRecordFoo_A2C12652}"
    }
  },
```

---

when passing a complex computed object

```
"consuming_fqdn_783793B2": {
  "value": {
    "terraformResource": "${aws_acm_certificate.typescriptawscloudfrontproxy_cert_EDF00C08}",
    "terraformAttribute": "domain_validation_options",
    "complexComputedListIndex": "0"
  }
}
```

---

is this semantically the same? Currently, `this.myFqdn` creates an output only when referenced in a dependent stack.

```
new TerraformOutput(this, "distribution_domain_name", {
      value: distribution.domainName,
    });

this.myFqdn = record.fqdn;
```

---

when using a Function, while this would probably be working, I think it's unexpected behaviour:

```ts
// producing
this.myFqdn = Fn.split('.', record.fqdn);
```

```
// producing
  "output": {
    "typescriptawscloudfrontproxy_crossstackoutputawsroute53recordtypescriptawscloudfrontproxyCertValidationRecordFooA2C12652fqdn_37E460F3": {
      "value": "${aws_route53_record.typescriptawscloudfrontproxy_CertValidationRecordFoo_A2C12652.fqdn}",
      "sensitive": true
    }
  }

// consuming
  "output": {
    "consuming_fqdn_783793B2": {
      "value": "${split(\".\", data.terraform_remote_state.consuming_crossstackreferenceinputtypescriptawscloudfrontproxy_A0330E7F.outputs.typescriptawscloudfrontproxy_crossstackoutputawsroute53recordtypescriptawscloudfrontproxyCertValidationRecordFooA2C12652fqdn_37E460F3)}"
    }
  },
```

DanielMSchmidt:
> could you elaborate on why sensitive true?

Yes, as these outputs are implicitly added they are not supposed to be used for anything besides the x-stack ref. Since they are implicit a user can not configure one to be sensitive so defaulting to sensitive seems like a good choice.

> when passing a full resource across a stack, it seems to work when synthing.

I wonder if this is a real use-case though, wouldn't folks use the attributes of a resource? I think this is never a good idea and would argue for disallowing passing these into outputs and terraform functions in general


> when passing a complex computed object

That is really weird, this should have triggered a cross stack ref to exist. I'm going to add a test case for this 👍 

> Currently, this.myFqdn creates an output only when referenced in a dependent stack.

Yeah that's the plan. Outputs are only created for referenced attributes.


> when using a Function, while this would probably be working, I think it's unexpected behaviour:

Why is this unexpected? The data origins in the producing stack so a cross stack ref is created. the function doesn't live in any state so it's pulled in the using stack. The only issue I could see is when using it with `Fn.timestamp`, but that seems like an artificial use-case to me. The alternative would be to create a terraform local when cross stack referencing (either in general or when inside a TF function)

jsteinich:
> I wonder if this is a real use-case though, wouldn't folks use the attributes of a resource? I think this is never a good idea and would argue for disallowing passing these into outputs and terraform functions in general

Passing complete resources into an output is definitely valid. Terraform accepts that and there was a ticket fairly recently where a user requested the functionality and we made it work. It's not possible to pass to any other resource though and also seems a bit far fetched to pass into another stack's output.

skorfmann:
> > when using a Function, while this would probably be working, I think it's unexpected behaviour:
> 
> Why is this unexpected? The data origins in the producing stack so a cross stack ref is created. the function doesn't live in any state so it's pulled in the using stack. The only issue I could see is when using it with `Fn.timestamp`, but that seems like an artificial use-case to me. The alternative would be to create a terraform local when cross stack referencing (either in general or when inside a TF function)

The output of the producing stack should contain the fully resolved value of Terraform. That a function is involved is not a concern of the consuming stack.

Another thing which came to my mind: What happens if there are multiple consuming stacks for one reference, is this tested somewhere?

skorfmann:
>  also seems a bit far fetched to pass into another stack's output.

Yes, for now - would be awesome to have at some point though

skorfmann:
> > Currently, this.myFqdn creates an output only when referenced in a dependent stack.
> 
> Yeah that's the plan. Outputs are only created for referenced attributes.

What happens if I have deployed a producing and consuming stack. After the deployment, I remove the reference and hence the automatically created output. I have a hunch that I might run into problems planning / deploying the consuming stack once the output of the producing stack is gone.

DanielMSchmidt:
> The output of the producing stack should contain the fully resolved value of Terraform. That a function is involved is not a concern of the consuming stack.

Is that more of a general guideline or would you say that's important in the first iteration? We could do a follow-up issue for this, would that work for you?

DanielMSchmidt:
> Another thing which came to my mind: What happens if there are multiple consuming stacks for one reference, is this tested somewhere?

It's complicated :D This is the test case: https://github.com/hashicorp/terraform-cdk/pull/1416/files#diff-0102ed5920d6ff5abbb7ab623415cf2ebb76f07e1fd62bdea962a459a324103bR539

It's not fully fully working, as you can see it's using the same resource, but not the same `ref` by itself. This is due to a bug I discovered: https://github.com/hashicorp/terraform-cdk/issues/1447 I scoped it out since the same bug applies currently as well and it has nothing to do with cross stack. (I verified the cross stack part to be working, it was just adding or lacking tfExpression parenthesis)

DanielMSchmidt:
> What happens if I have deployed a producing and consuming stack. After the deployment, I remove the reference and hence the automatically created output. I have a hunch that I might run into problems planning / deploying the consuming stack once the output of the producing stack is gone.

That's a great point, @ansgarm brought that up as well, this is sth the AWS CDK struggles with. I created a test to check and it does not cause problems for TF: https://github.com/hashicorp/terraform-cdk/pull/1416/files#diff-99c6047d4d933587477aa2e3622ea8bc6fa07034cc9b32861c3f3c07d74ba734R81 and here is the main.ts for that test: https://github.com/hashicorp/terraform-cdk/pull/1416/files#diff-5162569d3dc22ac1cf854a3aaaf4b4cc41eb9a7ece4f578006ce94b6fcb321ceR140

ansgarm:
> > What happens if I have deployed a producing and consuming stack. After the deployment, I remove the reference and hence the automatically created output. I have a hunch that I might run into problems planning / deploying the consuming stack once the output of the producing stack is gone.
> 
> That's a great point, @ansgarm brought that up as well, this is sth the AWS CDK struggles with. I created a test to check and it does not cause problems for TF: https://github.com/hashicorp/terraform-cdk/pull/1416/files#diff-99c6047d4d933587477aa2e3622ea8bc6fa07034cc9b32861c3f3c07d74ba734R81 and here is the main.ts for that test: https://github.com/hashicorp/terraform-cdk/pull/1416/files#diff-5162569d3dc22ac1cf854a3aaaf4b4cc41eb9a7ece4f578006ce94b6fcb321ceR140

@skorfmann @DanielMSchmidt 
I think a main difference that cause trouble in CloudFormation for this is the rollback that CloudFormation attempts if something fails.
So for example:
1. Deploy Stack A (v1, with output for Stack B) successfully
2. Deploy Stack B (v1, depends on stack A) successfully
3. Remove dependency between stacks
4. Deploy Stack A (v2, **without** output for Stack B) successfully
5. Deploy Stack B (v2, **without** dependency on Stack A) **fails** (due to some unrelated error, does not really matter why)
6. CloudFormation rolls back Stack B to v1 which **fails** because there's no output in Stack A v2

In Terraform there is not automatic rollback in case of an error, so this case won't happen. Instead the user can fix and retry deploying Stack B.

DanielMSchmidt:
@skorfmann does this test case cover your concern around complex computed lists or did I do sth wrong here? https://github.com/hashicorp/terraform-cdk/pull/1416/commits/103a5be6c14ac4065b6627ac05e096bcad6dd811

skorfmann:
> > What happens if I have deployed a producing and consuming stack. After the deployment, I remove the reference and hence the automatically created output. I have a hunch that I might run into problems planning / deploying the consuming stack once the output of the producing stack is gone.
> 
> That's a great point, @ansgarm brought that up as well, this is sth the AWS CDK struggles with. I created a test to check and it does not cause problems for TF: https://github.com/hashicorp/terraform-cdk/pull/1416/files#diff-99c6047d4d933587477aa2e3622ea8bc6fa07034cc9b32861c3f3c07d74ba734R81 and here is the main.ts for that test: https://github.com/hashicorp/terraform-cdk/pull/1416/files#diff-5162569d3dc22ac1cf854a3aaaf4b4cc41eb9a7ece4f578006ce94b6fcb321ceR140

The links appear to be broken. But anyway, if it works that's ok.

skorfmann:
> > The output of the producing stack should contain the fully resolved value of Terraform. That a function is involved is not a concern of the consuming stack.
> 
> Is that more of a general guideline or would you say that's important in the first iteration? We could do a follow-up issue for this, would that work for you?

I think it would be good to have this properly working from the beginning, seems very unexpected to me.

skorfmann:
> @skorfmann does this test case cover your concern around complex computed lists or did I do sth wrong here? [103a5be](https://github.com/hashicorp/terraform-cdk/commit/103a5be6c14ac4065b6627ac05e096bcad6dd811)

It's not really straightforward to understand what's actually being tested there. Here's my case: I was passing the entire [domain validation options](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/acm_certificate#domain_validation_options) object (ComplexComputedList), not a single element. Not even sure if that's a thing which should be working or which should be caught with an error.

skorfmann:
> > I wonder if this is a real use-case though, wouldn't folks use the attributes of a resource? I think this is never a good idea and would argue for disallowing passing these into outputs and terraform functions in general
> 
> Passing complete resources into an output is definitely valid. Terraform accepts that and there was a ticket fairly recently where a user requested the functionality and we made it work. It's not possible to pass to any other resource though and also seems a bit far fetched to pass into another stack's output.

So, this should raise an error when referencing across stacks then, correct?

skorfmann (approved):
Haven't gone through the code since others did this already. As laid out in the comments I'd add a few more explicit errors for things which are not working (yet). Plus, the functions magically appearing in a referencing stack should be fixed. Other than that, I think it's good and works as expected 👍

DanielMSchmidt:
I'm going to create follow up issues for the outstanding work and mention them here 👍

DanielMSchmidt:
- Fn should be saved in origin stack: https://github.com/hashicorp/terraform-cdk/issues/1480
- Passing complex objects / resources across stacks: https://github.com/hashicorp/terraform-cdk/issues/1481

dhmistry3:
Is there an example of how to do this in python?

DanielMSchmidt:
We have a test case in Python, hope that helps :) https://github.com/hashicorp/terraform-cdk/blob/main/test/python/cross-stack-references/main.py

dhmistry3:
@DanielMSchmidt Thanks that is helpful.  However, I am trying to pass an aws ARN from one stack to another.  When i set it in one stack and reference it in another, the variable gets set as 

${module.s3_bucket_id.s3_bucket_arn}

and then i get an error of 

│ No module call named "s3_bucket_id" is declared in the root module.

DanielMSchmidt:
@dhmistry3 Could you create an issue for this with a reproducible example so that I can take a look?

PR #1436, opened 2021-12-27
fix: Update Remote State Outputs documentation

adilatilgan:
You can access to RemoteState by id of the Output which is `random-pet` in this case.


jsteinich (approved):
Thanks for fixing this.

DanielMSchmidt (approved):
Awesome, thank you 👍

PR #1456, opened 2022-01-05
feat(cli): dont throw in get if there are no providers and modules are specified

DanielMSchmidt:
This allows us to just chain get after install or prebuilt making it more seemless to integrate.

E.g. one could have this package.json

```json
{
  "postinstall": "cdktf get"
}
```

with no need to think about this step explicitly. Before if you'd replace your last provider with a prebuilt one you would have a failing CI.


skorfmann:
> Before if you'd replace your last provider with a prebuilt one you would have a failing CI.

Which one could and should just remove as a step in the CI then? Are there other cases where this is relevant?

ansgarm:
> Which one could and should just remove as a step in the CI then?

I personally wouldn't want to change CI whenever I add or remove the last provider/module from the config. I'd prefer a no-op with a warning/info to an error.

DanielMSchmidt:
> Which one could and should just remove as a step in the CI then?

Yeah or it could just work :D IMHO one could add the cdktf get command in the TS template as prebuild step and it would just auto-magically work. 

Plus, from my point of view we loose nothing, there is still a warning telling you that cdktf get was not doing anything and where to put your provider / module requirements

skorfmann:
The case I have in mind: e.g. for the pre built providers, we run `get` to build the providers explicitly (via the [fetch](https://github.com/hashicorp/cdktf-provider-aws/blob/c449c8a360a107e65eed822dca50ef5694ac874b/package.json#L31) task in this [workflow](https://github.com/hashicorp/cdktf-provider-aws/blob/c449c8a360a107e65eed822dca50ef5694ac874b/.github/workflows/provider-upgrade.yml#L20)). So, we run `get` in the expectation that there is something to build, and if it can't be build for whatever reason, I'd expect an error which stops the pipeline. I'd argue, that this is a common assumption for users who are running `get`.


> > Which one could and should just remove as a step in the CI then?
> 
> I personally wouldn't want to change CI whenever I add or remove the last provider/module from the config. I'd prefer a no-op with a warning/info to an error.

I'd consider this as a rare edge case and wouldn't optimize for it.

> > Which one could and should just remove as a step in the CI then?
> 
> Yeah or it could just work :D IMHO one could add the cdktf get command in the TS template as prebuild step and it would just auto-magically work.

I'd prefer to be intentional about this. Either you want to build stuff via `get` or you don't.

> 
> Plus, from my point of view we loose nothing, there is still a warning telling you that cdktf get was not doing anything and where to put your provider / module requirements

We change the behaviour by returning a different exit code.

DanielMSchmidt:
> I'd prefer to be intentional about this. Either you want to build stuff via get or you don't.

I think we have a different semantic meaning for `get` here. From my point of view it's similar to an `npm install`, I would translate it to "get everything I specified" and if I have nothing specified yet it does not need to do anything and that's fine.

It seems like for you it's more similar to an `npm install this-package` type of operation where an absent package is a problem.


> I'd consider this as a rare edge case and wouldn't optimize for it.

I have run into this enough times to be annoyed enough to do a PR. For me being able to add get to a prebuilt script is being able to completely forget about it during dev, making the experience way more seemless. This prevents me from doing this without some weird `cdktf get || echo "nothing to do"` hacks.


> We change the behaviour by returning a different exit code.

Yeah we do, my point is not that this change does nothing. I perceived the "warn a user we didn't detect a provider or module" mission as the value the code ships and the exit code being not the main part here. Seems like I have been mistaken.

skorfmann:
Well, I'm still not really convinced that this change is worthwhile - in particular with #1084 on the horizon. However, for the pre-built generator workflow case we need additional guardrails anyway. So, feel free to merge.

ansgarm (commented):
I think we should mention the behavior change in the changelog

ansgarm (approved):

PR #1463, opened 2022-01-06
chore: remove depreciation warning for `fs.rmdir`

rirze:
I believe this is a Node16 depreciation message:
``` shell
(node:22795) [DEP0147] DeprecationWarning: In future versions of Node.js, fs.rmdir(path, { recursive: true }) will be removed. Use fs.rm(path, { recursive: true }) instead
(Use `node --trace-deprecation ...` to show where the warning was created)

```

I encountered it today when `cdktf` deleted an old stack that was unused. It performed fine, but threw out this error.

### DIFF
--- a/packages/cdktf-cli/bin/cmds/helper/synth-stack.ts
+++ b/packages/cdktf-cli/bin/cmds/helper/synth-stack.ts
@@ -130,7 +130,7 @@ Command output on stdout:
     );
 
     for (const orphanedDirectory of orphanedDirectories) {
-      fs.rmdirSync(orphanedDirectory, { recursive: true });
+      fs.rmSync(orphanedDirectory, { recursive: true });
     }
 
     return stacks;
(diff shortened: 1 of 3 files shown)


DanielMSchmidt (commented):
Thanks for your help! Looks like a great fix for the warning. One problem with it is that we currently support node 10 and rmSync was only added in node 12. This means we either increase the node version requirement or add a fallback in case rmSync is not defined

ansgarm (approved):
Thank you, @rirze!
I just checked and we already require Node v12 (we even have [this check](https://github.com/hashicorp/terraform-cdk/blob/c4222503c4571b4b368fea8c4921ed1d5ab8b899/packages/cdktf-cli/bin/cmds/helper/check-environment.ts#L55)).
I created https://github.com/hashicorp/terraform-cdk/pull/1466 to fix the parts which @DanielMSchmidt mentioned.

ansgarm:
(fyi @rirze: I just force pushed to your branch and prepended `chore: ` to your commit message to satisfy our commit linter)

DanielMSchmidt (approved):

rirze:
awesome, thanks for taking care of that for me!

ansgarm:
No, thank _you_ for fixing this, @rirze 😄

PR #1467, opened 2022-01-07
fix(lib): don't treat strings as maps

DanielMSchmidt:
Solves a bug where passing a reference value to a list of objects destroyed the reference


packages/cdktf/test/runtime.test.ts
```diff
@@ -0,0 +1,86 @@
...
+      const reference = ref("some_resource.my_resource.some_attribute_array");
+
+      expect(
+        resolveExpression(listMapper(identity)(reference))
+      ).toMatchInlineSnapshot(
+        `"\${some_resource.my_resource.some_attribute_array}"`
+      );
+    });
+
+    it("leaves directly passed references in tact with a tokenized list", () => {
```
ansgarm, line 55:
```suggestion
    it("leaves directly passed references intact with a tokenized list", () => {
```

ansgarm (approved):

Issue #993, opened 2021-09-10
Replace `ComplexComputedList` with native array with a single `ComplexComputedObject`


DanielMSchmidt:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Currently we have a very limited way of working with lists as outputs of providers and the things we support make it clear that we don't deal with a native array but something else.
To make lists behave seamlessly we need to make them first class citizens by using native arrays.

To implement this properly we need to 

- return a list of Complex Elements from the getter, where
  - each native property is a token
  - each computed property is readonly
  - each complex property with computed attributes adds a `put${propertyName}` method to the element that takes an object with only the not-computed types as a property
- Enhance the resolve workflow to to recognize complex lists and resolve them properly
- Make the list 1 element long and throw an error if that element is used directly (i.e. through accessing the array)

This also enables us to advance the work of https://github.com/hashicorp/terraform-cdk/pull/925 and expose computed types as well. The current problem there is that the interface expects a getter to have the type `DataKubernetesPersistentVolumeClaimSpecResource[]`, but in the current implementation it has `(index: string) => DataKubernetesPersistentVolumeClaimSpecResources`.

### What this should enable
Currently inputs that are passed to a resource share the same type with references to that same attribute. For example, if we pass a string, we get back a string (which is an token encoded reference, but still appears to be a string). This is true for almost all types but not for lists. We want to achieve this for lists which contain objects (nested blocks) instead of plain values (lists of numbers or strings already return tokens encoded as a list of number / string containing the reference to that attribute).

#### Referencing list arguments also works in other languages than TS
When a list exists which can be set as input by the user, referencing that list does not work in other languages than TS (see: https://github.com/hashicorp/terraform-cdk/issues/1371). This allows the following to work:

todo: in Go referencing a list, passing it to e.g. `Fn.length`

#### Passing a reference to a list works without "fqn" hack
Example in HCL
```hcl
data "aws_availability_zones" "azs" {
  state = "available"
}

resource "aws_subnet" "this" {
  # ...
  availability_zone_id = data.aws_availability_zones.azs.zone_ids[0]
}
[...]

PR #1468, opened 2022-01-07
feat(provider-generator): reference computed values nested in lists

### LINKED ISSUE #1332: API Gateway v2 Domain Name Resource: Missing Computed Attributes
### Community Note

### cdktf & Language Versions

**`cdktf` and `cdktf-cli`:** 0.7.0
**Node.js version:** 16.9.0

### Affected Resource(s)

`APIGatewayV2.Apigatewayv2DomainName`

### Debug Output

Varies depending on access pattern. If you're unable to get a reproduction, please leave a comment with the intended access pattern, and I'll happily attach this in a Gist.

### Expected Behavior

Target domain name / hosted zone ID is accessible from the `domainNameConfiguration` output.

### Actual Behavior

Target domain name / hosted zone ID is not accessible from the `domainNameConfiguration` attribute.

With an access pattern of `(apiGatewayDomainName.domainNameConfiguration as any)[0].targetDomainName`, the following error is produced:

```
    TypeError: Cannot read properties of undefined (reading
'targetDomainName')
        at new ApiStack (C:\path\to\repo\packages\infrastructure\dist\src\stacks\api-stack.js:138:26)
        at Object.<anonymous> (C:\path\to\repo\packages\infrastructure\dist\src\index.js:18:1)
        at Module._compile (node:internal/modules/cjs/loader:1101:14)
        at Object.Module._extensions..js
(node:internal/modules/cjs/loader:1153:10)
        at Module.load (node:internal/modules/cjs/loader:981:32)
        at Function.Module._load (node:internal/modules/cjs/loader:822:12)   
        at Function.executeUserEntryPoint [as runMain]
(node:internal/modules/run_main:79:12)
        at node:internal/main/run_main_module:17:47
```

### Steps to Reproduce

```typescript
...

const apiGatewayDomainName = new APIGatewayV2.Apigatewayv2DomainName(
  this,
  "api_gateway_domain_name",
  {
    domainName: apiDomain,
    domainNameConfiguration: {
      certificateArn: certificate.acmCertificateArnOutput,
      endpointType: "REGIONAL",
      securityPolicy: "TLS_1_2",
    },
  }
);

new Route53.Route53Record(this, "api_dns_record", {
  name: apiGatewayDomainName.domainName,
  zoneId: hostedZone.zoneId!,
  type: "A",
  alias: [
    {
      name: ???, // have tried (apiGatewayDomainName.domainNameConfiguration as any).targetDomainName
                 // and (apiGatewayDomainName.domainNameConfiguration as any)[0].targetDomainName
      zoneId: ???,
      evaluateTargetHealth: false,
    },
  ],
});
```

### Important Factoids

Nothing to note.

### References

Nothing to note.

DanielMSchmidt:
Resolves #1332


packages/cdktf/lib/complex-computed-list.ts
```diff
@@ -1,9 +1,13 @@
 import { Token } from "./tokens";
-import { ITerraformResource } from "./terraform-resource";
+import { IResolvable } from "./tokens/resolvable";
+
+export interface IInterpolatingParent {
```
jsteinich, line 4: I think we would be explicit about marking types that implement this interface.
One benefit of doing so is that it makes the types consuming this interface more jsii friendly.

packages/@cdktf/provider-generator/lib/get/generator/models/attribute-model.ts
```diff
@@ -63,7 +63,14 @@ export class AttributeModel {
   }
 
   public get isAssignable() {
-    return this.required || this.optional;
```
jsteinich, line 66: I think this check was easier to read as it was.

packages/cdktf/lib/complex-computed-list.ts
jsteinich: Also, not my favorite name, but not sure what else to call it.

jsteinich (changes requested):
This change really only addresses 1 & 3 from https://github.com/hashicorp/terraform-cdk/issues/1332#issuecomment-977475370.
That's an improvement, but it would be nice to just finish it off finally.

Also, regarding number 3, https://github.com/hashicorp/terraform-cdk/blob/52de5d42dcc38ade5fccdefe98d35447938a809e/packages/%40cdktf/provider-generator/lib/get/generator/emitter/struct-emitter.ts#L197 really should take `IInterpolatingParent`. Functionally it works, but we are masking the compile error with https://github.com/hashicorp/terraform-cdk/blob/52de5d42dcc38ade5fccdefe98d35447938a809e/packages/%40cdktf/provider-generator/lib/get/generator/emitter/attributes-emitter.ts#L121.

This does also generate some dangling types (since the 2nd case is not implemented).

DanielMSchmidt:
@jsteinich Number two is planned to be tackled by #993

packages/@cdktf/provider-generator/lib/get/generator/models/attribute-model.ts
```diff
@@ -63,6 +63,11 @@ export class AttributeModel {
   }
 
   public get isAssignable() {
+    if (this.computed) {
+      // only computed && optional attributes are assignable
+      // if a computed attribute is required, it is never assignable
```
jsteinich, line 68: Are you sure? That doesn't really make sense.

jsteinich (commented):
Just one remaining question. Otherwise it looks good.

packages/@cdktf/provider-generator/lib/get/generator/models/attribute-model.ts
ansgarm: That was the case we encountered with the snapshot tests – but yeah, it seems weird to have something computed & required.

jsteinich: Which snapshot? I just tested the provider-generator tests without this changes and didn't notice anything.
I just want to make sure we aren't masking a different issue.

ansgarm: It could have been the edge provider in which case we could've produced this weirdness ourselves 🤔 @DanielMSchmidt @skorfmann do you remember whether it was the edge provider? If that is the case, this change might not be needed and we should check the edge provider test case 😅

DanielMSchmidt: I am unsure, I'd be down to removing this. If we run into issues we will probably see them arise soon

ansgarm (approved):

PR #1471, opened 2022-01-09
feat(lib): number[] Tokens

jsteinich:
Dividing up #1299, adding token support for `number[]`.

Combines the concepts from `number` and `string[]` tokens by using a single element array using a subset of the available bits as a marker for the token.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,14 +2,17 @@
 
 **Breaking Changes**
 
-### Map Tokens [#1411](https://github.com/hashicorp/terraform-cdk/pull/1411)
+### Number[] Tokens [#1471](https://github.com/hashicorp/terraform-cdk/pull/1471)
 
-As part of an effort to use more native types, there are now tokens for maps of primitive values.
+As part of an effort to use more native types, there are now tokens for number[].
+There is now `Token.asNumberList()` which can be used to conver other values into number[].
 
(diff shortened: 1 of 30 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
+## Unreleased
+
+**Breaking Changes**
+
+### Number[] Tokens [#1471](https://github.com/hashicorp/terraform-cdk/pull/1471)
+
+As part of an effort to use more native types, there are now tokens for number[].
+There is now `Token.asNumberList()` which can be used to conver other values into number[].
```
DanielMSchmidt, line 8: I'd add that this is mostly used under the hood and end users probably don't have to care too much about it

DanielMSchmidt (commented):
Not sure if I read over it, but can we make sure we throw and error when a list number token is found outside a list? I'm not sure if we have a check for it with strings, but having more than one item in a tokenized list is also a potential bug isn't it?

jsteinich:
> Not sure if I read over it, but can we make sure we throw and error when a list number token is found outside a list? I'm not sure if we have a check for it with strings, but having more than one item in a tokenized list is also a potential bug isn't it?

Yep, I added that check at the same level as the existing string check: `DefaultTokenResolver.resolveNumberList`.

DanielMSchmidt (approved):

PR #1476, opened 2022-01-11
feat(lib): Create assets for local modules

### LINKED ISSUE #946: Incorrect module source path generated for local modules

jsteinich:
Automatically create a `TerraformAsset` for any local module as specified in https://github.com/hashicorp/terraform-cdk/issues/664#issuecomment-831092035 and #948

Closes #946 (Really already fixed)
Supersedes #961

Draft since need to update some tests

### DIFF
--- a/packages/cdktf/lib/terraform-module.ts
+++ b/packages/cdktf/lib/terraform-module.ts
@@ -4,9 +4,9 @@ import { TerraformProvider } from "./terraform-provider";
 import { deepMerge } from "./util";
 import { ITerraformDependable } from "./terraform-dependable";
 import { Token } from "./tokens";
-import * as path from "path";
 import { ref } from "./tfExpression";
 import { TokenMap } from "./tokens/private/token-map";
+import { TerraformAsset } from "./terraform-asset";
 
 export interface TerraformModuleOptions {
   readonly source: string;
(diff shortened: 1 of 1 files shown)


packages/cdktf/test/terraform-hcl-module.test.ts
```diff
@@ -2,21 +2,21 @@ import { Testing, TerraformStack, TerraformHclModule } from "../lib";
...
   new TerraformHclModule(stack, "test", {
-    source: "./foo",
+    source: "./test/fixtures/hcl-module/",
   });
   expect(Testing.synth(stack)).toMatchSnapshot();
 });
 
 test("pass variables", () => {
-  const app = Testing.app();
+  const app = Testing.fakeCdktfJsonPath(Testing.app());
```
DanielMSchmidt, line 15: I'd prefer this to be a flag, just because it's easier to discover

packages/cdktf/lib/terraform-asset.ts
```diff
@@ -49,7 +49,9 @@ export class TerraformAsset extends Resource {
     if (path.isAbsolute(config.path)) {
       this.sourcePath = config.path;
     } else {
-      const cdktfJsonPath = findFileAboveCwd("cdktf.json");
+      const cdktfJsonPath =
+        scope.node.tryGetContext("cdktfJsonPath") ??
```
DanielMSchmidt, line 53: Is this context just for testing or has it a real world use-case?

DanielMSchmidt (commented):
Just some nit picks, looks good so far :)

DanielMSchmidt (approved):

PR #1488, opened 2022-01-16
feat(lib): cross stack references other backends

### LINKED ISSUE #1451: Support Cross Stack References on different backends
### Community Note

### Description

Currently, cross stack references can't be done when the origin backend something else besides local and remote.
The goal of this issue is to have all backends support cross stack references.

### References

- Cross Stack References: https://github.com/hashicorp/terraform-cdk/issues/651
- Multi-Stack Deployment: https://github.com/hashicorp/terraform-cdk/issues/650

jsteinich:
Closes #1451 

- Added support to all currently supported backends.
- Fixed an issues with the local backend support.
- Added workspace support to all backends that support it.


packages/cdktf/lib/backends/cos-backend.ts
```diff
@@ -16,13 +16,14 @@ export class CosBackend extends TerraformBackend {
...
+    scope: Construct,
+    name: string,
     _fromStack: string
   ): TerraformRemoteState {
-    throw new Error(
-      "The CosBackend does not yet implement cross stack references"
-    );
+    return new DataTerraformRemoteStateCos(scope, name, {
+      ...this.props,
+      workspace: "${terraform.workspace}",
```
DanielMSchmidt, line 25: If there is any case where one would want to overwrite `workspace` there is no chance now since the spread comes before the `workspace` value. If that case doesn't exist we have no issue, if it does we should put the spread in as last element so it overwrites previous values.

packages/cdktf/test/app.test.ts
```diff
@@ -291,7 +291,7 @@ describe("Cross Stack references", () => {
       }
     );
   });
-  it("errors if cross stack references are used with unsupported backends", () => {
+  it.skip("errors if cross stack references are used with unsupported backends", () => {
```
DanielMSchmidt, line 294: I think we can remove this now :)

DanielMSchmidt (commented):
Nice 👍

packages/cdktf/lib/backends/cos-backend.ts
jsteinich: This method of overwriting `workspace` would be very obscure and only work in Typescript since the type of the `props` variable doesn't actually include `workspace`. If we wanted to provide overwrite support, we'd probably want it as an option at the stack level. I see that as more of a follow up task (assuming there is even demand for it).

For context, I added the workspace in since I currently have environments separated by workspaces in Terraform and am considering keep it that way when migrating to cdktf. I'm not certain that I'll go that way, but nice to have the option. I see two dimensions of stacks: one for different parts of the application and one for different environments. Since different environments are typically deployed at different times, switching Terraform workspace in-between seems viable.

packages/cdktf/test/app.test.ts
jsteinich: removed

packages/cdktf/lib/backends/cos-backend.ts
DanielMSchmidt: Ah that makes sense, so it's not related but adjacent :)

PR #1495, opened 2022-01-18
feat(cli): output command

### LINKED ISSUE #227: Output command
### Community Note

### Description

When a stack has no changes, the `apply` is omitted  and therefore no outputs are printed.

Two things here:

- `cdktf deploy` should always print outputs regardless of changes being present 
- introduce `cdktf output` to explicitly retrieve outputs only

DanielMSchmidt:
Previously when there was no change we did not show the outputs

Closes #227


packages/cdktf-cli/bin/cmds/helper/outputs.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export async function saveOutputs(
+  filePath: string,
+  outputs: NestedTerraformOutput,
+  includeSensitiveOutputs: boolean
+) {
+  fs.writeFileSync(
+    filePath,
+    JSON.stringify(
+      unpackTerraformOutput(outputs, includeSensitiveOutputs),
```
ansgarm, line 36: This will print `<sensitive>` if `includeSensitiveOutputs` is `false`, correct?
I'd rather have expected that sensitive outputs don't appear at all in the output file in this case. If an output is switched to a sensitive value (i.e. for a previously hard-coded password) it could cause trouble if the user does not notice the change in the contents of the file and deploys services based on the password "<sensitive>" 😅

packages/cdktf-cli/bin/cmds/output.ts
```diff
@@ -0,0 +1,51 @@
...
+      .option("app", {
+        default: config.app,
+        required: true,
+        desc: "Command to use in order to execute cdktf app",
+        alias: "a",
+      })
+      .option("output", {
+        default: config.output,
+        required: true,
+        desc: "Output directory",
```
ansgarm, line 27: I know this flag existed before (for all commands requiring a synth) but in this context I'd love to see the description improved a bit (and maybe also copy that for the others) because it might be confused as the directory for the `outputs-file` option.
Maybe something along the lines of "output directory for the synthesized Terraform config"?

packages/cdktf-cli/bin/cmds/helper/outputs.ts
```diff
@@ -0,0 +1,47 @@
...
+      null,
+      2
+    )
+  );
+}
+
+export function normalizeOutputPath(filePath: string) {
```
ansgarm, line 43: Would it make sense to also automatically add a `.json` suffix if the passed filename does not end with `.json` already? I have no strong opinion on this, just curious what you think.

packages/cdktf-cli/bin/cmds/ui/models/terraform.ts
```diff
@@ -49,6 +49,10 @@ export interface TerraformOutput {
   value: string | Record<string, unknown> | Array<any>;
 }
 
+export function isTerraformOutput(output: any): output is TerraformOutput {
+  return typeof output.value === "string";
```
ansgarm, line 53: What if `value` is `Record<string, unknown>` or `Array<any>` as allowed by the type definition?

packages/cdktf-cli/bin/cmds/ui/terraform-context.tsx
```diff
@@ -184,6 +186,52 @@ export const parseOutput = (str: string): DeployingResource[] => {
...
+    Object.keys(obj).length === 0 ||
+    Object.values(obj).every(
+      (v) => v === undefined || v === null || isObjectEmpty(v)
+    )
+  );
+};
+
+export const getConstructIdsForOutput = (
+  stackContent: Record<string, any>,
+  output: { [key: string]: TerraformOutput }
```
ansgarm, line 203: Commenting here but it affects more parts of the PR:
I'd advocate for using `outputs` as name if we handle multiple outputs and `output` only if we are talking about a single one. Currently we seem to use `output` for single and multiple ones which confuses me when reading the code. Luckily we have types that make this clear, but I'd still like the naming to reflect this, too.
As the Terraform command is called `output` I'd keep the name of our `Output` (root) react component the way it is but would rename the other one to `Outputs` as it renders multiple outputs.
I'm not having a particularly strong opinion about this, so if there are good reasons to keep it as is, I'm fine with that, too.
P.S. The Terraform codebase also calls the map [outputs](https://github.com/hashicorp/terraform/blob/main/internal/command/views/json/output.go#L21) 🤓😁

website/docs/cdktf/cli-reference/commands.mdx
```diff
@@ -462,3 +472,52 @@ Set the `CDKTF_LOG_LEVEL` environment variable to `all` and set `CDKTF_DISABLE_L
...
+      --outputs-file-include-sensitive-outputs  Whether to include sensitive outputs in the output file                                                                                             [boolean] [default: false]
+  -h, --help                                    Show help                                                                                                                                                            [boolean]
+```
+
+### `--outputs-file`
+
+The `--outputs-file` option allows you to specify a file where the stack outputs will be written as JSON.
+The JSONs structure is matching your construct structure, the name of each construct is used as a key.
+
+```json
```
ansgarm, line 512: If you change this to `jsonc` the `//`-comment should be rendered better

website/docs/cdktf/cli-reference/commands.mdx
```diff
@@ -462,3 +472,52 @@ Set the `CDKTF_LOG_LEVEL` environment variable to `all` and set `CDKTF_DISABLE_L
...
+  "MyStack": {
+    "MyConstructInstance": {
+      "MyOutput": "my-value",
+      "MySecretOutput": "my-secret-value" // <sensitive> if --outputs-file-include-sensitive-outputs is not set
+    }
+  }
+}
+```
+
+Please be aware that sensitive outputs are only written to the file if the `--outputs-file-include-sensitive-outputs` option is used.
```
ansgarm, line 523: nit: Currently they are written but their value is masked 😉

packages/cdktf-cli/bin/cmds/helper/outputs.ts
DanielMSchmidt: Yeah that makes sense to me 👍

packages/cdktf-cli/bin/cmds/helper/outputs.ts
DanielMSchmidt: I'd vote for predictability and therefore not add anything :)

ansgarm: Sounds good 👍

packages/cdktf-cli/bin/cmds/ui/components/outputs.tsx
```diff
@@ -0,0 +1,31 @@
+import React from "react";
+import { Text, Box } from "ink";
+import { TerraformOutput } from "../models/terraform";
+
+export interface OutputsConfig {
+  output: { [key: string]: TerraformOutput };
```
ansgarm, line 6: Found one `output` which could be renamed to `outputs` 😁

packages/cdktf-cli/bin/cmds/ui/deploy.tsx
```diff
@@ -177,23 +164,28 @@ interface DeployConfig {
...
   targetDir,
   targetStack,
   synthCommand,
   autoApprove,
+  onOutputsRetrieved,
+  outputsPath,
 }: DeployConfig): React.ReactElement => {
   const {
-    state: { status, currentStack, errors, plan },
+    state: { status, currentStack, errors, plan, outputs: output },
```
ansgarm, line 180: We could also keep it as `outputs` here, right?

packages/cdktf-cli/bin/cmds/ui/models/terraform.ts
```diff
@@ -49,6 +49,10 @@ export interface TerraformOutput {
   value: string | Record<string, unknown> | Array<any>;
 }
 
+export function isTerraformOutput(output: any): output is TerraformOutput {
+  return typeof output === "object" && typeof output.value === "string";
```
ansgarm, line 53: But `output.value` is still allowed to be an array which would not return true here?

website/docs/cdktf/cli-reference/commands.mdx
```diff
@@ -159,6 +167,8 @@ Deploy an application with automatic approval of the diff (Terraform plan).
 $ cdktf deploy --auto-approve
 ```
 
+For more infos in the `--outputs-file` option, [see the `output` command below](/cdktf/cli-reference/commands#output).
```
ansgarm, line 170:
```suggestion
For more info on the `--outputs-file` option, [see the `output` command below](/cdktf/cli-reference/commands#output).
```

packages/cdktf-cli/bin/cmds/ui/models/terraform.ts
```diff
@@ -49,6 +49,14 @@ export interface TerraformOutput {
   value: string | Record<string, unknown> | Array<any>;
 }
 
+export function isTerraformOutput(output: any): output is TerraformOutput {
+  return (
+    typeof output === "object" &&
+    typeof output.sensitive === "boolean" &&
+    typeof output.type === "string"
```
ansgarm, line 56: Those are always set, right?

packages/cdktf-cli/test/ui/deploy.test.tsx
```diff
@@ -55,7 +55,7 @@ test("Output", async () => {
     },
   };
 
-  const { lastFrame } = render(<Output output={output} />);
+  const { lastFrame } = render(<Outputs output={output} />);
```
ansgarm, line 58: The prop needs to be `outputs` now, right?

ansgarm (approved):

packages/cdktf-cli/bin/cmds/ui/models/terraform.ts
DanielMSchmidt: Hmm now that you mention it, I trusted on the type system, but let me verify

PR #1499, opened 2022-01-20
feat: Model ComplexComputedLists as ComplexLists and ComputedObjects

### LINKED ISSUE #993: Replace `ComplexComputedList` with native array with a single `ComplexComputedObject`

ansgarm:
Closes #993

## Open Subtasks

- [ ] Write upgrade guide (ucloud examples might be helpful to link)
- [ ] Add link to upgrade guide to Changelog

### DIFF
--- a/packages/@cdktf/provider-generator/lib/get/__tests__/generator/__snapshots__/complex-computed-types.test.ts.snap
+++ b/packages/@cdktf/provider-generator/lib/get/__tests__/generator/__snapshots__/complex-computed-types.test.ts.snap
@@ -94,10 +94,9 @@ export class AcmCertificateOptionsOutputReference extends cdktf.ComplexObject {
   /**
   * @param terraformResource The parent resource
   * @param terraformAttribute The attribute on the parent resource this class is referencing
-  * @param isSingleItem True if this is a block, false if it's a list
   */
-  public constructor(terraformResource: cdktf.IInterpolatingParent, terraformAttribute: string, isSingleItem: boolean) {
-    super(terraformResource, terraformAttribute, isSingleItem);
+  public constructor(terraformResource: cdktf.IInterpolatingParent, terraformAttribute: string) {
+    super(terraformResource, terraformAttribute);
   }
 
   public get internalValue(): AcmCertificateOptions | undefined {
--- a/packages/@cdktf/provider-generator/lib/get/__tests__/generator/__snapshots__/complex-computed-types.test.ts.snap
+++ b/packages/@cdktf/provider-generator/lib/get/__tests__/generator/__snapshots__/complex-computed-types.test.ts.snap
@@ -49,7 +49,45 @@ export interface AcmCertificateConfig extends cdktf.TerraformMetaArguments {
   */
   readonly options?: AcmCertificateOptions;
 }
-export class AcmCertificateDomainValidationOptions extends cdktf.ComplexComputedList {
+export interface AcmCertificateDomainValidationOptions {
+}
+
+export function acmCertificateDomainValidationOptionsToTerraform(struct?: AcmCertificateDomainValidationOptions): any {
+  if (!cdktf.canInspect(struct) || cdktf.Tokenization.isResolvable(struct)) { return struct; }
+  if (cdktf.isComplexElement(struct)) {
+    throw new Error(\\"A complex element was used as configuration, this is not supported: https://cdk.tf/complex-object-as-configuration\\");
+  }
+  return {
+  }
+}
+
+export class AcmCertificateDomainValidationOptionsOutputReference extends cdktf.ComplexObject {
+  private isEmptyObject = false;
+
+  /**
+  * @param terraformResource The parent resource
+  * @param terraformAttribute The attribute on the parent resource this class is referencing
+  * @param complexObjectIndex the index of this item in the list
+  * @param complexObjectIsFromSet whether the list is wrapping a set (will add tolist() to be able to access an item via an index)
+  */
+  public constructor(terraformResource: cdktf.IInterpolatingParent, terraformAttribute: string, complexObjectIndex: number, complexObjectIsFromSet: boolean) {
+    super(terraformResource, terraformAttribute, complexObjectIndex, complexObjectIsFromSet);
+  }
+
+  public get internalValue(): AcmCertificateDomainValidationOptions | undefined {
+    let hasAnyValues = this.isEmptyObject;
+    const internalValueResult: any = {};
+    return hasAnyValues ? internalValueResult : undefined;
+  }
+
+  public set internalValue(value: AcmCertificateDomainValidationOptions | undefined) {
+    if (value === undefined) {
+      this.isEmptyObject = false;
+    }
+    else {
+      this.isEmptyObject = Object.keys(value).length === 0;
+    }
+  }
 
   // domain_name - computed: true, optional: false, required: false
   public get domainName() {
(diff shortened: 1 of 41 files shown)


ansgarm:
### Quick update on this PR
Today we (@skorfmann, @DanielMSchmidt and @jsteinich) had a deeper look into the changes that this PR would introduce. 
It models complex computed lists (i.e. "read-only" lists of objects that only have values after a resource has been provisioned) as an array which better reflects the underlying data. But as the length and contents of this array are only known "at runtime", they are modeled as arrays of a single token (similar to how e.g. string arrays are also modeled currently). While this allows accessing the whole list at once (which is currently not possible without self constructed references), it makes accessing a single item for a (at synth time) known index particularly hard. Which is an understatement. It is really bad. See the following two examples for yourselves:
```go
// with ComplexComputedLists (previously)
ImageId: images.Images(jsii.String("0")).Id(),
// with Arrays of ComplexListItems (this PR)
ImageId: cdktf.Token_AsString(cdktf.Fn_Lookup(cdktf.Fn_Element(images.Images, 0), jsii.String("id"), jsii.String("")), &cdktf.EncodingOptions{}),
```
TypeScript is less bad, but still terrible:
```ts
// with ComplexComputedLists (previously)
imageId: images.images("0").id,
// with Arrays of ComplexListItems (now)
imageId: Token.asString(Fn.lookup(Fn.element(images.images, 0), "id", "")),
```

As this does not makes things simpler (besides allowing to access a whole list directly via e.g. `images.Images`), we decided to revisit the implementation and make some adjustments to it.

We'd like a solution to still be able to autocomplete the properties on individual items (e.g. `.id`) via intellisense in editors while still being able to reference to the whole list.

Sidenote: Maps probably share a similar problem and we should also address that (separately).

packages/@cdktf/provider-generator/lib/get/generator/emitter/attributes-emitter.ts
```diff
@@ -114,11 +125,14 @@ export class AttributesEmitter {
...
+  // returns an invocation of a stored class, e.g. 'new DeplotmentMetadataOutputReference(this, "metadata")'
   private storedClassInit(att: AttributeModel) {
-    return `new ${att.type.name}${
-      att.type.isSingleItem ? "OutputReference" : ""
-    }(this, "${att.terraformName}")`;
+    if (att.type.isSingleItem) {
+      return `new ${att.type.name}OutputReference(this, "${att.terraformName}")`;
+    } else {
+      // list
+      return `new ${att.type.name}List(this, "${att.terraformName}", ${att.type.isSet})`;
```
DanielMSchmidt, line 134: Is this a list or an item in a list?

DanielMSchmidt (approved):
Looks good from the code perspective 👍

test/typescript/edge/main.ts
```diff
@@ -99,6 +99,17 @@ export class ReferenceStack extends TerraformStack {
       req: set.set,
       singlereq: { reqbool: true, reqnum: 1, reqstr: "reqstr" },
     });
+
+    // passing a list ref of a complex list type (no block) into an output
+    new TerraformOutput(this, "list_from_list_type_ref", {
```
jsteinich, line 104: Why not have this case in the other language tests?

packages/cdktf/lib/complex-computed-list.ts
```diff
@@ -136,29 +146,62 @@ export class ComplexComputedList extends ComplexComputedAttribute {
...
+
     return this.terraformResource.interpolationForAttribute(
-      `${this.terraformAttribute}${this.isSingleItem ? "[0]" : ""}.${property}`
+      `${this.terraformAttribute}[${this.complexObjectIndex}].${property}`
     );
   }
 
   protected interpolationAsList() {
     return this.terraformResource.interpolationForAttribute(
-      `${this.terraformAttribute}.*`
```
jsteinich, line 161: Should this stick around or otherwise be exposed? I'm not sure where it's used currently.

jsteinich (commented):
Looking good. I imagine there will be follow up PRs for maps and non-computed properties?

packages/@cdktf/provider-generator/lib/get/__tests__/generator/__snapshots__/complex-computed-types.test.ts.snap
```diff
@@ -49,7 +49,45 @@ export interface AcmCertificateConfig extends cdktf.TerraformMetaArguments {
   */
   readonly options?: AcmCertificateOptions;
 }
-export class AcmCertificateDomainValidationOptions extends cdktf.ComplexComputedList {
+export interface AcmCertificateDomainValidationOptions {
+}
+
+export function acmCertificateDomainValidationOptionsToTerraform(struct?: AcmCertificateDomainValidationOptions): any {
```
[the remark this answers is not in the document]
ansgarm: Good find 👍 Maybe there's a way we can skip this

test/typescript/edge/main.ts
ansgarm: Ah, good idea

packages/cdktf/lib/complex-computed-list.ts
ansgarm: It is used here (but just to check the type, not actually called): https://github.com/hashicorp/terraform-cdk/blob/b0d5a7e5871bc4a5ca9674137482f1e79596704c/packages/cdktf/lib/tokens/private/encoding.ts#L162
But yeah, maybe I should just leave it as is, even if we thing it might not be used any longer.

ansgarm: Reverted it.

packages/@cdktf/provider-generator/lib/get/__tests__/generator/__snapshots__/complex-computed-types.test.ts.snap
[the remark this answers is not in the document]
ansgarm: I created a [follow-up issue](https://github.com/hashicorp/terraform-cdk/issues/1626). I can address this in a separate PR as it does not directly affect users.

website/docs/cdktf/release/upgrade-guide-v0-10.mdx
```diff
@@ -0,0 +1,68 @@
...
+
+#### Typescript
+
+```ts
+// previously
+const firstItemId = resource.listAttribute("0").id;
+
+// new
+const firstItemId = resource.listAttribute.get("0").id;
+const firstItem = resource.listAttribute.get("0"); // now possible
```
jsteinich, line 23: Shouldn't these be numbers?

jsteinich (changes requested):

website/docs/cdktf/release/upgrade-guide-v0-10.mdx
ansgarm: Yes! Missed the TypeScript example

jsteinich (approved):

PR #1508, opened 2022-01-25
Unique resource names only within provider scope

ukautz:
This CL modifies the scope of uniqueness for generated resource (class / interface) names from "across all providers" to "within each provider".

This guarantees consistently generated resource names, independent of how many providers are used, and thereby fixes two issues of the current implementation:

## 1. Adding / removing providers invalidates code

Adding or removing providers, which have resources with the same name, requires users to change their own code, even if the code does not relate to the added or removed providers.

For example: using the [`hcloud`](https://registry.terraform.io/providers/hetznercloud/hcloud/latest/docs) and the [`digitalocean`](https://registry.terraform.io/providers/digitalocean/digitalocean/latest/docs) providers. Both have a `ssh_key` resource ([1](https://registry.terraform.io/providers/hetznercloud/hcloud/latest/docs/resources/ssh_key), [2](https://registry.terraform.io/providers/digitalocean/digitalocean/latest/docs/resources/ssh_key)). Using each of them individually generates that resource into a respective `SshKey` resource class. Using both of them together, renames one of them (the alphanumeric greater) into `SshKeyA` - unnecessarily, because either lives in a unique namespace already.

- Valid code, as long as only one (`hcloud`) provider is used: https://github.com/ukautz/cdktf-inconsistent-name-bug/blob/first-provider/main.ts#L13
- Updated code, that requires the use of a different name of the same class, once both providers are used: https://github.com/ukautz/cdktf-inconsistent-name-bug/blob/second-provider/main.ts#L15

## 2. Higher level constructs / modularization

Without having a guarantee that generated resource names are the same "for everyone" (independent of how many / what other providers used), it is not possible to create reusable, higher level constructs (i.e. constructs composed of constructs), that are build on the generated constructs.

- - - - 

**Note:** From what I can tell the uniqueness cannot be "broken" within one provider. However, I am not overly deep into the code and would trust someone with greater knowledge about the implementation to judge that better - and remove the uniqueness code. This CL just moves the scope.

However, if there is chance of multiple resources within one provider having the same name: The current implementation does not address that. It can handle only at most two occurences. Any third or subsequent would just get the same name as the second. Again, this CL does not address this, only the scope.


jsteinich:
@ukautz Thanks for putting this together. I believe you have the right idea, but not quite the right scope. A new instance of the `Parser` class is created for each resource within a provider. The next level up (`ResourceParser`) is shared for every provider generated, but I don't think it would be unreasonable to make a new one per provider (`TerraformProviderGenerator.emiProvider`).



> From what I can tell the uniqueness cannot be "broken" within one provider. However, I am not overly deep into the code and would trust someone with greater knowledge about the implementation to judge that better - and remove the uniqueness code. This CL just moves the scope.

While resource names are unique, there are many other generated types added. The other generated names can (and sometimes do) collide.

> However, if there is chance of multiple resources within one provider having the same name: The current implementation does not address that. It can handle only at most two occurences. Any third or subsequent would just get the same name as the second. Again, this CL does not address this, only the scope.

I don't believe there have been issues with more than two occurrences, but it wouldn't hurt to address it either.

ukautz:
Thanks @jsteinich for shedding some light on the details. Appreciate it. I will put some time into an updated PR in the coming week.

skorfmann:
@ukautz thanks for contributing this PR - please make sure to [sign the CLA](https://github.com/hashicorp/terraform-cdk/pull/1508#issuecomment-1021641448) as well.

DanielMSchmidt:
@ukautz Hey, thank you for this great PR! I'd really love to get it in and I'll help to make it green :) It would be great if you could sign the CLA, otherwise we can't use your code :(

DanielMSchmidt:
Thank you for raising the issue, closed by #1870

PR #1511, opened 2022-01-26
refactor: separate logic from UI layer

DanielMSchmidt:
## Summary

The goal of this PR is to separate our CLI rendering from the code that does the work. This makes each part easier to test and provides a stepping stone in the direction of a driving the CLI from node.js.


examples/typescript/docker/cdktf.json
```diff
@@ -3,5 +3,6 @@
   "app": "npx ts-node main.ts",
   "terraformProviders": [
     "kreuzwerker/docker@~> 2.0"
-  ]
+  ],
+  "projectId": "e9d26543-b972-4800-93c1-4fea0aec36f8"
```
ansgarm, line 7: I think we should remove this 🙈

packages/cdktf-cli/bin/cmds/ui/deploy.tsx
```diff
@@ -176,42 +97,51 @@ export const Deploy = ({
...
-    currentStack.name === "" ? (
-      <Text>{status}...</Text>
-    ) : (
-      <Text>
-        {status}
-        <Text bold>&nbsp;{currentStack.name}</Text>...
-      </Text>
-    );
+  if (error) {
+    return <ErrorComponent error={error} />;
```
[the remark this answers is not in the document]
DanielMSchmidt: There is a useEffect that exits with exitCode 1 if fatal is set

packages/cdktf-cli/lib/index.ts
```diff
@@ -0,0 +1,333 @@
...
+  PLANNED = "planned",
+  DEPLOYING = "deploying",
+  DESTROYING = "destroying",
+  OUTPUT_FETCHED = "output fetched",
+  DONE = "done",
+}
+
+export type ProjectUpdates =
+  | {
+      type: "synthing";
```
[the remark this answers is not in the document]
DanielMSchmidt: No reason, just confusion :D

packages/cdktf-cli/bin/cmds/ui/deploy.tsx
```diff
@@ -230,29 +160,103 @@ export const Deploy = ({
         )}
       </>
     );
+  }
 
-  return (
-    <Box>
-      {isPlanning ? (
+  switch (projectUpdate?.type) {
```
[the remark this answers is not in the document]
DanielMSchmidt: Yeah not 100 percent ideal, but I guess this will be gone soon anyways :D

packages/cdktf-cli/bin/cmds/ui/deploy.tsx
```diff
@@ -230,29 +160,103 @@ export const Deploy = ({
...
-          {isConfirmed && <Apply outputsPath={outputsPath} />}
-        </>
-      )}
-    </Box>
-  );
+      );
+    }
+    default:
+      return (
+        <Text>{`Trying to render unknown state: ${projectUpdate?.type}`}</Text>
```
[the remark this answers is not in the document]
DanielMSchmidt: I rephrased it 👍

packages/cdktf-cli/bin/cmds/ui/diff.tsx
```diff
@@ -45,39 +46,46 @@ const PlanSummary = ({ resources }: PlanSummaryConfig): React.ReactElement => {
...
 
-export const CloudRunInfo = (): React.ReactElement => {
-  const { url } = useTerraformState();
-  if (url == undefined) return <></>;
+export const CloudRunInfo = ({ url }: { url?: string }): React.ReactElement => {
+  if (url === undefined) return <></>;
 
   const staticElements = [url];
 
+  // TODO: create follow up issue to use https://github.com/sindresorhus/ink-link
```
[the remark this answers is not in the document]
DanielMSchmidt: https://github.com/hashicorp/terraform-cdk/issues/1578

packages/cdktf-cli/lib/get.ts
```diff
@@ -0,0 +1,46 @@
...
+export async function get({
+  constructsOptions,
+  constraints,
+  onUpdate,
+  reportTelemetry,
+}: GetConfig) {
+  await fs.remove(constructsOptions.codeMakerOutput);
+  const constructsMaker = new ConstructsMaker(
+    constructsOptions,
+    constraints as config.TerraformDependencyConstraint[], // ConstructsMaker handles both string and extended form, but is not consistent type wise
```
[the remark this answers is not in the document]
DanielMSchmidt: Yeah, we could refactor it, not sure if it helps anyone though? I mean it improves the typesafety but no effect besides

packages/cdktf-cli/lib/project-execution.ts
```diff
@@ -0,0 +1,448 @@
...
+  ) => event.data.needsApply === false,
+};
+
+const projectExecutionMachine = createMachine<ProjectContext, ProjectEvent>(
+  {
+    id: "project",
+    initial: "idle",
+    context: {
+      onProgress: () => {}, // eslint-disable-line @typescript-eslint/no-empty-function
+      // Mandantorily set by withContext
```
ansgarm, line 248:
```suggestion
      // Mandatorily set by withContext
```

packages/cdktf-cli/lib/project-execution.ts
```diff
@@ -0,0 +1,448 @@
...
+      targetAction: "synth",
+      synthCommand: "",
```
ansgarm, line 250: Are these two required? They are optional in the `ProjectContext` type

packages/cdktf-cli/lib/project-execution.ts
```diff
@@ -0,0 +1,448 @@
...
+      JSON.parse(stack.content),
+      outputs
+    );
+
+    return Promise.resolve({ outputs, outputsByConstructId });
+  },
+};
+
+const guards = {
+  onTargetAction: (
```
ansgarm, line 230: I had a hard time grasping what this does (when reading the state machine config). Could this be improved somehow to create a more speaking name when using the guard?
Maybe something like `targetActionIs`?
That way the usage could read (with reordered lines):
```
cond: {
  name: "targetActionIsSynth",
  type: "targetActionIs",
  value: "synth",
}
```

packages/cdktf-cli/lib/project-execution.ts
```diff
@@ -0,0 +1,448 @@
...
+            actions: assign({
+              message: (_context, event) => event.data,
+            }),
+          },
+          onDone: {
+            target: "gatherOutput",
+          },
+        },
+      },
+      gatherOutput: {
```
ansgarm, line 417: I'd like plural more as that differentiates it from "output" as in cli / shell output and brings "Terraform outputs" closer 🙂
```suggestion
      gatherOutputs: {
```

packages/cdktf-cli/lib/project-execution.ts
```diff
@@ -0,0 +1,448 @@
...
+                value: "diff",
+              },
+            },
+
+            {
+              actions: assign({
+                targetStackPlan: (_context, event) => event.data,
+              }),
+              target: "gatherOutput",
+              cond: "planNeedsNoApply",
```
ansgarm, line 339: Why is a diff printing outputs in this case but not in the case above? Is there any magic to `planNeedsNoApply` (such as only being present if a deploy was the goal and absent in case of a plan?)

ansgarm: Ah, wait. As the above case catches the diff target, it will end there (without printing outputs).
This will hence only print outputs if the target was not diff.
Can we make this more clear?

Does it make sense to run `gatherOutput` if the target is `destroy`? (It would run that, wouldn't it?)

packages/cdktf-cli/lib/project-execution.ts
```diff
@@ -0,0 +1,448 @@
...
+            {
+              actions: assign({
+                targetStackPlan: (_context, event) => event.data,
+              }),
+              target: "waitingForApproval",
+            },
+          ],
+        },
+      },
+      waitingForApproval: {
```
ansgarm, line 357: nit: This state waits for approval to either deploy or destroy, right? Can we make this more clear somehow? Do we need to? Probably not, just wanted to share my confusion 😁 Feel free to just resolve this after reading ^^

packages/cdktf-cli/lib/project-execution.ts
ansgarm: Ah, btw: You could also make deploy ignore a plan that needs no deploy (no-op) and then remove this check. I don't know which strategy is better, however. No idea at all, might as well stay as is. State Machine would be less complicated but also less explicit 🤷🤷🤷🤷

packages/cdktf-cli/lib/project-execution.ts
```diff
@@ -0,0 +1,448 @@
...
+            target: "error",
+            actions: assign({
+              message: (_context, event) => event.data,
+            }),
+          },
+          onDone: {
+            target: "gatherOutput",
+          },
+        },
+      },
```
ansgarm, line 401: I really like how short this is 👏

packages/cdktf-cli/lib/index.ts
```diff
@@ -0,0 +1,338 @@
...
+          onUpdate({
+            type: "waiting for approval",
+            stackName: this.stackName!,
+            plan: this.currentPlan!,
+            approve: () => {
+              this.stateMachine.send("APPROVAL_GIVEN");
+            },
+            reject: () => {
+              this.stateMachine.send("APPROVAL_ABORTED");
+            },
```
ansgarm, line 267: Love this!

packages/cdktf-cli/lib/models/terraform-cli.ts
```diff
@@ -82,16 +109,26 @@ export class TerraformCli implements Terraform {
       terraformBinaryName,
       ["destroy", "-auto-approve", "-input=false"],
       { cwd: this.workdir, env: process.env },
-      stdout
+      (buffer: Buffer) => {
+        this.onStdOut(buffer);
+        stdout(buffer);
+      },
+      this.onStdErr
```
ansgarm, line 116: Never noticed this before, but why don't we supply the `-json` option (or whatever it is called) here? Do we even do that somewhere?

packages/cdktf-cli/lib/models/terraform-cloud.ts
```diff
@@ -116,7 +116,9 @@ export class TerraformCloud implements Terraform {
   constructor(
     public readonly stack: SynthesizedStack,
     public readonly config: TerraformJsonConfigBackendRemote,
-    isSpeculative = false
+    isSpeculative = false,
+    // TODO: find out how to forward logs from terraform cloud (only public to ignore ts error)
```
ansgarm, line 120: todo left -> follow up issue?

packages/cdktf-cli/lib/project-execution.ts
```diff
@@ -0,0 +1,448 @@
...
+      done: {
+        type: "final",
+      },
+      error: {
+        type: "final",
+      },
+    },
+  },
+  { services, guards: guards as any }
```
ansgarm, line 445: Is `as any` really necessary? Can't this be typed?

ansgarm (commented):
Finally a full review, really looking forward to merging this refactoring soon!

packages/cdktf-cli/lib/index.ts
```diff
@@ -0,0 +1,338 @@
...
+        `ProjectExecution State machine transitions: ${
+          lastState || "unknown"
+        } -> ${this.currentState}`
+      );
+      const ctx = state.context;
+
+      switch (lastState) {
+        case "synth":
+          this.stacks = ctx.synthesizedStacks || [];
+          this.status = Status.SYNTHESIZED;
```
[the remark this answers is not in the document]
DanielMSchmidt: Yeah on second thought the status field doesnt make too much sense, I removed it entirely

packages/cdktf-cli/lib/models/terraform-cli.ts
```diff
@@ -72,7 +95,11 @@ export class TerraformCli implements Terraform {
         ...(planFile ? [planFile] : []),
       ],
       { cwd: this.workdir, env: process.env },
-      stdout
+      (buffer: Buffer) => {
+        this.onStdOut(buffer);
+        stdout(buffer);
```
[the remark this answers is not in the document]
DanielMSchmidt: One is on the global level for forwarding the other on the per call level for parsing. While we could simplify this we will get rid of it in one of the next PRs so I didn't optimize here

packages/cdktf-cli/test/lib/index.test.ts
```diff
@@ -0,0 +1,243 @@
...
+        expect.arrayContaining([
+          "synthesizing",
+          "synthesized",
+          "planning",
+          "planned",
+          // "deploying", // WHYYYY??
+          "deployed",
+        ])
+      );
+      return expect(eventTypes.includes("waiting for approval")).toBeFalsy();
```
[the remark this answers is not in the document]
DanielMSchmidt: Yeah it would never happen. I think the state machine could not possibly not halt and have the test timeout

packages/cdktf-cli/lib/project-execution.ts
DanielMSchmidt: From my point of view all "workflow" related logic should live in the state machine. I'd say the functions we call have too much logic right now, they should be as dumb as possible. But I didn't want to overstretch it, I rather look at it when we work on the upcoming features

packages/cdktf-cli/lib/models/terraform-cli.ts
DanielMSchmidt: no seems like we don't

packages/cdktf-cli/lib/project-execution.ts
DanielMSchmidt: yes, the xstate types are lacking, they can't handle custom guards right now

packages/cdktf-cli/lib/models/terraform-cloud.ts
DanielMSchmidt: Nope, big chunk of work missing from this PR 🙈

test/typescript/providers/test.ts
```diff
@@ -1,118 +1,129 @@
...
-      expect(stack.byId("lambdaFn")).toBeTruthy();
-    });
-
-    test("has data source", () => {
-      expect(stack.byId("ami")).toBeTruthy();
-      expect(stack.byId("callerIdentity")).toBeTruthy();
+  describe("deploy", () => {
+    test("build providers", async () => {
+      // We have an arcane bug where the deploy is invoked twice although it's just called once.
+      // This is a workaround for that, by havig a new driver every run we don't have this bug.
```
ansgarm, line 7:
```suggestion
      // This is a workaround for that, by having a new driver every run we don't have this bug.
```

packages/cdktf-cli/test/lib/index.test.ts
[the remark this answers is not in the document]
ansgarm: Should we add an
```ts
if (event.type === "waiting for approval") {
  event.approve();
}
```
then?

packages/cdktf-cli/lib/index.ts
```diff
@@ -0,0 +1,338 @@
...
+              (r: PlannedResource) =>
+                Object.assign({}, r, {
+                  applyState: DeployingResourceApplyState.WAITING,
+                })
+            );
+          }
+          onUpdate({
+            type: "planned",
+            stackName: ctx.targetStack!,
+            plan: ctx.targetStackPlan!,
```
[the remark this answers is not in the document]
ansgarm: @DanielMSchmidt have you addressed this?

ansgarm (approved):
some small things still open, but overall approve ✅ Nice work, @DanielMSchmidt! Can't repeat myself too much saying this 👏

ansgarm (approved):

PR #1522, opened 2022-01-27
chore(docs): remove old layouts folder

schersh:
The new navigation file is now located under the /data directory. We should remove the old /layouts directory to prevent confusion and ensure people will update the correct nav file. 

@dstaley Let me know if there is any reason we need to keep the /layouts directory.


laurapacilio (approved):
Yay - thank you! having all of these here was confusing! Quick question - I see we're deleting the files out of the directory - do we need to delete the actual folder too, or will that automatically happen once we do this?

ansgarm:
> do we need to delete the actual folder too, or will that automatically happen once we do this?

@laurapacilio That'll happen automatically as Git does not track directories (that's why sometimes there are directories with an empty `.gitkeep` file which is commonly used as a workaround).

dstaley (approved):

PR #1531, opened 2022-02-01
fix(lib): Add missing regex Function

bollohz:


jsteinich:
Thanks for this.
I wonder if the simplified return type will cause any problems (Terraform documentation mentions variable return types depending on pattern). This does match what we have for `regexAll` which has similar semantics, so likely ok here as well.

DanielMSchmidt (approved):

PR #1533, opened 2022-02-02
feat(lib): add validation block in tf variables

bollohz:
This PR add `validation` block inside Terraform Variables.


packages/cdktf/lib/terraform-variable.ts
```diff
@@ -51,6 +51,11 @@ export abstract class VariableType {
   }
 }
 
+export interface TerraformVariableValidationConfig {
+  readonly condition: any;
+  readonly error_message: string;
```
ansgarm, line 56: I think we'd want to change this to `errorMessage` to be consistent with other parts of our codebase.
This would then need some mapping in `synthesizeAttributes()` to change change it back to `error_message` for the resulting Terraform JSON.

packages/cdktf/lib/terraform-variable.ts
```diff
@@ -86,6 +91,11 @@ export interface TerraformVariableConfig {
    * The nullable argument in a variable block controls whether the module caller may assign the value null to the variable.
    */
   readonly nullable?: boolean;
+
+  /**
+   * Specify arbitrary custom validation rules for a particular variable using a validation block nested within the corresponding variable block
+   */
+  readonly validation?: TerraformVariableValidationConfig;
```
ansgarm, line 98: This should probably take an (optional) array of `TerraformVariableValidationConfig`, as it is valid to pass multiple validation blocks (see bottom of this [docs section](https://www.terraform.io/language/values/variables#custom-validation-rules)).

packages/cdktf/lib/terraform-variable.ts
```diff
@@ -51,6 +51,11 @@ export abstract class VariableType {
   }
 }
 
+export interface TerraformVariableValidationConfig {
+  readonly condition: any;
```
ansgarm, line 55: I'm unsure whether we could further narrow down this type?
Maybe allow `string | IResolvable`? This will effectively have the same behaviour as `any` in languages like e.g. Java (treating it as `Object`) but will carry more type information in TypeScript and Python.
If we expect a `can` function to be used at the root of the expression, we could even go with only `IResolvable` and drop `string`.

ansgarm (commented):
Thank you so much for leading the way on this @bollohz 🚀

I left a couple of comments and I think it might make sense to add some tests for this. Probably snapshot tests, which confirm the correct `.tf.json` output when using `Fn.can` etc. in the `condition` expression passed for the validation.
Happy to help with this of course!

packages/cdktf/lib/terraform-variable.ts
jsteinich: `Expression` is pretty close to the correct type (technically any), though there are limits to what can be contained in the expression.

jsteinich (commented):
Thanks for this. Pretty much echo what Ansgar said.

We should probably also update https://github.com/hashicorp/terraform-cdk/blob/b4bab56f671c395ea932bcb8222a485229444521/packages/%40cdktf/hcl2cdk/lib/schema.ts#L15, so that the convert command handles it as well (we missed adding nullable to that).

bollohz:
I've changed the code as you guys as suggested. 
Where I can find things related to snapshot tests in order to make a test also for the validation block?

packages/cdktf/lib/terraform-variable.ts
bollohz: Done!

packages/cdktf/lib/terraform-variable.ts
bollohz: My bad, done and added logic!

packages/cdktf/lib/terraform-variable.ts
bollohz: I have added the `Expression` as @jsteinich said.

ansgarm:
Thank you @bollohz!

> Where I can find things related to snapshot tests in order to make a test also for the validation block?

I think here would be a good place (similar to the one for `nullable`):
https://github.com/hashicorp/terraform-cdk/blob/0bfc9d5c1c4e1a94b529c74317e6aa04038c63ab/packages/cdktf/test/variable.test.ts#L161-L170

packages/cdktf/lib/terraform-variable.ts
```diff
@@ -142,6 +154,10 @@ export class TerraformVariable
       type: this.type,
       sensitive: this.sensitive,
       nullable: this.nullable,
+      validation: this.validation.map((validation) => ({
```
jsteinich, line 157: Need to check if `this.validation` is undefined.

jsteinich (commented):
One small thing and please make sure to add a test case. Let us know if you run into any trouble with that.

packages/cdktf/lib/terraform-variable.ts
```diff
@@ -154,7 +154,7 @@ export class TerraformVariable
       type: this.type,
       sensitive: this.sensitive,
       nullable: this.nullable,
-      validation: this.validation.map((validation) => ({
+      validation: this?.validation.map((validation) => ({
```
ansgarm, line 157: This would check whether `this` is undefined. See below for checking `this.validation`:
```suggestion
      validation: this.validation?.map((validation) => ({
```

packages/cdktf/lib/testing/adapters/__tests__/jest.test.ts
```diff
@@ -30,15 +30,9 @@ describe("jest-adapter", () => {
...
-        Found 1 test_resource resources instead:
-        [
-          {
-            \\"name\\": \\"bar\\"
-          }
-        ]"
-      `);
+      ).toThrowErrorMatchingInlineSnapshot(
+        `"Class constructor Construct cannot be invoked without 'new'"`
+      );
```
ansgarm, line 35: This does not seem right 🤔

packages/cdktf/test/variable.test.ts
```diff
@@ -168,3 +169,19 @@ test("nullable variable", () => {
...
+test("validation block variable", () => {
+  const app = Testing.app();
+  const stack = new TerraformStack(app, "test");
+
+  new TerraformVariable(stack, "test-variable", {
+    type: "string",
+    validation: [
+      {
+        errorMessage: "Validation failed..",
+        condition: Fn.regex("mickey_mouse", "mickey_mouse"),
```
jsteinich, line 182: This doesn't seem like a very realistic test case. I'd suggest using one or both of the examples in the documentation: https://www.terraform.io/language/values/variables#custom-validation-rules

jsteinich:
@bollohz would you be willing to update this and address the last couple feedback points?

bollohz:
> 

Hi, sorry for not being able to continue on that. I try to address the PR in this day!

packages/cdktf/test/variable.test.ts
jsteinich: The new test is much better, thank you.
I do still think we should actually reference the variable in the condition since that seems almost necessary for any practical usage of this. Unfortunately, I don't think a self reference during construction will work, so we may need an alternative way to specify it.

bollohz:
> @bollohz would you be willing to update this and address the last couple feedback points?

Hello there, with the help of @DanielMSchmidt I think we address some points. Please check it :)

jsteinich:
> Hello there, with the help of @DanielMSchmidt I think we address some points. Please check it :)

@bollohz thanks for updating. I think there is still one last point to be addressed:

https://github.com/hashicorp/terraform-cdk/pull/1533#discussion_r824238798

jsteinich:
> @jsteinich is it good to merge?

There's still one outstanding piece of feedback which I believe is critical to verifying that this is actually usable.

packages/cdktf/test/variable.test.ts
bollohz: @jsteinich Following the TF documentation, seems that it evaluates directly the variabile during the assertion phase.
During construction I think is not possible, as you said, to use the `this` keyword, but I don't have idea on how we could adress this. 
Maybe for the test is valuable something like `Fn.can(Fn.regex("^ami-", "${var.variable_name}"))` making possible the check during the `apply/deploy` command using directly the TF directives. 

(Sorry for ghosting here, I'm really sorry but I need to address a lot of stuff.. 😓 )

jsteinich: No worries. I'll poke around a bit to try and come up with a pattern that makes sense.

jsteinich: @bollohz sorry for taking so long to get back to this. I ended up just making it possible to add a validation after creation. That makes it possible to reference the created variable via `var.fqn`.

I thought about using something like `this` or `self` or `@var`, but it seemed too arbitrary and no clear sense of what it should be (especially when thinking about different languages).

DanielMSchmidt (approved):

PR #1537, opened 2022-02-03
chore: document submodules

DanielMSchmidt:
I didn't knew we support this, when I found out I thought a quick addition to the docs might be nice

### DIFF
--- a/website/docs/cdktf/concepts/modules.mdx
+++ b/website/docs/cdktf/concepts/modules.mdx
@@ -87,9 +87,7 @@ For local modules, use the object format.
 }
 ```
 
-#### Generating bindings for submodules
-
-For performance reasons we don't generate bindings for submodules. To generate bindings for submodules, need to specify the module source as `terraform-aws-modules/eks/aws//modules/eks-managed-node-group`, where after the `//` is the path to the submodule in the modules repository.
+For performance reasons we don't automatically generate bindings for submodules. To generate bindings for submodules, specify the module source as `terraform-aws-modules/eks/aws//modules/eks-managed-node-group`, where after the `//` is the path to the submodule in the modules repository.
 
 ### Generate Module Bindings
 
--- a/website/docs/cdktf/concepts/modules.mdx
+++ b/website/docs/cdktf/concepts/modules.mdx
@@ -87,7 +87,7 @@ For local modules, use the object format.
 }
 ```
 
-For performance reasons we don't automatically generate bindings for submodules. To generate bindings for submodules, specify the module source as `terraform-aws-modules/vpc/aws//submodules/vpc-endpoints`, where after the `//` is the path to the submodule in the modules repository. Please visit [the terraform source specification](https://www.terraform.io/language/modules/sources) for more details.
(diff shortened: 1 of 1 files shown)


website/docs/cdktf/concepts/modules.mdx
```diff
@@ -87,6 +87,10 @@ For local modules, use the object format.
 }
 ```
 
+#### Generating bindings for submodules
+
+For performance reasons we don't generate bindings for submodules. To generate bindings for submodules, need to specify the module source as `terraform-aws-modules/eks/aws//modules/eks-managed-node-group`, where after the `//` is the path to the submodule in the modules repository.
```
laurapacilio, line 92: Could we try putting this in the same section with "Generating Bindings for modules"?  I'm not sure it needs its own subsection. 

```suggestion
For performance reasons we don't automatically generate bindings for submodules. To generate bindings for submodules, specify the module source as `terraform-aws-modules/eks/aws//modules/eks-managed-node-group`, where after the `//` is the path to the submodule in the modules repository.
```

laurapacilio (changes requested):
A couple thoughts for you :) Thank you!

website/docs/cdktf/concepts/modules.mdx
```diff
@@ -87,6 +87,8 @@ For local modules, use the object format.
 }
 ```
 
+For performance reasons we don't automatically generate bindings for submodules. To generate bindings for submodules, specify the module source as `terraform-aws-modules/eks/aws//modules/eks-managed-node-group`, where after the `//` is the path to the submodule in the modules repository.
```
jsteinich, line 90: Given the wording, it feels like a jump to go from the vpc module example above this to some completely different module. Might be better to use a related submodule (https://registry.terraform.io/modules/terraform-aws-modules/vpc/aws/latest/submodules/vpc-endpoints) or tweak the wording slightly to make it clear that this a different example.

Might also be nice to indicate that any module source format specified in https://www.terraform.io/language/modules/sources should work.

website/docs/cdktf/concepts/modules.mdx
```diff
@@ -87,6 +87,8 @@ For local modules, use the object format.
 }
 ```
 
+For performance reasons we don't automatically generate bindings for submodules. To generate bindings for submodules, specify the module source as `terraform-aws-modules/vpc/aws//submodules/vpc-endpoints`, where after the `//` is the path to the submodule in the modules repository. Please visit [the terraform source specification](https://www.terraform.io/language/modules/sources) for more details.
```
laurapacilio, line 90: I don't think we need to do a full link here because this is internal to our site! I think we can just do the path after the terraform.io part :-) 

```suggestion
For performance reasons, we don't automatically generate bindings for submodules. To generate bindings for submodules, specify the module source as `terraform-aws-modules/vpc/aws//submodules/vpc-endpoints`, where after the `//` is the path to the submodule in the modules repository. Refer to [the Terraform source specification](/language/modules/sources) for more details.
```

laurapacilio (approved):
Thank you!

PR #1539, opened 2022-02-04
test(hcl2cdk): check if produced code can generate HCL

DanielMSchmidt:
I was wondering in what state convert is and it's not super great, but not awful. I have an enum with all the ways it fails (and never if it's not intended to work, e.g. due to local modules) and these are the cases:

- `MissingCdktfImport`: `import * as cdktf` is added only when needed, seems like we missed a spot here, should be a quick fix
- `SingleItemBlocksAreTreatedAsArrays`: We changed our provider API after introducing convert and it seems like we didn't adjust convert properly to deal with the change
- `BooleanAsIResolvable`:  Type 'IResolvable' is not assignable to type 'boolean'." errors, I think because I'm running against 0.9 and not a built dist folder not sure though
- `MaximumCallStackSizeExceeded`: Seems like the resulting CDK code produces an infinite loop, I guess we should look into it
- `UnforseenRename`: e.g. User to UserA, convert needs to know about renames and we need to check if it's needed in this case
- `StringIsNotAssignableToListOfString`: e.g. Type 'string' is not assignable to type 'string[] | undefined', not sure why this happened

I am not going to fix any more tests in this PR, the main goal here is to keep the state we have working and give us all the opportunity to fix individual bugs if we feel like it :)


packages/@cdktf/hcl2cdk/test/hcl2cdk.test.ts
```diff
@@ -1,39 +1,89 @@
...
 ];
 
+enum Synth {
+  yes,
+  needsAFix_MissingCdktfImport,
+  needsAFix_SingleItemBlocksAreTreatedAsArrays,
+  needsAFix_BooleanAsIResolvable, // Type 'IResolvable' is not assignable to type 'boolean'.",
+  needsAFix_MaximumCallStackSizeExceeded,
+  needsAFix_UnforseenRename, // e.g. User to UserA
+  needsAFix_StringIsNotAssignableToListOfString, // e.g. Type 'string' is not assignable to type 'string[] | undefined'.
```
ansgarm, line 29: Nice!

ansgarm (approved):
Is/Are there a follow-up issue(s) for the cases mentioned that can be improved in further iterations? Would be perfect to even link it in the enum / them in the enum values.

Issue #1134, opened 2021-10-11
Support `terraform import` via the cdk cli


fmmoret:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Support importing existing resources into a stack's state file.

### References

https://www.terraform.io/docs/cli/import/index.html

remyleone:
That would be very useful to have a way to generate HCL configuration from existing configuration such as https://github.com/cycloidio/terracognita seems to provide for different vendors.

DanielMSchmidt:
With the `cdktf convert` command you can take the output of terracognita and use it to create an equivalent CDK program

jsteinich:
@rirze Thanks for sharing your workflow.

I could see having some like `importFrom` available on all resources. This would basically just do what you are already doing with an override and a basic script of Terraform commands. That feels like a relatively low scope way of providing import support.
Could perhaps add support for some custom setting/generation for resources with strange semantics. Likely wouldn't want to go to far down this route though.

Might be interesting to have some sort of `createResource` method available on data sources which effectively does an import, but with the necessary id queried through friendlier properties. This would be harder to automatically generate and doesn't directly fit within current Terraform operations, but could be pretty slick.

rirze:
@jsteinich Glad to see some feedback on this-- I think providing a basic level of "use this resource_id when using `terraform import`" would go a long way for this feature. I think due to the high level of you guys are trying to support, I think it'll be a while before a more native approach would manifest. Even for the peculiar case of `aws_security_group_rule`, I could support this workflow by making API calls during the synthesize process (which I'm not too keen on, but just saying it is possible). 

If you want to move forward with this idea, do you see a particular way of exposing this to the user? I've seen @ansgarm detail an approach using an `Imports.of(resource...).addImportId(resource_id)` pattern (https://github.com/hashicorp/terraform-cdk/pull/1543#issuecomment-1033611825). That seems like the easiest way to implement this, although it might be a little verbose for someone not familiar with that style.

jsteinich:
[...]

PR #1543, opened 2022-02-06
feat(lib): update attributes['//'] instead of assigning to it

rirze:
Should allow end users to also write to the `"//"` attribute in Terraform resources. Currently, `cdktf` overwrites this attribute with the metadata values.


DanielMSchmidt:
That's an interesting use-case, what do you plan to use the // field for?

rirze:
Personally, I'm using to test out an Pulumi-style `import` behavior. Essentially, I would attach an `importId` parameter into the `//` attribute, then run a script that generates `terraform import ...` statements based on the type of resource and the `importId` given. 

Without this, I don't any side-effect free way to 'tag' Terraform resources so I can identify which resources to import just by looking at the generated `cdktf.json` alone. 

Hopefully that gives you some context-- I have no intention to alter the existing metadata that `cdktf` attaches. If you're collecting that data, maybe I can restrict the library to only set the `//.metadata` attribute?

DanielMSchmidt (approved):

DanielMSchmidt:
Could you run `yarn format` to make CI pass?

rirze:
Sure, I'll make sure to run `yarn format`. 

I was thinking, as a way to expand this idea; should I implement a `addComment` function to `TerraformResource` (and probably `TerraformStack` too...)? Currently I'm using overrides to add this "comment" to a resource, but it results in this top level list of "overrides" in the `cdktf.json`. I'm trying to figure out if a more native solution would be ok. 

Let me know your thoughts.

jsteinich:
> I was thinking, as a way to expand this idea; should I implement a `addComment` function to `TerraformResource` (and probably `TerraformStack` too...)

I don't think `addComment` makes sense as the json output isn't meant to be read by a human. Perhaps `addMetadata` would be a better name.



> Personally, I'm using this to test out a Pulumi-style `import` behavior. Essentially, I would attach an `importId` parameter into the `//` attribute, then run a script that generates `terraform import ...` statements based on the type of resource and the `importId` given.

This sounds like something could be built into cdktf. Probably not in the context of this PR, but possibly as part of https://github.com/hashicorp/terraform-cdk/issues/1134.

ansgarm:
Seconding in on `addComment`:

We already have support for `Annotations` (used for example in the last code block in the [Aspects docs](https://www.terraform.io/cdktf/concepts/aspects#define-aspects)). Those `Annotations` serve the purpose of adding a comment to a construct. They are printed when running `synth` and (in the case of `error`) can even fail the synth process.

While `addMetadata` would certainly be a better name, I am a bit hesitant to just add that functionality without having done a bit more research on what kinds of workflows this is supposed to enable.

As Jon said, the import behavior would certainly be something that should be included as a first class citizen in the CDKTF with a clear and straight forward API.

---

> Currently I'm using overrides to add this "comment" to a resource, but it results in this top level list of "overrides" in the cdktf.json.

I think we could exclude all overrides nested in `//` to get rid of this to help with that.

To make the way of interacting with the overrides for `//` more beautiful to work with, you could build yourself a helper with an API similar to `Annotations.of()` or `Aspects.of()`. Something like this (untested):

```ts
class Imports {
  protected constructor(private target: TerraformResource) {}

  public static of(target: TerraformResource) {
    return new Imports(target);
  }

  public addSomeConfig(config: any): this {
    this.target.addOverride("//.__imports_config", config);
    return this;
  }
}

Imports.of(myS3Bucket).addSomeConfig({ todoo: 'put something here' })
```

rirze:
All those who were interested in my approach on how I was automating importing resources, I wrote a detailed comment @ https://github.com/hashicorp/terraform-cdk/issues/1134#issuecomment-1035246790. 

I can provide more details as needed and I'd be comfortable moving the discussion regarding that to another issue/PR if necessary.

PR #1567, opened 2022-02-15
fix(tests): Windows needs special docker host override to work

x3cion:
Hey there, 👋

this is a little addition to the test helper, adding `npipe:////.//pipe//docker_engine` as host for the docker provider, if the tests are run on windows.

It is a workaround mentioned here: https://github.com/hashicorp/terraform-provider-docker/issues/180#issuecomment-579144846

### DIFF
--- a/packages/cdktf/test/helper/provider.ts
+++ b/packages/cdktf/test/helper/provider.ts
@@ -57,5 +57,10 @@ export class DockerProvider extends TerraformProvider {
       },
       terraformProviderSource: "kreuzwerker/docker",
     });
+
+    // Windows needs the special docker host configuration to work
+    if (process.platform === "win32") {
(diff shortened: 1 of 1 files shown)


ansgarm:
Thank you for the fix @x3cion! Much appreciated!

This brings the question: Should we run start to run unit tests on Windows in CI? /cc @DanielMSchmidt @skorfmann
While integration tests (which we already run on Windows) confirm that everything works from an end user perspective, running unit tests on Windows could give us confidence that contributors using Windows are able to run tests locally without encountering errors like the one @x3cion fixed with this PR. What do you think?

DanielMSchmidt:
@ansgarm I think supporting windows unit tests is a good idea. We could use a similar mechanism as in the integration tests to make snapshots working on both platforms when there are differences

DanielMSchmidt (approved):

PR #1595, opened 2022-02-25
fix: Support Auto Updating Snapshots for forks

jsteinich:
The goal is to support the `ci/update-snapshots` for PRs originating in forks.
I'm not certain if this will work as is and I'm not sure the best way to test this without merging in.


ansgarm:
Me neither 😁 I'd be happy with just trying – we can always revert 👍

ansgarm (approved):

ansgarm:
@jsteinich you could try this on a new PR of yours now :)

jsteinich:
It took almost 2.5 hours, but it seems like it worked.

ansgarm:
Yeah, they run in sequence because else they could overwrite each other (with git not properly detecting changes)

PR #1617, opened 2022-03-09
fix(cli): parse JSON log lines if possible

DanielMSchmidt:
Previously we would print an inlined JSON when running against TFC with structured log output enabled


ansgarm (approved):
Nice

PR #1621, opened 2022-03-14
feat(cli): handle dismissing a single stack vs stopping all stacks

DanielMSchmidt:
This is important to ensure we handle cancelation workflow nicely when a lot of concurrent work might be going on.

dismiss-stop


packages/cdktf-cli/bin/cmds/ui/components/bottom-bars/approve.tsx
```diff
@@ -1,9 +1,64 @@
...
+    label: "Approve",
+    description: `Applies the changes outlined in the plan.`,
+  },
+  dismiss: {
+    label: "Dismiss",
+    description: `Does not apply the changes outlined in the plan. This will also prevent depending stacks from planning.`,
+  },
+  stop: {
+    label: "Stop",
+    description: `Does not apply the changes. Currently running stacks will be finished, but no new ones will be started.`,
```
skorfmann, line 24: What is the exact difference between `stop` and `dismiss`? Is it that `dismiss` only stops the current branch of execution where `stop` stops the entire application deployment?

skorfmann (commented):
Looks pretty slick :)

packages/cdktf-cli/bin/cmds/ui/components/bottom-bars/approve.tsx
DanielMSchmidt: Yeah exactly. Stop is a nicer /  needed because a ctrl-c would kill running applies (if we have parallelism) so you'd have to individually dismiss every single plan.

DanielMSchmidt: No, stopping is pretty immediate. Plus you see all the output so you have a way to tell. Showing the status of all the stacks in progress is part of the parallelism work, so it's next up :)

PR #1623, opened 2022-03-14
chore: use distinct GitHub token to push changes

DanielMSchmidt:
Using the actions default token does not trigger a CI run, see https://github.community/t/push-from-action-does-not-trigger-subsequent-action/16854

I created a new token with the team-tf-cdk user and used it here.

### DIFF
--- a/packages/cdktf/test/__snapshots__/variable.test.ts.snap
+++ b/packages/cdktf/test/__snapshots__/variable.test.ts.snap
@@ -106,7 +106,7 @@ exports[`reference 1`] = `
   \\"resource\\": {
     \\"test_resource\\": {
       \\"test-resource\\": {
-        \\"name\\": \\"\${var.test-variable}\\"
+        \\"name\\": \\"\${var.test-variablasdasdadse}\\"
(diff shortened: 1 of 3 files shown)


packages/cdktf/test/__snapshots__/variable.test.ts.snap
```diff
@@ -106,7 +106,7 @@ exports[`reference 1`] = `
   \\"resource\\": {
     \\"test_resource\\": {
       \\"test-resource\\": {
-        \\"name\\": \\"\${var.test-variable}\\"
+        \\"name\\": \\"\${var.test-variablasdasdadse}\\"
```
ansgarm, line 109: On purpose?

ansgarm: Oh 🙈

ansgarm (approved):

jsteinich:
You won't really be able to test this until it is merged in. That's a side effect of making it so that it actually functions for PRs from forks.

PR #1625, opened 2022-03-15
feat(cli): parallelize cdktf deploy/destroy

DanielMSchmidt:
This PR allows you to Deploy / Destroy multiple stacks as parallel as possible / permitted.

## TODOs

- [ ] implement parallel deploy
- [ ] implement parallel destroy
- [ ] implement multi-stack status bar
- [ ] stop log output while a plan is surfaced
- [ ] color code stacks
- [ ] highlight the name of the stack to approve
- [ ] write test case that ensures we don't go over parallelism limit


packages/cdktf-cli/lib/cdktf-project.ts
```diff
@@ -253,31 +324,89 @@ export class CdktfProject {
...
+          // We have to wait for approval again, therefore we can not resume
+          return;
+        }
+      }
+      if (event.type === "logMessage") {
+        event.cb(event.value);
+      }
+
+      event = this.eventBuffer.shift();
+    }
```
ansgarm, line 347: nit: This could've been a do-while loop, which would get rid of the duplicated `event = this.eventBuffer.shift();` at the bottom.
Despite that, you could also move the `if (event.value.type === "waiting for approval")` into the condition of the do-while loop to stop iterating when you encounter an `waiting for approval` event.
To return without setting waitingForApproval to false, you could check the last event for its type.
Sorry for being so nitty here 😅 I'm unsure whether my proposal is actually easier to understand or not, so I'm happy with you just dismissing this discussion 😁
For comparison, this is what I had in mind:
```ts
// We first need to flush all events, but we can not resume if there is a new waiting for approval update
let event;
do {
  event = this.eventBuffer.shift();
  event.cb(event.value);
} while (event && event.value.type !== "waiting for approval");

if (event.value.type !== "waiting for approval") {
  this.waitingForApproval = false;  
} // else: We have to wait for approval again, therefore we can not resume
```

packages/cdktf-cli/lib/cdktf-project.ts
```diff
@@ -253,31 +324,89 @@ export class CdktfProject {
...
-        cb({
+        const callbacks = (update: StackApprovalUpdate) => ({
+          approve: () => {
+            update.approve();
+            this.resumeAfterApproval();
+          },
+          dismiss: () => {
+            update.reject();
+
+            this.stopAllStacksThatCanNotRunWithout(update.stackName);
```
ansgarm, line 373: I love the name of that function 😄

packages/cdktf-cli/lib/cdktf-project.ts
```diff
@@ -325,7 +464,43 @@ export class CdktfProject {
...
+
+    // We end the loop when all stacks are started, not we need to wait for them to be done
+    await Promise.all(
+      this.stacksToRun.reduce((carry, stack) => {
+        const promise = stack.currentWorkPromise;
+        if (promise) {
+          carry.push(promise);
+        }
+        return carry;
+      }, [] as Promise<void>[])
```
ansgarm, line 498: another .filter().map()? 😁

packages/cdktf-cli/lib/cdktf-project.ts
```diff
@@ -325,7 +464,43 @@ export class CdktfProject {
...
+          break;
+        }
+
+        method === "deploy"
+          ? nextRunningExecutor.deploy()
+          : nextRunningExecutor.destroy();
+      }
+    }
+
+    // We end the loop when all stacks are started, not we need to wait for them to be done
```
ansgarm, line 490: what? I don't understand the second part of the sentence 😅

packages/cdktf-cli/lib/cdktf-project.ts
```diff
@@ -341,16 +516,15 @@ export class CdktfProject {
...
       this.getStackExecutor(stack, opts)
     );
+
     const next = opts.ignoreMissingStackDependencies
-      ? () => this.stacksToRun.filter((stack) => stack.isPending)[0]
+      ? () =>
+          Promise.resolve(
+            this.stacksToRun.filter((stack) => stack.isPending)[0]
+          )
       : () => getStackWithNoUnmetDependencies(this.stacksToRun);
```
ansgarm, line 525: I like 👍

packages/cdktf-cli/test/lib/project-execution.test.ts
```diff
@@ -538,4 +559,53 @@ describe("CdktfProject", () => {
...
+          events.push(event);
+        },
+      });
+
+      await cdktfProject.deploy({
+        stackNames: new Array(5).fill(0).map((_, i) => `stack-${i}`),
+        parallelism: 100,
+      });
+
+      expect(eventsDuringWaitForApprove.length).toBe(0);
```
ansgarm, line 608: nice testcase

packages/cdktf-cli/bin/cmds/ui/components/bottom-bars/status.tsx
```diff
@@ -1,118 +1,97 @@
...
 
-  if (done) {
-    return {
-      text: "Done",
-      color: "green",
-      showSpinner: false,
-    };
-  }
+    case "waiting for approval of stack":
+      return <Text>Not implemented yet</Text>;
```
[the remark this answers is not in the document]
DanielMSchmidt: I think it;s not needed, this is a differnet bar that's rendered by the top level 🤔  I'll remove the text for now

packages/cdktf-cli/bin/cmds/ui/list.tsx
```diff
@@ -32,7 +32,7 @@ export const List = ({
                     <Text bold>Path</Text>
                   </Box>
                 </Box>
-                {projectUpdate.stacks.map((stack) => (
+                {returnValue!.map((stack) => (
```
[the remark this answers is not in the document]
DanielMSchmidt: That is being done, the fundamental problem is that I don't know how to tell xstates Typescript types that this value will be there after step XYZ, therefore the types running through have an undefined in them

packages/cdktf-cli/lib/cdktf-project.ts
```diff
@@ -82,18 +83,45 @@ function getMultipleStacks(
...
+    }
+  }, [] as Promise<void>[]);
+
+  if (!stackExecutionPromises.length) {
+    return undefined;
+  }
+
+  await Promise.race(stackExecutionPromises);
+
+  return await getStackWithNoUnmetDependencies(stackExecutors);
```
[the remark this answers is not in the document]
DanielMSchmidt: > Another short question: What happens if a stack is waiting for user approval? Will its currentWorkPromise be resolved at that point (which might cause a loop of this function in that case; because it is partially done but not currentState === "done")?

The currentWorkPromise is resolved when the state machine is in the done state, it will not resolve if you e.g. approve. if you dismiss it will resolve, but at that point all the executors what depended on this stack will have been removed from the stacksToRun list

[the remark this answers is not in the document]
ansgarm: Ah, alright 👍 Sounds safe enough to me.
I'd still add a tracelog message because that might really save us if we break the statemachine or something like that and it wouldn't hurt if not :)

packages/cdktf-cli/lib/cdktf-project.ts
```diff
@@ -220,11 +283,19 @@ export class CdktfProject {
   private onLog?: (log: { stackName: string; message: string }) => void;
   private abortSignal: AbortSignal;
 
-  // Set during deploy / destroy, may be mutated by approve / stop actions
-  private stacksToRun: CdktfStack[] = [];
+  // Set during deploy / destroy
+  public stacksToRun: CdktfStack[] = [];
```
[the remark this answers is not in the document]
DanielMSchmidt: No, should not be set by outside, but adding yet another property is not a good idea, having _stacksToRun and stacksToRun can be confusing

packages/cdktf-cli/lib/cdktf-project.ts
DanielMSchmidt: While I don't like what we have I have to say I'd prefer a double  `event = ...` part over checking for the reason we left the loop after the loop was left. I guess this piece of code is what we get for not using RxJS, but on the other hand we don't have to deal with RxJS, so I'd consider it a fair trade :D

packages/cdktf-cli/lib/cdktf-project.ts
DanielMSchmidt: typo, not should be now

packages/cdktf-cli/lib/cdktf-project.ts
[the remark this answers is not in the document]
ansgarm: yeah, agree'd

packages/cdktf-cli/lib/cdktf-project.ts
ansgarm: fine with me, no strong feelings here :)

ansgarm (approved):
Nice work 💯

PR #1642, opened 2022-03-23
fix(tests): Ensure watch tests are idempotent

skorfmann:
Doing the setup / teardown work on a global before/afterAll level lead to loosing references to the spawned processes and therefore to zombie processes which prevented jest from exiting cleanly. Changing this to before/afterEach ensures that the references are intact, even when the tests are automatically retried via jest.

Beyond that, this PR drops the global retries for the integration tests and slightly improves the logging for the watch tests.


DanielMSchmidt:
This seems stale and we don't seem to have problems with these tests anymore, closing it for now

PR #1652, opened 2022-03-24
chore(docs): local preview, post split

thiskevinwang:
# Description

This adds `make website/build-local` & `make website/local` to support local dev/preview, _after_ the `terraform-cdk` submodule has been removed from https://github.com/hashicorp/terraform-website

## Todo
- [ ] Update the branch to `main` in the remote Dockerfile

---

This is dependent on 

- https://github.com/hashicorp/terraform-website/pull/2209
  - this branch removes the `terraform-cdk` submodule
  - this depends on 
    - https://github.com/hashicorp/react-components/pull/524


scripts/website-preview.sh
```diff
@@ -0,0 +1,97 @@
+#!/bin/bash
```
BRKalow, line 1: So this kind of blends the two approaches we have been exposing in our other preview workflows, which is: (1) running the app directly via node and (2) running the app via docker.

With the docker workflow, we should not need to do the clone step. We should be relying on the published docker image, or in the case of the "local" workflow we can build the docker image locally from the remote repository without the clone step ([ref](https://stackoverflow.com/questions/26753030/how-to-build-docker-image-from-github-repository)). for consistency, this would all be instrumented in a make command (one for using the published image, one for building and using a local image).

For the node based workflow, we would do the clone step like you've got here then run the tf website with the correct env variables so it reads the local CDK content.

thiskevinwang: https://github.com/hashicorp/terraform-cdk/pull/1652/commits/b50fa8658bc876d03deed96954fdb2579562f133 This eliminates the _clone_ step, and builds the Docker image from git

thiskevinwang: https://github.com/hashicorp/terraform-cdk/pull/1652/commits/9a2f3aacb566287ceb8c5f58e7d36cacfc4159dc Thie replaces `npm run website-preview` with `make website-preview`

Makefile
```diff
@@ -14,4 +14,7 @@ website:
 		npm start
 
 .DEFAULT_GOAL := website
-.PHONY: website
\ No newline at end of file
+.PHONY: website
+
+website-preview:
+	scripts/website-preview.sh
```
BRKalow, line 20: I kinda like the script convention here, as opposed to inlining everything in the makefile. Goal: can we support both `make website` (published image, no docker build) and `make website/local` (builds local image from terraform-website repo) with one script and adjusting inputs with env vars?

BRKalow: An alternative is trying to leverage env vars further in the makefile so we end up with something like this:

https://github.com/hashicorp/waypoint/blob/main/website/Makefile

BRKalow: Something like the above makes it very easy roll it out to other projects as well, as all we need to do is change one env var

thiskevinwang: > Goal: can we support both make website (published image, no docker build) and make website/local

Yea I can add support for both!

💭 To make the script _portable_, I could expose the various hardcoded variables as flags.
- The script could be generic, and copy-pasted around.
- The `Makefile`s would pass in repo-specific env vars as flags
  - This could help keep the `Makefile`s a bit leaner

thiskevinwang: 🤔 Not yet sure if that'll get unwieldy or not

scripts/website-preview.sh
```diff
@@ -0,0 +1,50 @@
...
+echo ""
+
+docker run \
+    -e "IS_CONTENT_PREVIEW=true" \
+    -e "NAV_DATA_PATH=$NAV_DATA_PATH" \
+    -e "CONTENT_DIR=$CONTENT_DIR" \
+    --interactive \
+    --rm \
+    --tty \
+    --workdir "$WORKFIR" \
```
BRKalow, line 43: fir

Makefile
thiskevinwang: Ok, so going to go for a 100% Makefile alternative

BRKalow (approved):
@thiskevinwang I switched the base branch to be my PR, but this looks good so I think we can merge it there and prep for the full implementation! Nice work

PR #1664, opened 2022-03-28
fix(readme): update links to getting started guides

ansgarm:
Our getting started guides where migrated to a single Learn tutorial (with tabs to select the target language).
We can't currently preselect the language, so we link to the same tutorial for each language for now.

Should we adjust the wording of the next sentence as it also links to Learn?

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,11 +26,11 @@ CDKTF includes two packages:
 
 Choose a language:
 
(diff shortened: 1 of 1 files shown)


laurapacilio (approved):
thank you for fixing!!

schersh (approved):
thanks!

PR #1670, opened 2022-03-29
refactor(cli): remove xstate as state machine

DanielMSchmidt:
Xstate lead to too much indirection due to our need to communicate progress and the weak typings we need to have since context stores everything.


packages/cdktf-cli/lib/cdktf-stack.ts
```diff
@@ -266,83 +132,201 @@ export class CdktfStack {
...
-    return new Promise((resolve, reject) => {
-      if (this.currentState === "done") {
-        resolve();
-      }
-
-      if (this.currentState === "error") {
-        reject(new Error(this.stateMachine.state.context.message));
+  private logCallback(
+    stateName: string
+  ): (message: string, isError?: boolean) => void {
```
ansgarm, line 171: In `TerraformCloud` this is defined as:
```
private readonly createLogSender = (_phase: string) => (_stdout: string, _isErr = false) => {}
```
In `TerraformCli` this is defined as:
```
onLog = (_stateName: string) => (_stdout: string, _isErr = false) => {}
```
I personally think that `phase` is the right term to not confuse this with the state / current state of this class.
I'd opt to name all methods / arguments `createLogSender` due to this being a higher order function.
For the other args I vote for `message` and `isError`.

ansgarm: After reading more of the code, I noticed that this does not handle just any logs but logs emitted by Terraform (either the CLI or the CLI running in TFC). So, I'd propose to make this clear.
Something along the lines of: `createTerraformLogHandler` even, because that is what it does in the end (afaik). Or am I missing something?

packages/cdktf-cli/lib/cdktf-stack.ts
```diff
@@ -266,83 +132,201 @@ export class CdktfStack {
   public get isRunning(): boolean {
     return !this.isPending && !this.isDone;
   }
-  public get currentWorkPromise(): Promise<void> | undefined {
-    if (!this.isRunning) {
-      return undefined;
-    }
 
-    return this.waitOnMachineDone();
+  private notifyState(
```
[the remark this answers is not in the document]
ansgarm: Just saw that there's already a `handleState()` method 😁 But I'm going to propose a different name for that one 😜

packages/cdktf-cli/lib/cdktf-stack.ts
```diff
@@ -266,83 +132,201 @@ export class CdktfStack {
...
+      this.notifyState({
+        type: "waiting for stack approval",
+        stackName: this.stack.name,
+        plan: plan,
+        approve: () => {
+          resolve(true);
+        },
+        reject: () => {
+          resolve(false);
+        },
```
ansgarm, line 193: Love this ❤️

packages/cdktf-cli/lib/cdktf-stack.ts
```diff
@@ -266,83 +132,201 @@ export class CdktfStack {
...
+    await this.handleState(async () => {
+      this.notifyState({ type: "planning", stackName: this.stack.name });
+      const terraform = await this.initalizeTerraform({ isSpeculative: false });
+
+      const plan = await terraform.plan(true);
+      this.notifyState({ type: "planned", stackName: this.stack.name, plan });
+
+      const approved = this.context.autoApprove
+        ? true
+        : await this.waitForApproval(plan);
```
ansgarm, line 293: this is so seamless, wow

packages/cdktf-cli/lib/cdktf-stack.ts
```diff
@@ -266,83 +132,201 @@ export class CdktfStack {
...
+      if (!approved) {
+        this.notifyState({ type: "dismissed", stackName: this.stack.name });
+        return;
+      }
 
-    return this.waitOnMachineDone();
+      this.notifyState({ type: "destroying", stackName: this.stack.name });
```
ansgarm, line 299: We could `Omit<StackUpdate, 'stackName'> in `notifyState` to get rid of all the `stackName: this.stack.name` in numerous places and add the `stackName` in `notifyState`. What do you think? Would this make the code simpler in your eyes? Or harder to follow / grasp?

packages/cdktf-cli/lib/cdktf-stack.ts
```diff
@@ -266,83 +132,201 @@ export class CdktfStack {
...
+    return terraform;
   }
 
-  public async deploy(stackName?: string) {
-    this.stateMachine.send({
-      type: "START",
-      targetAction: "deploy",
-      targetStack: stackName,
-    });
+  private async handleState(cb: () => Promise<void>) {
```
ansgarm, line 215: I'd propose something like `handleExecution()` or even a short `run()` because it also handles the `currentWorkPromise` and checks `this.stopped`. While it does handle state, it only handles the "run" state part of all the types that `currentState` may hold. `handleRunState` would also differentiate this enough for me.

ansgarm (commented):
I like this a **lot** more than `xstate` 🎉
My comments are only about making naming more consistent and clear (from my point of view at least 😄). So sorry in advance if this feels like bikeshedding 😬😂

packages/cdktf-cli/lib/cdktf-stack.ts
```diff
@@ -65,194 +69,56 @@ export type StackApprovalUpdate = {
...
+}
+
+type CdktfStackOptions = {
+  stack: SynthesizedStack;
+  onUpdate: (update: StackUpdate | StackApprovalUpdate) => void;
+  onLog?: (log: { message: string; isError: boolean }) => void;
+  autoApprove?: boolean;
+  abortSignal: AbortSignal;
+};
 export class CdktfStack {
```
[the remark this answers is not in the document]
DanielMSchmidt: I tried, but this is not possible. Our TS version can not use the type then anymore to distinguish between the different structures of updates. I'd rather replicate the stackName than interfere with the typing here

packages/cdktf-cli/lib/cdktf-stack.ts
DanielMSchmidt: I tried, but it does not work with our typescript version. The `type` discriminator can not be used anymore to infer the message structure after passed through Omit. I'd rather duplicate stackName then interfering with TS here

packages/cdktf-cli/lib/cdktf-stack.ts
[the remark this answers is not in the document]
DanielMSchmidt: I'd rather leave it to be in sync with cdktf project. I think being in the CLI differentiates enough

packages/cdktf-cli/lib/models/terraform-cli.ts
```diff
@@ -30,14 +30,14 @@ export class TerraformCli implements Terraform {
...
+    createTerraformLogHandler = (_stateName: string) =>
       (_stdout: string, _isErr = false) => {} // eslint-disable-line @typescript-eslint/no-empty-function
   ) {
     this.workdir = stack.workingDirectory;
     this.onStdout = (stateName: string) => (stdout: Buffer) =>
-      onLog(stateName)(stdout.toString());
+      createTerraformLogHandler(stateName)(stdout.toString());
     this.onStderr = (stateName: string) => (stderr: string | Uint8Array) =>
-      onLog(stateName)(stderr.toString(), true);
+      createTerraformLogHandler(stateName)(stderr.toString(), true);
```
ansgarm, line 40: nit: I think `phase` or `command` would be more precise and less colliding than `stateName`

packages/cdktf-cli/lib/models/terraform-cloud.ts
```diff
@@ -123,7 +123,7 @@ export class TerraformCloud implements Terraform {
     public readonly stack: SynthesizedStack,
     public readonly config: TerraformJsonConfigBackendRemote,
     isSpeculative = false,
-    private readonly createLogSender = (_phase: string) =>
+    private readonly createTerraformLogHandler = (_phase: string) =>
```
ansgarm, line 126: here it is `phase` 😁

packages/cdktf-cli/lib/cdktf-stack.ts
```diff
@@ -65,194 +69,56 @@ export type StackApprovalUpdate = {
   reject: () => void;
 };
 
+async function getTerraformClient(
+  abortSignal: AbortSignal,
+  stack: SynthesizedStack,
+  isSpeculative: boolean,
+  sendLog: (stateName: string) => (message: string, isError?: boolean) => void
```
ansgarm, line 76: still called `sendLog` should be `createTerraformLogHandler` + `phase` here as well 🙈

packages/cdktf-cli/lib/cdktf-stack.ts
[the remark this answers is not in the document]
ansgarm: fine with me

packages/cdktf-cli/lib/cdktf-stack.ts
```diff
@@ -266,83 +132,201 @@ export class CdktfStack {
...
 
-  private waitOnMachineDone(): Promise<void> {
-    return new Promise((resolve, reject) => {
-      if (this.currentState === "done") {
-        resolve();
-      }
-
-      if (this.currentState === "error") {
-        reject(new Error(this.stateMachine.state.context.message));
+  private logCallback(
```
ansgarm, line 169: `createTerraformLogHandler` here as well?

ansgarm (approved):
Love it ❤️

PR #1687, opened 2022-04-05
fix(cli): improve cli error handling

ansgarm:
- chore(deps): update yargs type defs
- fix(cli): be explicit about yargs not exiting the process on failure
- fix(cli): make stack positional argument optional
- fix(cli): exit the process on uncaught errors from yargs handlers
- fix(cli): throw Usage error instead of local process.exit()
- refactor(cli): Don't catch errors locally in command handlers as they are now caught properly on a global level in .fail()

### DIFF
--- a/packages/cdktf-cli/bin/cmds/deploy.ts
+++ b/packages/cdktf-cli/bin/cmds/deploy.ts
@@ -2,10 +2,11 @@ import * as yargs from "yargs";
 import { config as cfg } from "@cdktf/provider-generator";
 import { requireHandlers } from "./helper/utilities";
 import { Errors } from "../../lib/errors";
+import { BaseCommand } from "./helper/base-command";
 
 const config = cfg.readConfigSync();
 
-class Command implements yargs.CommandModule {
+class Command extends BaseCommand {
   public readonly command = "deploy [stacks...]";
   public readonly describe = "Deploy the given stacks";
   public readonly aliases = ["apply"];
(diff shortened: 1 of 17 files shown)


ansgarm:
CI currently fails because somehow yargs tries to run the command when returning completions for it (see https://github.com/yargs/yargs/issues/1965). I'm going to implement the workaround mentioned in the issue in a base command class.

packages/cdktf-cli/bin/cmds/deploy.ts
```diff
@@ -2,11 +2,12 @@ import * as yargs from "yargs";
...
 import { requireHandlers } from "./helper/utilities";
 import { Errors } from "../../lib/errors";
+import { BaseCommand } from "./helper/base-command";
 
 const config = cfg.readConfigSync();
 
-class Command implements yargs.CommandModule {
-  public readonly command = "deploy [OPTIONS] <stacks..>";
+class Command extends BaseCommand {
+  public readonly command = "deploy [stacks...]";
```
DanielMSchmidt, line 10: The brackets here puzzle me, do you know their respective meaning?

ansgarm: Yes.. Figuring that out took the most amount of time 😅
Yargs does not support `required: false` as a setting on `.positional()` but it does support specifying them using `[]` instead of `<>` which switches between them being optional or required 😁
Also: The TS types for `.positional()` are wrong, they mention `required` but yargs does not support that there..

PR #1700, opened 2022-04-08
feat(cli): warn users about misaligned CLI and library versions

DanielMSchmidt:
This often a point of confiusion, especially if one accidentally updates the CLI and wonders why the previously working cdktf program does not work anymore

### DIFF
--- a/packages/cdktf-cli/bin/cmds/helper/check-environment.ts
+++ b/packages/cdktf-cli/bin/cmds/helper/check-environment.ts
@@ -1,7 +1,9 @@
 import * as path from "path";
 import * as semver from "semver";
 import { Errors } from "../../../lib/errors";
+import { logger } from "../../../lib/logging";
 import { exec } from "../../../lib/util";
+import { versionNumber } from "./version-check";
 
 function throwIfLowerVersion(
   language: string,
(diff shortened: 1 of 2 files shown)


packages/cdktf-cli/bin/cmds/helper/check-environment.ts
```diff
@@ -64,3 +65,52 @@ export async function checkEnvironment(projectPath = process.cwd()) {
       await checkGoVersion();
   }
 }
+
+export async function verifySimilarLibraryVersion() {
+  let output;
+  try {
+    output = await exec("npm", ["list", "cdktf", "--json"], {
+      env: process.env,
+    });
```
ansgarm, line 74: This only works in TS projects, doesn't it?

packages/cdktf-cli/bin/cmds/helper/check-environment.ts
```diff
@@ -64,3 +65,52 @@ export async function checkEnvironment(projectPath = process.cwd()) {
...
+  const cliVersion = versionNumber();
+
+  if (cliVersion === "0.0.0") {
+    // We are running a development version
+    return;
+  }
+
+  if (semver.major(libVersion) !== semver.major(cliVersion)) {
+    throw Errors.Usage(
+      `The major version of the library (${libVersion}) and the CLI (${cliVersion}) are different. Please update the library to the same major version and regenerate your provider bindings with 'cdktf get'.`
```
ansgarm, line 107: *if you have any. Else / And update your pre-built provider packages.

packages/cdktf-cli/bin/cmds/helper/check-environment.ts
DanielMSchmidt: 🤦 yeah good point, I sort of forgot that we install cdktf through different package managers. I'll restrict the check to TS

DanielMSchmidt: I changed it so we can add support for more languages down the road 👍

packages/cdktf-cli/bin/cmds/helper/check-environment.ts
```diff
@@ -48,19 +49,97 @@ async function checkNodeVersion() {
...
+    () => Promise<string | undefined | void>
+  > = {
+    typescript: getNodeModuleVersion,
+    go: noOp,
+    python: noOp,
+    csharp: noOp,
+    java: noOp,
+  };
+
+  const libVersion = await (getLibraryVersionMap[language] || noOp)();
```
ansgarm, line 121: If we narrow the type of the return value of `getLanguage`, we wouldn't need this `|| noOp`, right?

DanielMSchmidt: Yes, but we would lie to TS since anyone can edit the cdktf.json :) we could use zod here to parse, but I felt like that might be a bit of an overkill

packages/cdktf-cli/bin/cmds/helper/check-environment.ts
```diff
@@ -48,19 +49,97 @@ async function checkNodeVersion() {
...
+  const noOp = async () => {}; // eslint-disable-line @typescript-eslint/no-empty-function
+  const getLibraryVersionMap: Record<
+    string,
+    () => Promise<string | undefined | void>
+  > = {
+    typescript: getNodeModuleVersion,
+    go: noOp,
+    python: noOp,
+    csharp: noOp,
+    java: noOp,
```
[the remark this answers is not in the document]
DanielMSchmidt: Added an issue: https://github.com/hashicorp/terraform-cdk/issues/1703

packages/cdktf-cli/bin/cmds/helper/check-environment.ts
ansgarm: Yeah, true!

ansgarm (approved):

PR #1722, opened 2022-04-14
fix(cli): allow init command to run in a directory containing a README.md

ansgarm:
When creating a new Git repository via Github one can easily add a README.md resulting in an non-empty new git repo in which cdktf init won't work then

Related to #1697


DanielMSchmidt (approved):
🎉🎉🎉🎉

PR #1738, opened 2022-04-21
chore(docs): document feature flags

DanielMSchmidt:
This adds docs and prints `WARNING: The flags excludeStackIdFromLogicalIds & allowSepCharsInLogicalIds are deprecated and will be removed in the next minor release. Please set them to true in your cdktf.json's context map.` if the flags are set to false or not set at all.


ansgarm:
First time we can use the [Vercel Preview](https://terraform-3gu43txoe-hashicorp.vercel.app/cdktf/feature-flags) 🎉 
So nice to have that one ❤️

website/docs/cdktf/feature-flags.mdx
```diff
@@ -0,0 +1,34 @@
...
+
+## `excludeStackIdFromLogicalIds`
+
+## `allowSepCharsInLogicalIds`
+
+## Lifecycle
+
+Feature flags can only be used to change the logic when synthesizing code or CLI features. Since we have prebuilt providers we can not change the logic of the provider generation.
+
+A feature flag will be introduced in a release, adding it with a default `false` value to the templates used in `cdktf init`. The feature flag will graduate to be on by default in the next minor release, we will print a warning if it's false or not set. The feature flag will be removed in the next minor release.
```
ansgarm, line 34: > in the next minor release

Wouldn't this be major? I think semver denotes that differently below 1.0.0 🤔

In my mind we'd graduate them e.g. in the following version jumps
```
0.11.0 -> 0.12.0
1.0.0 -> 2.0.0
```
(which would be major version to me)

... just checked
> Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.

So, [semver](https://semver.org/spec/v1.0.0.html) does not denote it like I thought 😁

DanielMSchmidt: Thanks for checking. I was thinking once we hit 1.0 we can `:%s/minor/major/g` here

ansgarm: Yeah, that would be fine with me 👍

website/docs/cdktf/feature-flags.mdx
```diff
@@ -0,0 +1,39 @@
+---
+page_title: Feature Flags
```
laurapacilio, line 2: Hi! I think we've already got some of this content, so we don't need a new page for it :-). 

We talk about it here: https://www.terraform.io/cdktf/release#enable-features. and then we also have a corresponding section with more detail in the actual config file page: https://www.terraform.io/cdktf/create-and-deploy/configuration-file#enable-features

Would you want to make edits to that content on the config file page instead of adding this new page?

website/docs/cdktf/release/index.mdx
```diff
@@ -22,3 +22,20 @@ CDKTF includes upgrade guides for releases that have breaking changes.
 CDKTF uses feature flags to enable potentially breaking behaviors in a release. Feature flags are disabled by default, so existing projects that do not specify the flag will continue to work as expected with later CDKTF releases. New projects created using `cdktf init` include flags enabling all features available in the release that created the project.
 
 Edit the [`cdktf.json` configuration file](/cdktf/create-and-deploy/configuration-file#enable-features) to disable flags for which you prefer the old behavior, or to add flags to enable new behaviors after upgrading. The CHANGELOG contains a description of new feature flags in each release, and the CDKTF source file [features.ts](https://github.com/hashicorp/terraform-cdk/blob/main/packages/cdktf/lib/features.ts) provides a complete list of all feature flags.
+
+All feature flags are disabled if not set explicitly.
```
laurapacilio, line 26:
```suggestion

CDKTF uses feature flags to enable potentially breaking behaviors in a release. You can use them to change the logic when synthesizing code or CLI features, but they cannot change the logic of the provider generation.

Feature flags are disabled by default, so existing projects that do not specify the flag will continue to work as expected with later CDKTF releases. However, new projects created using `cdktf init` include flags enabling all features available in the release that created the project. 

Edit the [`cdktf.json` configuration file](/cdktf/create-and-deploy/configuration-file#enable-features) to disable flags for which you prefer the old behavior, or to add flags to enable new behaviors after upgrading. The CHANGELOG contains a description of new feature flags in each release, and the CDKTF source file [features.ts](https://github.com/hashicorp/terraform-cdk/blob/main/packages/cdktf/lib/features.ts) provides a complete list of all feature flags.
```

website/docs/cdktf/release/index.mdx
```diff
@@ -22,3 +22,20 @@ CDKTF includes upgrade guides for releases that have breaking changes.
...
+
+### `allowSepCharsInLogicalIds`
+
+When set, '\_' and '-' are allowed in logical ids.
+
+### Lifecycle
+
+Feature flags can only be used to change the logic when synthesizing code or CLI features. Since we have prebuilt providers we can not change the logic of the provider generation.
+
+A feature flag will be introduced in a release, adding it with a default `true` value to the templates used in `cdktf init`. We will print a warning if it's false or not set. The feature flag will be removed in the next minor release.
```
laurapacilio, line 41: Removing this here in favor of combining it with the existing information above :) 
```suggestion
```

website/docs/cdktf/release/index.mdx
```diff
@@ -22,3 +22,20 @@ CDKTF includes upgrade guides for releases that have breaking changes.
 CDKTF uses feature flags to enable potentially breaking behaviors in a release. Feature flags are disabled by default, so existing projects that do not specify the flag will continue to work as expected with later CDKTF releases. New projects created using `cdktf init` include flags enabling all features available in the release that created the project.
 
 Edit the [`cdktf.json` configuration file](/cdktf/create-and-deploy/configuration-file#enable-features) to disable flags for which you prefer the old behavior, or to add flags to enable new behaviors after upgrading. The CHANGELOG contains a description of new feature flags in each release, and the CDKTF source file [features.ts](https://github.com/hashicorp/terraform-cdk/blob/main/packages/cdktf/lib/features.ts) provides a complete list of all feature flags.
+
+All feature flags are disabled if not set explicitly.
+
+### `excludeStackIdFromLogicalIds`
```
laurapacilio, line 28: Are these two things feature flags for the next release? If they are, I wonder if we can remove them, as we tell folks that we list all feature flags and descriptions in the changelog. Thoughts?

website/docs/cdktf/release/index.mdx
laurapacilio, line 26: Quick question for you --> I want to add back in that "we remove feature flags in the next minor release" but then what happens? When we remove the flags, do folks just need to accept the new functionality as is if they upgrade further? What happens to older projects when they upgrade to the next minor version and the flags are removed? Do they still work as expected?

laurapacilio (changes requested):
We're close :) A couple of final questions/nits

website/docs/cdktf/release/index.mdx
DanielMSchmidt: Yes, users have to live with the new functionality (if we don't hear about catastrophic shortcomings). Once you upgrade to a version where the flag is removed you get the new feature as if you added it to the feature flags already

website/docs/cdktf/release/index.mdx
```diff
@@ -17,8 +17,25 @@ CDKTF includes upgrade guides for releases that have breaking changes.
...
+
+Feature flags are disabled by default, so existing projects that do not specify the flag will continue to work as expected with later CDKTF releases. However, new projects created using `cdktf init` include flags enabling all features available in the release that created the project.
 
 Edit the [`cdktf.json` configuration file](/cdktf/create-and-deploy/configuration-file#enable-features) to disable flags for which you prefer the old behavior, or to add flags to enable new behaviors after upgrading. The CHANGELOG contains a description of new feature flags in each release, and the CDKTF source file [features.ts](https://github.com/hashicorp/terraform-cdk/blob/main/packages/cdktf/lib/features.ts) provides a complete list of all feature flags.
+
+### `excludeStackIdFromLogicalIds`
+
+When set, the stack id is excluded from generated logical ids for constructs within the stack. This means constructs that are top-level to the stack will have a logical id that is the same as their construct id.
+When you use nested constructs the mechanism will add a suffix to the logical id to make it unique.
```
laurapacilio, line 31:
```suggestion
When you use nested constructs, the mechanism will add a suffix to the logical id to make it unique.
```

website/docs/cdktf/release/index.mdx
```diff
@@ -17,8 +17,25 @@ CDKTF includes upgrade guides for releases that have breaking changes.
...
+
+### `allowSepCharsInLogicalIds`
+
+When set, '\_' and '-' are allowed in logical ids.
+
+### Lifecycle
+
+Feature flags can only be used to change the logic when synthesizing code or CLI features. Since we have prebuilt providers we can not change the logic of the provider generation.
+
+A feature flag will be introduced in a release, adding it with a default `true` value to the templates used in `cdktf init`. We will print a warning if it's false or not set. The feature flag will be removed in the next minor release and the standard behaviour is the feature flag set to `true`.
```
laurapacilio, line 41:
```suggestion
```

website/docs/cdktf/release/index.mdx
```diff
@@ -17,8 +17,25 @@ CDKTF includes upgrade guides for releases that have breaking changes.
 - [Upgrading to v0.7](/cdktf/release/upgrade-guide-v0-7)
 - [Upgrading to v0.6](/cdktf/release/upgrade-guide-v0-6)
 
-## Enable Features
+## Features Flags for 0.11
 
-CDKTF uses feature flags to enable potentially breaking behaviors in a release. Feature flags are disabled by default, so existing projects that do not specify the flag will continue to work as expected with later CDKTF releases. New projects created using `cdktf init` include flags enabling all features available in the release that created the project.
+CDKTF uses feature flags to enable potentially breaking behaviors in a release. You can use them to change the logic when synthesizing code or CLI features, but they cannot change the logic of the provider generation.
```
laurapacilio, line 22: Moving this up, as I think it can be combined with the introduction. But question @DanielMSchmidt - the sentence about how the default is to be set to `true` feels like it directly contradicts the paragraph below this about how how they're disabled by default... help :) 

```suggestion
CDKTF uses feature flags to enable potentially breaking behaviors in a release. You can use them to change the logic when synthesizing code or CLI features, but they cannot change the logic of the provider generation. A feature flag will be introduced in a release, adding it with a default `true` value to the templates used in `cdktf init`. We will print a warning if it's false or not set. Feature flags will be removed in the next minor release, and the standard behavior is the feature flag set to `true`. 
```

laurapacilio (changes requested):
a few language things and then i had a question :)

packages/cdktf-cli/lib/synth.ts
```diff
@@ -51,3 +52,29 @@ export function printAnnotations(stacks: SynthesizedStack[]) {
...
   }
 }
+
+export function printWarningsForDeprecatedFlags() {
+  const context: Record<string, boolean | undefined> = JSON.parse(
+    process.env[CONTEXT_ENV] || "{}"
+  );
+
+  // We want to print a warning for every deprecated flag that is set to false or not set
+  const deprecatedFlagsInUse = DEPRECATED_FLAGS.filter(
```
ansgarm, line 62: I'm a bit confused about the naming "in use" and the filtered flags being either not set or set to undefined.

I suspect the underlying problem I'm having in understanding is that there might be two possible situations when a feature flag might get deprecated:
1. The behaviour will be adopted (this is what is currently handled by the code; in this case we want to warn the user that they should set it to `true` to adopt the behaviour, right?)
2. The behaviour will be dropped (this is not yet handled, but in that case we should advise the user to set it to `false` to switch the the behaviour that is going to apply in future versions)

With that in mind, I think we should change the wording in the error message, shouldn't we?

ansgarm: Ah, I just read the docs further down. They explain everything 🙈

I'm still a bit confused about how the exact process of "promoting" a feature / feature flag works. It feels that there is a part missing or the wording is just a bit inconsistent for my taste leading to false assumptions (probably around deprecating a feature flag which will turn on the feature by default in the future) but I can't really grasp it.

DanielMSchmidt: So the process is we introduce something new, on by default if you create a new project. By seeing the flag to false you can get back to the old behavior until we remove the flag. This is helpful to postpone the change needed by a little or to work around bugs on our end

website/docs/cdktf/release/index.mdx
```diff
@@ -18,8 +18,19 @@ CDKTF includes upgrade guides for releases that have breaking changes.
...
+
+Feature flags are disabled by default, so existing projects that do not specify the flag will continue to work as expected with later CDKTF releases. However, new projects created using `cdktf init` include flags enabling all features available in the release that created the project.
 
 Edit the [`cdktf.json` configuration file](/cdktf/create-and-deploy/configuration-file#enable-features) to disable flags for which you prefer the old behavior, or to add flags to enable new behaviors after upgrading. The CHANGELOG contains a description of new feature flags in each release, and the CDKTF source file [features.ts](https://github.com/hashicorp/terraform-cdk/blob/main/packages/cdktf/lib/features.ts) provides a complete list of all feature flags.
+
+### `excludeStackIdFromLogicalIds`
+
+When set, the stack id is excluded from generated logical ids for constructs within the stack. This means constructs that are top-level to the stack will have a logical id that is the same as their construct id.
+When you use nested constructs, the mechanism will add a suffix to the logical id to make it unique.
```
laurapacilio, line 32:
```suggestion
When set, CDKTF excludes the stack id from generated logical ids for constructs within the stack. Constructs that are top-level to the stack have a logical id that is the same as their construct id. When you use nested constructs, CDKTF adds a suffix to the logical id to make it unique.
```

website/docs/cdktf/release/index.mdx
```diff
@@ -18,8 +18,19 @@ CDKTF includes upgrade guides for releases that have breaking changes.
...
+
+### `allowSepCharsInLogicalIds`
+
+When set, '\_' and '-' are allowed in logical ids.
```
laurapacilio, line 36:
```suggestion
When set, CDKTF allows '\_' and '-' in logical ids.
```

laurapacilio (approved):
Some tiny wording edits - heh, I had to since this was still open.

PR #1752, opened 2022-04-26
fix(release): Use non default terraform versions as specified in test matrix for Release and Release Next workflows

ansgarm:
aligns with ci that runs on PR (and already does this)


DanielMSchmidt (approved):

DanielMSchmidt:
Merging as is as the changes only take effect after the merge on release

PR #1757, opened 2022-04-27
chore(cli): add environment variable to disable version checks

DanielMSchmidt:
This env var could be helpful in case of bugs in the version detection

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,7 +19,6 @@ jobs:
       image: docker.mirror.hashicorp.services/hashicorp/jsii-terraform
     env:
       CHECKPOINT_DISABLE: "1"
-      DISABLE_VERSION_CHECK: "1"
     steps:
       - uses: actions/checkout@v2
       - name: installing dependencies
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,7 +19,7 @@ jobs:
       image: docker.mirror.hashicorp.services/hashicorp/jsii-terraform
     env:
       CHECKPOINT_DISABLE: "1"
-
+      DISABLE_VERSION_CHECK: "1"
     steps:
       - uses: actions/checkout@v2
       - name: installing dependencies
--- a/packages/cdktf-cli/bin/cmds/helper/check-environment.ts
+++ b/packages/cdktf-cli/bin/cmds/helper/check-environment.ts
@@ -55,16 +55,19 @@ export async function checkEnvironment() {
 }
 
 export async function verifySimilarLibraryVersion() {
+  if (process.env.DISABLE_VERSION_CHECK) {
+    logger.debug("Version check disabled via Environment variable");
+    return;
+  }
+
   const language = getLanguage();
   if (!language) {
-    // We could not detect the language, disabling the check
     logger.debug("Unable to detect language, skipping version check");
     return;
   }
 
   const libVersion = await getPackageVersion(language, "cdktf");
   if (!libVersion) {
-    // We could not detect the library version, disabling the check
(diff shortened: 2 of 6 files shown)


.github/workflows/release.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       image: docker.mirror.hashicorp.services/hashicorp/jsii-terraform
     env:
       CHECKPOINT_DISABLE: "1"
-
+      DISABLE_VERSION_CHECK: "1"
```
ansgarm, line 22: While this unblocks us in the short term, I'd rather not do this and try to figure out and fix the underlying issue.

website/docs/cdktf/create-and-deploy/environment-variables.mdx
```diff
@@ -0,0 +1,58 @@
...
+
+## CDKTF_LOG_LEVEL
+
+Can be set to one of the following values:
+
+- `debug`
+- `info`
+- `warn`
+- `error`
+- `fatal`
```
ansgarm, line 36: It also supports `all`, right?

website/docs/cdktf/create-and-deploy/environment-variables.mdx
```diff
@@ -0,0 +1,58 @@
...
+
+If set this token will be used to authenticate against Terraform Cloud, it needs to be set to a valid [Terraform Cloud API token](/cloud-docs/users-teams-organizations/api-tokens).
+
+## http_proxy or HTTP_PROXY
+
+If set this proxy will be used for all HTTP requests.
+
+## DISABLE_VERSION_CHECK
+
+If set to a not-empty value, no version check will be performed. If it is not set we will verify your installed library version matches the CLI version and warn if they do not match.
```
ansgarm, line 58:
```suggestion
If set to a non-empty value, no version check will be performed. If it is not set we will verify your installed library version matches the CLI version and warn if they do not match.
```

packages/cdktf-cli/bin/cmds/helper/check-environment.ts
```diff
@@ -55,24 +55,26 @@ export async function checkEnvironment() {
 }
 
 export async function verifySimilarLibraryVersion() {
+  if (process.env.DISABLE_VERSION_CHECK) {
+    logger.debug("Version check disabled via Environment variable");
```
ansgarm, line 59:
```suggestion
    logger.debug("Version check disabled via environment variable");
```

packages/cdktf-cli/bin/cmds/helper/check-environment.ts
```diff
@@ -55,24 +55,26 @@ export async function checkEnvironment() {
 }
 
 export async function verifySimilarLibraryVersion() {
+  if (process.env.DISABLE_VERSION_CHECK) {
```
ansgarm, line 58: Should we allow `"false"` to not disable the version check?
```suggestion
  if (process.env.DISABLE_VERSION_CHECK && process.env.DISABLE_VERSION_CHECK !== "false") {
```

ansgarm (approved):

packages/cdktf-cli/bin/cmds/helper/check-environment.ts
DanielMSchmidt: I read through every one of our `process.env` calls today and none have a check like this. We have one for logging that checks if it's === false. I believe we should overhaul what we currently have and check if it's set to 1 or true, but I'd rather do it globally after this merged

ansgarm: fine with me

PR #1761, opened 2022-04-29
feat(cli): add cdktf provider add command

### LINKED ISSUE #877: CLI command for provider install/upgrade
### Community Note

### Description

Every time I create a new CDK for Terraform project for the sake of reproducing a GitHub issue, I find myself going through the same loop of

1. Looking up the the current version of the provider
2. checking if we have a pre-built provider
3. editing cdktf.json
3. running get
4. add relevant imports

I would like to be able to run something like the following:

```
$> cdktf provider search kubernetes

Searching registry.terraform.io for "kubernetes" providers...
Found
hashicorp/kubernetes (official) https://registry.terraform.io/providers/hashicorp/kubernetes
hashicorp/kubernetes-alpha (official) https://registry.terraform.io/providers/hashicorp/kubernetes-alpha
redux/kubernetes (community) https://registry.terraform.io/providers/redeux/kubernetes
...

$> cdktf provider install hashicorp/kubernetes

Installing hashicorp/kubernetes provider
Found pre-built provider
Use pre-built provider? (no)
Added  "hashicorp/kubernetes@~> 2.3.2" to cdktf.json
Generating provider...
hashicorp/kubernetes provider bindings generated in .gen/providers/kubernetes

Add import to main.ts? (yes)
// cdktf adds the following to main.ts below existing imports:
// import { KubernetesProvider } from "./.gen/providers/kubernetes";

Add provider to main.ts? (yes)
// cdktf adds `new KubernetesProvider(this, "kind", {})` to the constructor in main.ts
// maybe also add a comment explicitly linking to the registry page etc eg:
// // hashicorp/kubernetes provider (https://registry.terraform.io/providers/hashicorp/kubernetes/latest) 
```

Ideally, I would also like to be able to help me upgrade providers, e.g.

```
$> cdktf provider update hashicorp/kubernetes@latest
This project currently uses hashicorp/kubernetes 2.1.1. 2.3.2 is the latest available version. Upgrade? (yes)
// cdktf updates cdktf.json, gets it, gets the schema

Warning: the cron-job resources in hashicorp/kubernetes is in use in ./foo/bar/cron-job.ts, and the schema has changed between 2.1.1 and 2.3.2. You may need to modify this resource. Continue? (yes)
```

### References

The terraform registry currently lacks a suitable search API. The Terraform VSCode extensions wants similar functionality to support provider install/upgrade workflows, so collaboration with that team to add this to the Terraform registry would be needed to build the search portion of this functionality.

ansgarm:
Closes #877

### Todo (incomplete, only stuff left to do)
- [x] install packages in Python projects
- [x] install packages in C# projects
- [x] install packages in Java projects
- [x] write integration tests for `provider add` command (local execution)
- [x] write integration tests for `provider add` command (pre-built for TypeScript)
- [x] write integration tests for `provider add` command (pre-built for C#)
- [x] write integration tests for `provider add` command (pre-built for Java)
- [x] support adding multiple providers at the same time
- [x] write integration tests for `provider add` command (pre-built for Python)
- [x] add output like "added provider to cdktf.json" when adding locally (like it already exists for when pre-built provider packages are installed)
- [x] check cdktf version when evaluating pre-built providers
- [x] write unit tests for provider name parsing
- [x] write unit tests for provider version parsing
- [x] test that local add runs get (and run get)
- [x] make sure integration tests also get an `onWindows` case for the same stuff that is tested on posix
- [x] add utility that does fetch requests with a cdktf-cli user-agent
- [x] cache calls to NPM Registry (became a follow-up [issue](https://github.com/hashicorp/terraform-cdk/issues/1815))
- [ ] write docs for provider add command & add examples mentioning e.g. =,~>.> etc. (add to [this page](https://github.com/hashicorp/terraform-cdk/blob/main/website/docs/cdktf/cli-reference/commands.mdx))

### DIFF
--- a/packages/@cdktf/hcl2cdk/test/convertProject.test.ts
+++ b/packages/@cdktf/hcl2cdk/test/convertProject.test.ts
@@ -69,13 +69,13 @@ app.synth();`,
           "node": ">=14.0"
         },
         "dependencies": {
-          "cdktf": "next",
+          "cdktf": "0.11.0-pre.53",
           "constructs": "^10.0.5"
         },
         "devDependencies": {
           "@types/node": "^14.0.26",
           "typescript": "^3.9.7",
-          "cdktf-cli": "next"
+          "cdktf-cli": "0.11.0-pre.53"
         }
       }`,
     ],


packages/cdktf-cli/lib/dependencies/dependency-manager.ts
```diff
@@ -0,0 +1,102 @@
...
+      return false;
+    }
+
+    return Promise.resolve(false);
+  }
+
+  async addPrebuiltProvider(constraint: ProviderConstraint) {
+    constraint;
+    if (this.targetLanguage === Language.GO) {
+      throw Errors.Usage(
```
DanielMSchmidt, line 64:
```suggestion
      throw Errors.Internal(
```

Oder? Du solltest das ja vorher abgechecked haben

ansgarm: I thought about a `--pre-built` flag or similar, which might directly invoke `addPrebuiltProvider()` and in which case I would not have caught it before

packages/cdktf-cli/lib/dependencies/package-manager.ts
```diff
@@ -0,0 +1,146 @@
...
+      case Language.GO:
+        throw new Error("Go is not supported yet");
+      case Language.TYPESCRIPT:
+        return new NodePackageManager(workingDirectory);
+      case Language.PYTHON:
+        return new PythonPackageManager(workingDirectory);
+      case Language.CSHARP:
+        return new NugetPackageManager(workingDirectory);
+      case Language.JAVA:
+        return new MavenPackageManager(workingDirectory);
```
jsteinich, line 31: We should setup this up in a way that it is easy to switch out a different package manager. For example, Java has multiple available package managers. Maven is one of the most common and what we support elsewhere, but Gradle is also very common.

I believe a similar situation exists for python and we already support multiple methods with the templates.

test/csharp/provider-add-command/test.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  describe("pre-built", () => {
+    beforeAll(async () => {
+      driver = new TestDriver(__dirname, { CDKTF_DIST: "" }); // reset CDKTF_DIST set by run-against-dist script
+      await driver.setupCsharpProject({
+        init: { additionalOptions: "--cdktf-version 0.10.4" },
+      });
+    });
+
+    onPosix("installs pre-built provider using nuget", async () => {
```
jsteinich, line 14: Why not run theses tests on Windows as well?

packages/cdktf-cli/lib/dependencies/package-manager.ts
ansgarm: You're right. However, currently we only support the package managers, that our template supports. However, for example, the "NodePackageManager" (poor wording) does check for a `yarn.lock` to switch between yarn and npm.
But we could split that into two separate classes in the future and introduce an abstract "isProjectUsingThisPackageManager" method that those classes also implement 😁 But I see that as out of scope for the first iteration and I think the current design is easy enough to be refactored in such a way that we don't need to adjust them right now to enable to this kind of change in the future.

test/csharp/provider-add-command/test.ts
ansgarm: No reason, added a todo for this to this PR.
I started with onPosix because I can create those snapshots locally, will add onWindows later on before merging this 👍

packages/cdktf-cli/bin/cdktf.ts
```diff
@@ -79,6 +79,8 @@ yargs
   .command(require("./cmds/watch"))
   .command(require("./cmds/output"))
   .command(require("./cmds/debug"))
+  .command(require("./cmds/provider"))
+  // .command(require("./cmds/provider-add"))
```
DanielMSchmidt, line 83:
```suggestion
```

packages/cdktf-cli/bin/cmds/handlers.ts
```diff
@@ -167,19 +173,15 @@ export async function diff(argv: any) {
   );
 }
 
-export async function get(argv: any) {
+export async function get(argv: { output: string; language: Language }) {
```
DanielMSchmidt, line 176: Nice improvement, from type cast to type assertion

packages/cdktf-cli/lib/cdktf-config.ts
```diff
@@ -0,0 +1,89 @@
...
+import { CONFIG_DEFAULTS } from "@cdktf/provider-generator/lib/config";
+
+export type ProviderDependencySpec = {
+  name: string;
+  source: string;
+  version?: string;
+  namespace?: string;
+};
+
+// TODO: move this to some common package from where it is used by @cdktf/provider-generator, cdktf, and cdktf-cli
```
DanielMSchmidt, line 16: Nit: sounds good, let's move it and refactor all the other places at the same time, because I wrote this quite a few times already :D

packages/cdktf-cli/lib/dependencies/dependency-manager.ts
```diff
@@ -0,0 +1,260 @@
...
+  getNpmPackageName,
+  getPrebuiltProviderVersion,
+} from "./prebuilt-providers";
+import { getLatestVersion } from "./registry-api";
+import { versionMatchesConstraint } from "./version-constraints";
+
+// ref: https://www.terraform.io/language/providers/requirements#source-addresses
+const DEFAULT_HOSTNAME = "registry.terraform.io";
+const DEFAULT_NAMESPACE = "hashicorp";
+function normalizeProviderSource(source: string) {
```
DanielMSchmidt, line 18: Isn't this quite similar? https://github.com/hashicorp/terraform-cdk/blob/main/packages/%40cdktf/provider-generator/lib/config.ts#L75

packages/cdktf-cli/lib/dependencies/dependency-manager.ts
```diff
@@ -0,0 +1,260 @@
...
+    await new CdktfConfigManager().addProvider(constraint);
+  }
+
+  /**
+   * Converts an NPM package name of a pre-built provider package to the name in the target language
+   */
+  private convertPackageName(name: string): string {
+    switch (this.targetLanguage) {
+      case Language.GO:
+        throw new Error("pre-built providers are not supported for Go yet");
```
DanielMSchmidt, line 239: Can we link an issue to upvote here? And shouldn't there be a hint on how to add a provider for go? Lets say you run `cdktf provider add aws` then we find it's a prebuilt one and end up here where it explodes, how would you know you have to run `cdktf provider add aws --force-local`?

packages/cdktf-cli/lib/dependencies/package-manager.ts
```diff
@@ -0,0 +1,217 @@
...
+      default:
+        throw new Error(`Unknown language: ${language}`);
+    }
+  }
+
+  public abstract addPackage(
+    packageName: string,
+    packageVersion?: string
+  ): Promise<void>;
+  // public abstract listPackages(): Promise<todo>; future stuff..
```
DanielMSchmidt, line 41: Nit: Should these stay?

packages/cdktf-cli/bin/cmds/handlers.ts
```diff
@@ -360,3 +362,41 @@ export async function debug(argv: any) {
...
   }
 }
+
+export async function providerAdd(argv: any) {
+  const config = CdktfConfig.read();
+
+  const language = config.language;
+  const cdktfVersion = await getPackageVersion(language, "cdktf");
+
+  if (!cdktfVersion) throw Errors.External("Could not determine cdktf version"); // TODO: allow this? Only use local then?
```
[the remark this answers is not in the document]
ansgarm: > throwing an error

Did you forgot "User" error?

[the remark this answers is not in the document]
DanielMSchmidt: I mean yeah we can make it a user error, but I was mostly arguing agains the comment below that it should be adding a local provider instead

packages/cdktf-cli/lib/dependencies/dependency-manager.ts
ansgarm: Similar as in "we should refactor this into a single place" yes, but the one you linked also parses versions etc. Also: the add command currently does not support local providers or providers from VCS, so that's something for the future as well :)

packages/cdktf-cli/lib/dependencies/dependency-manager.ts
ansgarm: This error is not supposed to be surfaced to the user. The behaviour you are describing is happening [here](https://github.com/hashicorp/terraform-cdk/blob/8bd931729fc8bc6716e3c750ce7fd5144da026ec/packages/cdktf-cli/lib/dependencies/dependency-manager.ts#L184) and [here](https://github.com/hashicorp/terraform-cdk/blob/8bd931729fc8bc6716e3c750ce7fd5144da026ec/packages/cdktf-cli/lib/dependencies/dependency-manager.ts#L156-L161). Although the first one currently won't be reached, because we have no "force-pre-built" flag, that might directly jump into that method.
Also, you don't have to run with `--force-local` for Go, it will automatically fall back to local after printing the info that Go has no pre-built providers yet.

packages/cdktf-cli/lib/dependencies/package-manager.ts
ansgarm: would be fine with me, but I have no strong opinion at all here :)

packages/cdktf-cli/bin/cmds/handlers.ts
ansgarm: yeah, I moved that upwards, because I started calling that method from elsewhere, so I wanted those assumptions to be in the arguments 😁

packages/cdktf-cli/lib/dependencies/package-manager.ts
DanielMSchmidt: me neither

packages/cdktf-cli/lib/dependencies/dependency-manager.ts
DanielMSchmidt: Ah so it's solved on the prebuilt detection level? (then it would be `Errors.Internal` right?)

packages/cdktf-cli/lib/dependencies/dependency-manager.ts
DanielMSchmidt: Oh I didn't know, we need to make sure we document this and possible detect it and throw errors.

packages/cdktf-cli/lib/cdktf-config.ts
DanielMSchmidt: Yeah sounds like a good idea

packages/cdktf-cli/bin/cmds/handlers.ts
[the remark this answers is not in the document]
ansgarm: Ah, I see. Will improve the error message then.

packages/cdktf-cli/lib/dependencies/dependency-manager.ts
ansgarm: yeah, it could be errors internal – will change it accordingly 👍

packages/cdktf-cli/bin/cmds/handlers.ts
```diff
@@ -40,13 +40,19 @@ import {
...
+  DependencyManager,
+  ProviderConstraint,
+} from "../../lib/dependencies/dependency-manager";
+import { CdktfConfig } from "../../lib/cdktf-config";
 
 const chalkColour = new chalk.Instance();
 const config = cfg.readConfigSync();
 
 async function getProviderRequirements(provider: string[]) {
+  // TODO: replace with CdktfConfig calls
```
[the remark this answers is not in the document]
ansgarm: implemented

packages/cdktf-cli/lib/cdktf-config.ts
ansgarm: added https://github.com/hashicorp/terraform-cdk/issues/1814 and linked it in the comment

PR #1768, opened 2022-05-04
fix(docs): remove outdated docs from the previous watch implementation

ansgarm:


website/docs/cdktf/cli-reference/commands.mdx
```diff
@@ -465,8 +465,6 @@ Whenever a file matching the watch pattern is changed, the command will run `cdk
 
 Before using `watch` you should:
 
-- **Check your root `.gitignore` file.** Your root `.gitignore` must contain all files that are generated by `synth` and `deploy`. If that is not the case, the `watch` command will trigger itself and continuously try to deploy. Although Git supports nesting `.gitignore` files, the `watch` command currently only reads the root `.gitignore` file.
-
 - **Check your environment.** The `watch` command should only be used for development environments. We recommend making sure that the terminal where you want to run `watch` has no access keys that allow the `cdktf-cli` to deploy to your production environment.
```
laurapacilio, line 468: I think we should make this a paragraph instead of a single bullet. A list with only one item isn't a list! :-)

DanielMSchmidt (approved):

PR #1771, opened 2022-05-04
feat(release): support running releases on backport-release branches besides only on main

ansgarm:
Backported from #1770


ansgarm:
CI fails due to the issue that we're trying to backport – will merge anyway as only a Github Actions file was touched.

Issue #1656, opened 2022-03-24
CDKTF dependsOn broken when using overrideLogicalId


plaisted:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### cdktf & Language Versions

cdktf@0.9.4

### Affected Resource(s)

All.

### Expected Behavior

Resources using `overrideLogicalId()` should function when referenced by other resources.

### Actual Behavior

Resources using  `overrideLogicalId()` update the TF resource name to the new name but no longer properly handle references to them. This causes TF plan/apply failures shown below (`sqs-lambda-drop_dlq` is the TF resource name after using `overrideLogicalId()`, depends_on is attempting to reference `sqs-lambda-drop_dlq_EF847697` (the original name before override))

│ Error: Reference to undeclared resource
│ 
│   on redacted.tf.json line 452, in resource.aws_sqs_queue.sqs-lambda-drop_queue.depends_on:
│  452:           "aws_sqs_queue.sqs-lambda-drop_dlq_EF847697"
│ 
│ A managed resource "aws_sqs_queue" "sqs-lambda-drop_dlq_EF847697" has not
│ been declared in the root module.
╵

### Steps to Reproduce

1. Create a TF resource.
2. Override the logical name using overrideLogicalId()
3. Create a second TF resource and have it explicitly depend on the first (TF depends_on)
4. Run terraform plan

### Important Factoids

This may have regressed in https://github.com/hashicorp/terraform-cdk/commit/88f852bd82a7bb9391302ffbac6eb026f2b3bfbb .  The commit changes the FQN to be calculated in the constructor instead of lazily when requested which appears to be used by other resources when creating their depends on list.

### References

- https://github.com/hashicorp/terraform-cdk/commit/88f852bd82a7bb9391302ffbac6eb026f2b3bfbb

jsteinich:
@DanielMSchmidt do you recall the reason for moving `fqn` to the constructor. I can see it as an optimization, but it's not immediately apparent what other reason there is.

edge7:
Is this fixed also for the Python version?
[...]

PR #1725, opened 2022-04-19
feat(provider-generator): Use `ComplexList` for any complex list

jsteinich:
Closes #1610 
Fixes #1447
Fixes #1371
Fixes #1363
Fixes #1520
Fixes #1684
Fixes #25
Fixes #1698
Fixes #434
Fixes #518
Fixes #663
Fixes #1087

See #1610 for more information.

This is technically a breaking change. Work has been done to reduce the likelihood of impacting users; however, return types for any list/set/map of complex properties is changing.

PR #1779, opened 2022-05-06
fix: Avoid unstable fqn for TerraformElements

x3cion:
Hi,

when using `dependsOn` to depend on a resource with overwritten logical id, the fqn does not fit the resource anymore, as it is generated within the constructor. I took an approach to stabilizing the fqn and also avoiding potential problems when using the fqn before changing the id.

This change might be a breaking change.

Let me know if you need any more infos!

### DIFF
--- a/packages/cdktf/lib/terraform-element.ts
+++ b/packages/cdktf/lib/terraform-element.ts
@@ -1,6 +1,8 @@
+import { ok } from "assert";
 import { Construct } from "constructs";
 import { Token } from ".";
 import { TerraformStack } from "./terraform-stack";
+import { ref } from "./tfExpression";
 
 export interface TerraformElementMetadata {
   readonly path: string;
@@ -17,8 +19,14 @@ export class TerraformElement extends Construct {
    */
   private _logicalIdOverride?: string;
 
-  constructor(scope: Construct, id: string) {
+  /**
+   * Type of this element, used for fqn
+   */
+  private readonly _elementType?: string;
+
+  constructor(scope: Construct, id: string, elementType?: string) {
     super(scope, id);
+    this._elementType = elementType;
 
     if (Token.isUnresolved(id)) {
       throw new Error(
@@ -38,12 +46,29 @@ export class TerraformElement extends Construct {
     return {};
   }
 
-  public get friendlyUniqueId() {
-    if (this._logicalIdOverride) {
-      return this._logicalIdOverride;
-    } else {
-      return this.cdktfStack.getLogicalId(this);
+  private _fqnToken?: string;
+
+  public get fqn() {
+    if (!this._fqnToken) {
+      ok(!!this._elementType, "Element type not set");
+      this._fqnToken = Token.asString(
+        ref(`${this._elementType}.${this.friendlyUniqueId}`, this.cdktfStack)
+      );
     }
+    return this._fqnToken;
+  }
+
+  private _friendlyUniqueId?: string;
+
+  public get friendlyUniqueId() {
+    if (!this._friendlyUniqueId) {
+      if (this._logicalIdOverride) {
+        this._friendlyUniqueId = this._logicalIdOverride;
+      } else {
+        this._friendlyUniqueId = this.cdktfStack.getLogicalId(this);
+      }
+    }
+    return this._friendlyUniqueId;
   }
 
   /**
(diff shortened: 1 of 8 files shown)


jsteinich:
@x3cion thanks for putting this together.
One thing I'm curious about is whether fqn needs to be cached at all. A previous implementation didn't have it that way. I'm thinking there is a case where it matters, but the structure you put in place would make it easy to test that out.

This should resolve #1656.

x3cion:
I looked through the child classes and the only exception I found to the fqn generation was `TerraformProvider`, which uses an alias. Since the setter and getter for alias stated some weird usage, I didn't try to dig deeper for this case.

My reasoning behind caching the token was, considering how `Token.asString()` works, a new token would be created for every call to fqn. At the same time my understanding is, fqn's have to be stable identifiers, so the cached `_fqnToken` is used to disallow changing the fqn base values after fqn has been called once, uncovering a possible issue.

Let me know if I should change the implementation.

jsteinich (approved):

jsteinich:
> a new token would be created for every call to fqn

The internals of `Token.asString` do cache values so that the same token should be returned.

> he only exception I found to the fqn generation was `TerraformProvider`, which uses an alias

`TerraformProvider.fqn` is only used during synthesis (should probably be private), so that should be fine to leave as is.

> At the same time my understanding is, fqn's have to be stable identifiers, so the cached `_fqnToken` is used to disallow changing the fqn base values after fqn has been called once, uncovering a possible issue.

That is correct. Good catch there.

I'll need to make some adjustments to #1725 to account for these changes, but that will probably be a separate PR after they are both merged.

x3cion:
@jsteinich the point about `Token.asString` made me think, so I tried it out. I changed the code:
https://github.com/x3cion/terraform-cdk/blob/fiddle/unstable-fqn/packages/cdktf/lib/terraform-element.ts#L54

This makes the test fail:
```
  ● fqn is stable

    expect(received).toBe(expected) // Object.is equality

    Expected: "${TfToken[TOKEN.0]}"
    Received: "${TfToken[TOKEN.1]}"

      76 |   const elementWithFQN = new TerraformElement(stack, "test", "valid_type");
      77 |   const fqn = elementWithFQN.fqn;
    > 78 |   expect(elementWithFQN.fqn).toBe(fqn);
         |                              ^
      79 |
      80 |   // May not override logical id after fqn has been requested
      81 |   expect(() => elementWithFQN.overrideLogicalId("new-id")).toThrow();

      at Object.<anonymous> (test/resource.test.ts:78:30)
```

I was also guessing it works like that when I investigated the code. The entry is actually only cached on the token itself by `cachedValue`, but there is no global cache as far as I see. Could this uncover some other problems?

packages/cdktf/lib/terraform-element.ts
```diff
@@ -17,8 +19,14 @@ export class TerraformElement extends Construct {
    */
   private _logicalIdOverride?: string;
 
-  constructor(scope: Construct, id: string) {
+  /**
+   * Type of this element, used for fqn
+   */
```
ansgarm, line 24: I'm, wondering whether we should add a note why the `elementType` may be undefined. If I saw that correctly, that was for example the case for `Provider` constructs, right?

packages/cdktf/lib/terraform-element.ts
```diff
@@ -38,26 +46,51 @@ export class TerraformElement extends Construct {
     return {};
   }
 
+  private _fqnToken?: string;
+
+  public get fqn() {
+    if (!this._fqnToken) {
+      ok(!!this._elementType, "Element type not set");
```
ansgarm, line 53: nice! I like this. We should use the `assert` package more often 😇

ansgarm (commented):
Thank you for improving this @x3cion 🎉
Found only two small things, nothing major.

packages/cdktf/lib/terraform-element.ts
x3cion: We switched from if's to ok's, no regrets, it's easy to read and smaller at the same time, win-win! :)

packages/cdktf/lib/terraform-element.ts
x3cion: I added a bit more info in the comment, please check it out

ansgarm: looks good 👍

ansgarm (approved):

ansgarm:
@x3cion Let's get this merged! Do you want to rebase this branch onto `main` or should I?

x3cion:
@ansgarm I rebased now, you'll need to merge, though. :)

ansgarm:
Will do, once CI passed 👍

PR #1786, opened 2022-05-10
feat(provider-generator): Add `TerraformModuleOptions` to the generated source code

sonodar:
Add TerraformModuleOptions to the source code of the module generated by module-generator.

Motivation:

I want to use the [ACM module](https://registry.terraform.io/modules/terraform-aws-modules/acm/aws/latest) in the `us-east-1` region for CloudFront, but I can't do that without splitting the stack because I don't have the providers parameter.


ansgarm (commented):
Thank you for the PR @sonodar! Nice improvement 👍

I think for this to fully work, you might also need to adjust this part of the generator:
https://github.com/hashicorp/terraform-cdk/blob/3a387c44a4d4f4b05fa8143bf0a5693be45168dd/packages/%40cdktf/provider-generator/lib/get/generator/module-generator.ts#L66-L71
and pass `providers` and `dependsOn` to the `super()` call. Else the `TerraformModule` parent class would not know about them.

But other than that, it looks good to go 👍

ansgarm:
Just saw that [CI failed](https://github.com/hashicorp/terraform-cdk/runs/6369093603?check_suite_focus=true#step:7:62) with an JSII error. It seems like JSII does not support extending from this interface as the used type `TerraformProvider | TerraformModuleProvider` – but I don't know why exactly. But that is something we can figure out.

jsteinich:
@sonodar would using the `addProvider` method suffice for your use case?

sonodar:
> would using the `addProvider` method suffice for your use case?

I think I can solve my own use case. I'll try.

PR #1791, opened 2022-05-12
test(lib): ensure list mapper does not fail when passed IResolvables

### LINKED ISSUE #835: Property Mapper should Respect Tokens

### LINKED ISSUE #910: bug(lib): Fix Token Resolve

### LINKED ISSUE #1788: [Research] Token Resolution Issue
Placeholder to do some research into #910 and the original source issue, #835. 

Verify (a) that it's still an active issue, and (b) whether what Sebastian identified in #910 does seem to be the cause.

We'll do the research in 0.11 so that any implementation work can happen in 0.12.

DanielMSchmidt:
Closes #1788
Closes #910 
Closes #835

### DIFF
--- a/packages/cdktf/lib/util.ts
+++ b/packages/cdktf/lib/util.ts
@@ -7,9 +7,7 @@ import { Tokenization } from "./tokens/token";
 export function deepMerge(target: any, ...sources: any[]) {
   if (Tokenization.isResolvable(target) && sources.length > 0) {
     throw new Error(
-      `Invalid usage. Target (${JSON.stringify(
-        target
-      )}) can not be a resolvable token when overrides are specified. Please replace the value of the field you are overriding with a static value.`
+      `Invalid usage. Target (${target.toString()}) can not be a resolvable token when overrides are specified. Please replace the value of the field you are overriding with a static value.`
     );
   }
 
--- a/packages/cdktf/lib/util.ts
+++ b/packages/cdktf/lib/util.ts
@@ -5,6 +5,14 @@ import { Tokenization } from "./tokens/token";
  */
 
 export function deepMerge(target: any, ...sources: any[]) {
+  if (Tokenization.isResolvable(target) && sources.length > 0) {
+    throw new Error(
+      `Invalid usage. Target (${JSON.stringify(
+        target
+      )}) can not be a resolvable token when overrides are specified. Please replace the value of the field you are overriding with a static value.`
+    );
+  }
+
   for (const source of sources) {
     if (typeof source !== "object" || typeof target !== "object") {
       throw new Error(
--- a/packages/cdktf/test/runtime.test.ts
+++ b/packages/cdktf/test/runtime.test.ts
@@ -2,6 +2,7 @@ import { resolve } from "../lib/_tokens";
 import { listMapper, hashMapper, anyToTerraform } from "../lib/runtime";
 import { Token } from "../lib/tokens/token";
 import { Intrinsic } from "../lib/tokens/private/intrinsic";
+import { deepMerge } from "../lib/util";
 
 const resolveExpression = (expr: any) => resolve(null as any, expr);
 describe("Runtime", () => {
@@ -140,6 +141,25 @@ describe("Runtime", () => {
           ],
         }
       `);
+    });
+
+    it("throws an error when a value a token is overwriten", () => {
+      const reference = new Intrinsic(
+        "${some_resource.my_resource.some_attribute_array}"
+      );
+
+      expect(() =>
+        resolveExpression(
+          deepMerge(
+            {
+              x: reference,
+            },
+            { x: { foo: "bar" } }
+          )
+        )
+      ).toThrowErrorMatchingInlineSnapshot(
+        `"Invalid usage. Target (\\"<unresolved-token>\\") can not be a resolvable token when overrides are specified. Please replace the value of the field you are overriding with a static value."`
+      );
     });
   });
 });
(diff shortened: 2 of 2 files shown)


ansgarm (approved):
[.](https://www.youtube.com/watch?v=UBX8MWYel3s)

jsteinich:
I believe #835 has been fixed, but I'm not yet convinced that #910 still doesn't cause issues.

I haven't quite thought of a case where this would actually happen, but if `deepMerge` ends up mixing tokens and non-tokens within the same property the results could be wrong.
Another spot worth testing is `Lazy`. We don't really use it, but it does exist.

DanielMSchmidt:
>  if deepMerge ends up mixing tokens and non-tokens within the same property the results could be wrong.

on the same property as in different types of tokens in a list or as function invocations?

jsteinich:
> > if deepMerge ends up mixing tokens and non-tokens within the same property the results could be wrong.
> 
> on the same property as in different types of tokens in a list or as function invocations?

A couple possible cases:
* Trying to merge a token and non-token. One direct just overwrites, the other might append to the token object.
* Adding token and non-token to list. Resolution will fail later.

@ansgarm there was a slack thread recently about JSON.stringify; would that line up with this.

packages/cdktf/lib/util.ts
```diff
@@ -5,6 +5,14 @@ import { Tokenization } from "./tokens/token";
  */
 
 export function deepMerge(target: any, ...sources: any[]) {
+  if (Tokenization.isResolvable(target) && sources.length > 0) {
+    throw new Error(
+      `Invalid usage. Target (${JSON.stringify(
```
jsteinich, line 10: I wonder if just using `target.toString` would be more readable.

PR #1794, opened 2022-05-12
feat(hcl2json): expose hcl expression parser

### LINKED ISSUE #842: Convert: Expressions are not fully understood when converting
### Community Note

### Description

Convert needs to understand Terraform expressions to extract references from it and to create an equivalent template string. This is currently done by splitting the reference string into it's parts and searching for known ids in the existing parts. This behaviour mimics a lexer, but due to the oversimplified implementation fails in certain edge cases. 

This is a shortcut over using a similar approach to the `@cdktf/hcl2json` package and [using the Go implementation for expression parsing](https://github.com/hashicorp/hcl/blob/0827c2f3e6483d3a7680d41b4399f4a374747c14/json/public.go#L76) through WASM.

An alternative solution is using a proper parser like [ANTLR](https://www.antlr.org/). To use such tools a formal grammar for Terraform Expressions needs to be written.

This step would also enable us to determine the result type of an expression, that would remove another limitation of the current implementation, the missing type casts when the terraform expressions result in a different type than string.

DanielMSchmidt:
Closes #842 by using the official expression parser that terraform itself uses through WASM. This way we can find all references and only things that are actually references. This should reduce random errors from hcl2cdk and make the convert experience more predictable.


packages/@cdktf/hcl2cdk/lib/expressions.ts
```diff
@@ -12,101 +13,54 @@ export type Reference = {
...
-
-  delimiters.forEach((delimiter) => {
-    possibleVariableSpots = possibleVariableSpots.reduce(
-      (carry, str) => [...carry, ...str.split(delimiter)],
-      [] as string[]
-    );
-  });
+  logger.debug(
+    `found possible variable spots: ${JSON.stringify(possibleVariableSpots)}`
+  );
```
ansgarm, line 31: I love how much code is deleted here 😊

packages/@cdktf/hcl2cdk/lib/generation.ts
```diff
@@ -267,28 +289,28 @@ export function resource(
...
+  const expressions = await asExpression(
+    scope,
+    resource,
+    key,
+    mappedConfig,
+    nodeIds,
+    false,
+    false,
+    getReference(graph, id) || overrideReference
+  );
```
ansgarm, line 301: This does not clone the Array anymore, is that okay?

packages/@cdktf/hcl2cdk/lib/generation.ts
```diff
@@ -327,27 +354,29 @@ you need to keep this like it is.`;
...
+              content,
+            },
+            "path-for-dynamic-blocks-can-be-ignored",
+            nodeIds,
+            [scopedVar]
+          ),
+          loopComment
+        );
+      })
+    )) as any), // TODO: get rid of this any through type assertions
```
ansgarm, line 375: todo left?

packages/@cdktf/hcl2cdk/test/expressions.test.ts
```diff
@@ -23,13 +23,13 @@ describe("expressions", () => {
     it("finds no references in literals", () => {
       expect(
         extractReferencesFromExpression("nothingtobeseen", nodeIds)
-      ).toEqual([]);
+      ).resolves.toEqual([]);
```
ansgarm, line 26: Does jest track these kinds of checks? Or would this need an async test and an `await expect(...`?

packages/@cdktf/hcl2cdk/test/expressions.test.ts
```diff
@@ -419,27 +421,22 @@ describe("expressions", () => {
     });
   });
 
-  describe("#referencesToAst", () => {
-    it("nested terraform expressions without space", () => {
+  // TODO: check if this example is correct
+  describe.skip("#referencesToAst", () => {
```
ansgarm, line 425: todo left

packages/@cdktf/hcl2cdk/test/hcl2cdk.test.ts
```diff
@@ -31,9 +31,12 @@ const cdktfBin = path.join(__dirname, "../../../cdktf-cli/bin/cdktf");
 const cdktfDist = path.join(__dirname, "../../../../dist");
 
 const createTestCase =
-  (describeFn: typeof describe) =>
+  (opts: { skip?: true }) =>
   (name: string, hcl: string, shouldSynth: Synth) => {
-    describeFn(name, () => {
+    if (opts.skip) {
+      return;
```
ansgarm, line 37: We could use this:
```suggestion
      describe.skip(name, () => {});
      return;
```
to let jest know about all skipped tests (it would print them in that case). No hard feelings though.

packages/@cdktf/hcl2json/lib/index.ts
```diff
@@ -118,3 +119,233 @@ export async function convertFiles(
...
+  if ("Traversal" in entry) {
+    return [traversalToReference(input, entry)];
+  }
+
+  // Multiple terraform parts in an expression
+  if ("Parts" in entry) {
+    return entry.Parts.reduce(
+      (carry, part) => [...carry, ...findAllReferencesInAst(input, part)],
+      [] as Reference[]
+    );
```
ansgarm, line 276: Could be an `Array.flatMap` call which would make it more readable imho

ansgarm (commented):
Nice work!

packages/@cdktf/hcl2cdk/lib/expressions.ts
```diff
@@ -412,35 +369,40 @@ export const extractDynamicBlocks = (
...
-  return Object.values(item as Record<string, any>).reduce(
-    (carry, i) => [...carry, ...findUsedReferences(nodeIds, i)],
-    []
-  );
+
+  return (
+    await Promise.all(
+      Object.values(item as Record<string, any>).map(
+        (i) => findUsedReferences(nodeIds, i),
+        []
```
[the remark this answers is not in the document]
DanielMSchmidt: it used to be a reduce :D

packages/@cdktf/hcl2cdk/test/expressions.test.ts
DanielMSchmidt: Yes it does, you can either await directly or have an expect on a promise with the resolves property, both work. but with await you have to return the expect (I think)

packages/@cdktf/hcl2cdk/lib/expressions.ts
DanielMSchmidt: yeah, it was the most fragile code as well :)

packages/@cdktf/hcl2cdk/lib/generation.ts
DanielMSchmidt: Yeah, not sure why that even was a spread, I think it was left over from a time where we had more than one thing in this list by default ( I think I used to have ternary spreads in there `...(thisOverwrite ? ... : [])`

PR #1802, opened 2022-05-12
feat(lib): add a warning about id fields

DanielMSchmidt:


packages/@cdktf/provider-generator/lib/get/generator/emitter/struct-emitter.ts
```diff
@@ -38,12 +54,14 @@ export class StructEmitter {
         this.code.line(
           `* Docs at Terraform Registry: {@link ${resource.linkToDocs}#${att.terraformName} ${resource.className}#${att.terraformName}}`
         );
+        this.warnAboutIdField(att);
```
ansgarm, line 57: Is there a way we can figure out which provider sdk version has been used? Probably not, right?

DanielMSchmidt: Unfortunately this information is not part of the schema, I think one would need to monkey type it, e.g. if there is a nested type present somewhere it's a newer one. Not sure if it's worth the effort though

ansgarm: Ah, yeah, not worth it then 😇

packages/@cdktf/provider-generator/lib/get/generator/emitter/struct-emitter.ts
```diff
@@ -19,6 +20,21 @@ export class StructEmitter {
...
   }
 
+  // Due to https://github.com/hashicorp/terraform-plugin-sdk/commit/2387eb85e32c064b4a62718c9f5c80bf00dc7fb9 all
+  // resources from providers using the old SDK have the id field by default
+  // We have no way to distinguish them through the provider schema, so a word of warning for our users
+  private warnAboutIdField(att: AttributeModel) {
+    if (att.name === "id") {
+      this.code.line(`*`);
+      this.code.line(
+        `* Please be aware that the id field is automatically added to all resources in providers using a SDK version below 2.`
```
ansgarm, line 30:
```suggestion
        `* Please be aware that the id field is automatically added to all resources in Terraform providers using a Terraform provider SDK version below 2.`
```

ansgarm: Make it a bit more clear, so users don't wonder which SDK version exactly (if they are not familiar with what happens under the hood)

ansgarm (approved):

Issue #1268, opened 2021-11-08
API docs: Add docs to Terraform backends


ansgarm:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

There are currently no docstrings on the constructs we have for Terraform backends.

See this one for example: https://github.com/hashicorp/terraform-cdk/blob/main/packages/cdktf/lib/backends/s3-backend.ts

It should at least include a link to https://www.terraform.io/docs/language/settings/backends/s3.html

aayushharwani-aidash:
@ansgarm I am interested in working on this. Can you plz guide ?

ansgarm:
Hi @aayushharwani-aidash 👋 Wow! That's great!

I'll use the S3 backend as an example, but feel free to pick your favorite(s) – you also don't have to do all at once to get a PR accepted and merged 🙂

Our backends can be found [here](https://github.com/hashicorp/terraform-cdk/tree/main/packages/cdktf/lib/backends) and the docs on them are [here](https://www.terraform.io/language/settings/backends).

So let's take the [`S3Backend`](https://github.com/hashicorp/terraform-cdk/blob/main/packages/cdktf/lib/backends/s3-backend.ts) and the [related docs](https://www.terraform.io/language/settings/backends/s3) as an example.

The backend has a `S3BackendProps` type that defines the arguments that you can set on the backend:
https://github.com/hashicorp/terraform-cdk/blob/537c7da0ffffff0d7b31b840f31fb3c6c57ee531/packages/cdktf/lib/backends/s3-backend.ts#L40-L44

The [docs contain descriptions](https://www.terraform.io/language/settings/backends/s3#configuration) for those config keys. For example for the `region` property:
> [region](https://www.terraform.io/language/settings/backends/s3#region) - (Required) AWS Region of the S3 Bucket and DynamoDB Table (if used). This can also be sourced from the AWS_DEFAULT_REGION and AWS_REGION environment variables.
(...)

A docstring for this property could look like this:
```ts
export interface S3BackendProps {
  // ...
  
  /**
[...]

PR #1803, opened 2022-05-14
feat(lib): add docstrings to gcs/http/local backend resources

aayushharwani-aidash:
Have added doc-string to gcs,http,local backend resources.

Working on issue - https://github.com/hashicorp/terraform-cdk/issues/1268

### DIFF
--- a/packages/cdktf/lib/backends/local-backend.ts
+++ b/packages/cdktf/lib/backends/local-backend.ts
@@ -49,13 +49,22 @@ export class DataTerraformRemoteStateLocal extends TerraformRemoteState {
     super(scope, id, "local", config);
   }
 }
-
+/**
+ * The local backend stores state on the local filesystem, 
+ * locks that state using system APIs, and performs operations locally.
+ * 
+ * Read more about this backend in the Terraform docs:
+ * https://www.terraform.io/language/settings/backends/http
+ */
 export interface LocalBackendProps {
   /**
-   * Path where the state file is stored.
-   * @default - defaults to terraform.${stackId}.tfstate
+   * (Optional) The path to the tfstate file. 
+   * This defaults to "terraform.tfstate" relative to the root module by default.
    */
   readonly path?: string;
+  /**
+   * (Optional) The path to non-default workspaces.
+   */
   readonly workspaceDir?: string;
 }
 
--- a/packages/cdktf/lib/backends/local-backend.ts
+++ b/packages/cdktf/lib/backends/local-backend.ts
@@ -60,9 +60,6 @@ export interface LocalBackendProps {
   /**
    * Path where the state file is stored.
    * @default - defaults to terraform.${stackId}.tfstate
-   * 
-   * (Optional) The path to the tfstate file. 
-   * This defaults to "terraform.tfstate" relative to the root module by default.
    */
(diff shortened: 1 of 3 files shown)


aayushharwani-aidash:
@ansgarm

packages/cdktf/lib/backends/local-backend.ts
```diff
@@ -49,13 +49,22 @@ export class DataTerraformRemoteStateLocal extends TerraformRemoteState {
     super(scope, id, "local", config);
   }
 }
-
+/**
+ * The local backend stores state on the local filesystem, 
+ * locks that state using system APIs, and performs operations locally.
+ * 
+ * Read more about this backend in the Terraform docs:
+ * https://www.terraform.io/language/settings/backends/http
```
ansgarm, line 57: Seems to be the wrong link 😅

packages/cdktf/lib/backends/local-backend.ts
```diff
@@ -49,13 +49,22 @@ export class DataTerraformRemoteStateLocal extends TerraformRemoteState {
...
+ */
 export interface LocalBackendProps {
   /**
-   * Path where the state file is stored.
-   * @default - defaults to terraform.${stackId}.tfstate
+   * (Optional) The path to the tfstate file. 
+   * This defaults to "terraform.tfstate" relative to the root module by default.
```
ansgarm, line 62: While this is true for Terraform we have a different behaviour in place in CDKTF because we need to support multiple stacks. So this comment should be left as it was.

ansgarm (changes requested):
Hi @aayushharwani-aidash 👋 Thanks again for your work on this 👍
I found two small things to address before we can merge this.

packages/cdktf/lib/backends/local-backend.ts
aayushharwani-aidash: have made changes, as asked by u.

packages/cdktf/lib/backends/local-backend.ts
aayushharwani-aidash: have fixed that

packages/cdktf/lib/backends/local-backend.ts
ansgarm: Ah, sorry, I should've been a bit more clear: The previous comment was enough already – having both now with contradictory explanations is not correct either 🙈

packages/cdktf/lib/backends/local-backend.ts
```diff
@@ -49,13 +49,25 @@ export class DataTerraformRemoteStateLocal extends TerraformRemoteState {
...
+ * Read more about this backend in the Terraform docs:
+ * https://www.terraform.io/language/settings/backends/local
+ */
 export interface LocalBackendProps {
   /**
    * Path where the state file is stored.
    * @default - defaults to terraform.${stackId}.tfstate
+   * 
+   * (Optional) The path to the tfstate file. 
+   * This defaults to "terraform.tfstate" relative to the root module by default.
```
ansgarm, line 65:
```suggestion
```

ansgarm (approved):
Thanks a lot!

PR #1809, opened 2022-05-20
chore(cli): add sentry error reporting

DanielMSchmidt:
Adding crash reporting will help us improve the product faster 🎉

### DIFF
--- a/packages/cdktf-cli/bin/cdktf.ts
+++ b/packages/cdktf-cli/bin/cdktf.ts
@@ -1,7 +1,6 @@
 process.env.NODE_ENV = process.env.NODE_ENV || "production";
 
 import * as yargs from "yargs";
-import * as semver from "semver";
 import * as path from "path";
 import * as os from "os";
 import * as fs from "fs-extra";
@@ -9,6 +8,7 @@ import { readCDKTFManifest } from "../lib/util";
 import { IsErrorType } from "../lib/errors";
 import { collectDebugInformation } from "../lib/debug";
 import { CDKTF_DISABLE_PLUGIN_CACHE_ENV } from "../lib/environment";
+import * as Sentry from "@sentry/node";
 
 const ensurePluginCache = (): string => {
   const pluginCachePath =
@@ -22,11 +22,6 @@ const ensurePluginCache = (): string => {
 
 if (!CDKTF_DISABLE_PLUGIN_CACHE_ENV) {
   process.env.TF_PLUGIN_CACHE_DIR = ensurePluginCache();
-}
-
-if (semver.lt(process.version, "10.12.0")) {
-  console.error("Need at least Node v10.12 to run");
-  process.exit(1);
 }
 
 const customCompletion = function (
@@ -141,11 +136,13 @@ yargs
       console.error(error.stack);
       console.error("Collecting Debug Information...");
       const debugOutput = await collectDebugInformation();
+
       console.error("Debug Information:");
       Object.entries(debugOutput).forEach(([key, value]) => {
         console.log(`${key}: ${value === null ? "null" : value}`);
       });
     }
 
+    await Sentry.close(4000);
     process.exit(1);
   }).argv;
--- a/packages/cdktf-cli/bin/cmds/handlers.ts
+++ b/packages/cdktf-cli/bin/cmds/handlers.ts
@@ -41,6 +41,7 @@ import {
   verifySimilarLibraryVersion,
 } from "./helper/check-environment";
 import { collectDebugInformation } from "../../lib/debug";
+import { initializErrorReporting } from "../../lib/error-reporting";
 
 const chalkColour = new chalk.Instance();
 const config = cfg.readConfigSync();
@@ -59,6 +60,7 @@ async function getProviderRequirements(provider: string[]) {
 }
 
 export async function convert({ language, provider }: any) {
+  await initializErrorReporting();
   await displayVersionMessage();
 
   const providerRequirements = await getProviderRequirements(provider);
@@ -89,6 +91,7 @@ export async function convert({ language, provider }: any) {
 }
 
 export async function deploy(argv: any) {
+  await initializErrorReporting(true);
   throwIfNotProjectDirectory();
   await displayVersionMessage();
   await checkEnvironment();
@@ -124,6 +127,7 @@ export async function deploy(argv: any) {
 }
 
 export async function destroy(argv: any) {
+  await initializErrorReporting(true);
   throwIfNotProjectDirectory();
   await displayVersionMessage();
   await checkEnvironment();
@@ -146,6 +150,7 @@ export async function destroy(argv: any) {
 }
 
 export async function diff(argv: any) {
+  await initializErrorReporting(true);
   throwIfNotProjectDirectory();
   await displayVersionMessage();
   await checkEnvironment();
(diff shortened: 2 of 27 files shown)


packages/cdktf-cli/bin/cdktf.ts
```diff
@@ -24,11 +24,6 @@ if (!CDKTF_DISABLE_PLUGIN_CACHE_ENV) {
   process.env.TF_PLUGIN_CACHE_DIR = ensurePluginCache();
 }
 
-if (semver.lt(process.version, "10.12.0")) {
-  console.error("Need at least Node v10.12 to run");
-  process.exit(1);
-}
```
jsteinich, line 30: Is this being removed since inaccurate and doesn't really work? If so, would be nice to get it working some other way.

packages/cdktf-cli/bin/cmds/handlers.ts
```diff
@@ -163,8 +169,7 @@ export async function diff(argv: any) {
 }
 
 export async function get(argv: any) {
-  throwIfNotProjectDirectory();
-  await displayVersionMessage();
```
jsteinich, line 167: Why remove these lines?

packages/cdktf-cli/bin/cdktf.ts
DanielMSchmidt: We have the check environment function that takes care of it centrally with the right condition

packages/cdktf-cli/bin/cmds/handlers.ts
DanielMSchmidt: By accident 🙈

packages/cdktf-cli/bin/cdktf.ts
DanielMSchmidt: It's here: https://github.com/hashicorp/terraform-cdk/blob/fd88df429238ec85bf5d7550e58c19146f10cc65/packages/cdktf-cli/bin/cmds/helper/check-environment.ts#L46

website/docs/cdktf/create-and-deploy/configuration-file.mdx
```diff
@@ -33,6 +33,7 @@ export interface Config {
   readonly output: string; // Default: 'cdktf.out'. Where the synthesized JSON should go. Also will be the working directory for Terraform operations
   readonly codeMakerOutput: string; // Default: '.gen'. Path where generated provider bindings will be rendered to.
   readonly projectId: string; // Default: generated UUID. Unique identifier for the project used to differentiate projects
+  readonly sendCrashReports: boolean; // Default: false. Whether to send crash reports to the CDK team
```
ansgarm, line 36:
```suggestion
  readonly sendCrashReports: boolean; // Default: false. Whether to send crash reports to the CDKTF team
```

website/docs/cdktf/create-and-deploy/configuration-file.mdx
```diff
@@ -206,3 +207,15 @@ This generates the `aws` provider bindings in the folder `./imports`. This is us
...
+
+### Enable Crash Reporting for the CLI
+
+You can enable or disable crash reporting by setting the `sendCrashReports` property to `true` or `false`. Sending crash reports helps our team improve the CLI faster, you can learn more about what we track on our [telemetry page](/cdktf/telemetry#crash-reporting).
+
+```json
+{
+  "language": "typescript",
+  "app": "npm run --silent compile && node main.js",
+  "sendCrashReports": "true"
```
ansgarm, line 219: This example might be a bit misleading, as the property is parsed as a JSON and treated as a boolean.
And if I would write `"sendCrashReports": "false"` it would still enable them.
We should either catch `"false"` (preferred) or change this example to `"sendCrashReports": true`.

website/docs/cdktf/telemetry.mdx
```diff
@@ -14,3 +14,13 @@ Starting with CDK for Terraform 0.6, this information includes a random UUID tha
 CDK for Terraform fingerprints the type of CI used and includes that in telemetry instead of the UUID when CDK for Terraform runs in a Continuous Integration tool, such as GitHub Actions, Jenkins, or GitLab. The only information submitted is the type of CI system, and no unique information about accounts, paths, workspaces, environment variables, or other potentially private information is shared. This data helps the team plan where to focus future efforts, because running infrastructure as code tools in CI is a key workflow we seek to improve.
 
 The information that is sent to Checkpoint is anonymous and cannot be used to identify the user or host. The use of Checkpoint is completely optional and it can be disabled at any time by setting the `CHECKPOINT_DISABLE` environment variable to a non-empty value.
+
+## Crash Reporting
+
+To improve the stability of the [CDK for Terraform CLI](/cdktf/cli-reference/cli-configuration) we also [Sentry](https://sentry.io/) for error reporting.
```
ansgarm, line 20:
```suggestion
To improve the stability of the [CDK for Terraform CLI](/cdktf/cli-reference/cli-configuration) we also use [Sentry](https://sentry.io/) for error reporting.
```

website/docs/cdktf/telemetry.mdx
```diff
@@ -14,3 +14,13 @@ Starting with CDK for Terraform 0.6, this information includes a random UUID tha
...
+
+The error reports send include a stack trace and log events that happened before the error. This information only gets send if there is an error. If no error happens a session is tracked, so that we can calculate the percentage of crash free sessions as a metric.
```
ansgarm, line 22:
```suggestion
The error reports sent include a stack trace and log events that happened before the error. This information only gets sent if there is an error. If no error happens a session is tracked, so that we can calculate the percentage of crash free sessions as a metric.
```

website/docs/cdktf/telemetry.mdx
```diff
@@ -14,3 +14,13 @@ Starting with CDK for Terraform 0.6, this information includes a random UUID tha
...
+
+The error reporting is scoped to the CLI and no environment variables are tracked, limiting the risk of sending secret information by accident.
+
+If you don't want your project to track errors, change the [`sendCrashReports`](/cdktf/create-and-deploy/configuration-file#enable-crash-reporting-for-the-cli) property in your `cdktf.json` file to false.
```
ansgarm, line 26: This sounds like the default would be on. I think we should mention that we ask first, but if a user wants to disable them later one (after enabling them) they can set it to false.

website/docs/cdktf/telemetry.mdx
```diff
@@ -14,3 +14,14 @@ Starting with CDK for Terraform 0.6, this information includes a random UUID tha
...
+
+## Crash Reporting
+
+To improve the stability of the [CDK for Terraform CLI](/cdktf/cli-reference/cli-configuration) we also use [Sentry](https://sentry.io/) for error reporting.
+
+The error reports sent include a stack trace and log events that happened before the error. This information only gets sent if there is an error. If no error happens a session is tracked, so that we can calculate the percentage of crash free sessions as a metric.
+
+The error reporting is scoped to the CLI and no environment variables are tracked, limiting the risk of sending secret information by accident.
+
+We won't track any errors as long as the [`sendCrashReports`](/cdktf/create-and-deploy/configuration-file#enable-crash-reporting-for-the-cli) property in the `cdktf.json` file in not set or set to `false`. If it is not set we will ask you if you want to enable crash reporting and persist the answer in your `cdktf.json` file.
```
ansgarm, line 26:
```suggestion
We won't track any errors as long as the [`sendCrashReports`](/cdktf/create-and-deploy/configuration-file#enable-crash-reporting-for-the-cli) property in the `cdktf.json` file is not set or is set to `false`. If it is not set we will ask you if you want to enable crash reporting and persist the answer in your `cdktf.json` file.
```

ansgarm (approved):
only a typo left 🚀

PR #1825, opened 2022-05-25
feat(cli): move stacks in new files for fresh projects

DanielMSchmidt:


packages/cdktf-cli/templates/java/src/main/java/com/mycompany/app/MainStack.java
```diff
@@ -0,0 +1,14 @@
+package com.mycompany.app;
+
+import software.constructs.Construct;
+
+import com.hashicorp.cdktf.TerraformStack;
+
+public class Main extends TerraformStack
```
ansgarm, line 7: Should we name this "MyStack" to be similar to how our TypeScript template looks?

jsteinich: Either `MyStack` or `MainStack` would be better than `Main`. Also need to make sure the class name matches the file name.

packages/cdktf-cli/templates/java/src/main/java/com/mycompany/app/Main.java
```diff
@@ -5,17 +5,13 @@
...
 import com.hashicorp.cdktf.TerraformStack;
 
-public class Main extends TerraformStack
-{
-    public Main(final Construct scope, final String id) {
-        super(scope, id);
-
-        // define resources here
-    }
+import com.mycompany.app.MainStack;
```
jsteinich, line 8: This should be in the `com.mycompany.app` package and then this import won't be necessary.

packages/cdktf-cli/templates/csharp/MainStack.cs
```diff
@@ -0,0 +1,15 @@
+using System;
+using Constructs;
+using HashiCorp.Cdktf;
+
+
+namespace MyCompany.MyApp
+{
+    class MyApp : TerraformStack
```
jsteinich, line 8: Should use the same class name for the stack as decided with the Java version. Make sure the filename matches up as well (not a hard requirement in C#, but good practice).

packages/cdktf-cli/templates/csharp/Main.cs
```diff
@@ -5,13 +5,8 @@
 
 namespace MyCompany.MyApp
 {
-    class MyApp : TerraformStack
+    class MyApp
```
jsteinich, line 8: `Program` is the standard name for an entry point class.

jsteinich (changes requested):

packages/cdktf-cli/templates/csharp/Main.cs
```diff
@@ -5,13 +5,8 @@
 
 namespace MyCompany.MyApp
 {
-    class MyApp : TerraformStack
+    class MyApp : Program
```
jsteinich, line 8: Should probably rename the file to `Program.cs`. Will need to update hooks and some tests to account for that change.

jsteinich (changes requested):

ansgarm (approved):
Nice!
We need to update the Learn tutorials for C# and Java after we released this and check our docs whether they mention the older way.
Let's add some kind of "post 0.12 release" issue to keep track of that.

packages/cdktf-cli/templates/csharp/Program.cs
```diff
@@ -5,13 +5,8 @@
 
 namespace MyCompany.MyApp
 {
-    class MyApp : TerraformStack
+    class MyApp : Program
```
jsteinich, line 8: The class name should be `Program` and it won't extend/implement anything.

packages/cdktf-cli/templates/csharp/MainStack.cs
```diff
@@ -0,0 +1,15 @@
+using System;
+using Constructs;
+using HashiCorp.Cdktf;
+
+
+namespace MyCompany.MyApp
+{
+    class MainStack : TerraformStack
```
jsteinich, line 8: Would be good to actually use this in the template.

packages/cdktf-cli/templates/csharp/Program.cs
jsteinich: Also creating an instance of this below will fail since that constructor doesn't exist.

PR #1830, opened 2022-05-30
iterators

ansgarm:
- feat(lib): Introduce iterator and for_each to implement apply time iteration
- tests(lib): add tests for iterator.ts
- chore(provider-generator): update snapshots


packages/cdktf/lib/iterator.ts
```diff
@@ -0,0 +1,198 @@
...
+   */
+  abstract _getForEachExpression(): any;
+
+  /**
+   * Creates a new iterator from a list
+   */
+  public static fromList(
+    list: string[] | number[] | ComplexList
+  ): ListIterator {
+    // TODO: this could return different iterators depending on the type of the list
```
DanielMSchmidt, line 26: TODO left

packages/cdktf/lib/iterator.ts
```diff
@@ -0,0 +1,198 @@
...
+   */
+  getBooleanMap(attribute: string): { [key: string]: boolean } {
+    return Token.asBooleanMap(propertyAccess(this._getValue(), [attribute]));
+  }
+
+  /**
+   * @internal
+   */
+  protected _getValue(): any {
+    return Lazy.anyValue(
```
DanielMSchmidt, line 113: Could you add a comment why lazy is needed, I think I remember why but I'm certain I'll be unsure in half a year

packages/cdktf/lib/terraform-data-source.ts
```diff
@@ -12,6 +12,8 @@ import { keysToSnakeCase, deepMerge } from "./util";
 import { ITerraformDependable } from "./terraform-dependable";
 import { ref, dependable } from "./tfExpression";
 import { IInterpolatingParent } from "./terraform-addressable";
+import { IIterator } from "./iterator";
+import assert = require("assert");
```
DanielMSchmidt, line 16: Why are we not using the normal import from syntax here?

packages/cdktf/lib/terraform-data-source.ts
```diff
@@ -87,11 +91,16 @@ export class TerraformDataSource
   }
 
   public get terraformMetaArguments(): { [name: string]: any } {
+    assert(
+      !this.forEach || typeof this.count === "undefined",
+      `forEach and count are mutually exclusive. You can only use either of them. Check data source at path: ${this.node.path}`
```
DanielMSchmidt, line 96:
```suggestion
      `forEach and count are both set, but they are mutually exclusive. You can only use either of them. Check the data source at path: ${this.node.path}`
```

DanielMSchmidt: But good idea to assert that 👍

packages/cdktf/lib/terraform-resource.ts
```diff
@@ -119,11 +125,16 @@ export class TerraformResource
   }
 
   public get terraformMetaArguments(): { [name: string]: any } {
+    assert(
+      !this.forEach || typeof this.count === "undefined",
+      `forEach and count are mutually exclusive. You can only use either of them. Check resource at path: ${this.node.path}`
```
DanielMSchmidt, line 130:
```suggestion
      `forEach and count are both set, but they are mutually exclusive. You can only use either of them. Check the resource at path: ${this.node.path}`
```

packages/cdktf/test/iterator.test.ts
```diff
@@ -0,0 +1,241 @@
...
+  const it = Iterator.fromList(["a", "b", "c"]);
+
+  new TestResource(stack, "test", {
+    forEach: it,
+    name: it.value,
+  });
+
+  const synth = JSON.parse(Testing.synth(stack));
+
+  expect(synth).toHaveProperty(
```
DanielMSchmidt, line 24: Nice, didn't know about this supporting paths, how nice!

DanielMSchmidt (approved):
Great work 👍

packages/cdktf/lib/iterator.ts
ansgarm: That one was on purpose. Should I just keep the comment without the todo prefix?

packages/cdktf/lib/terraform-data-source.ts
ansgarm: The normal import failed `yarn watch`, this one didn't 😅 (We're also already using this weird syntax for other third party deps in cdktf)

packages/cdktf/lib/iterator.ts
```diff
@@ -0,0 +1,200 @@
...
+  /**
+   * @internal used by TerraformResource to set the for_each expression
+   */
+  abstract _getForEachExpression(): any;
+
+  /**
+   * Creates a new iterator from a list
+   */
+  public static fromList(
+    list: string[] | number[] | ComplexList
```
jsteinich, line 24: This doesn't cover all types of lists.

packages/cdktf/lib/iterator.ts
```diff
@@ -0,0 +1,200 @@
...
+      | { [key: string]: boolean }
+  ): MapIterator {
+    return new MapIterator(map);
+  }
+
+  /**
+   * @param attribute name of the property to retrieve
+   * @returns the given attribute of the current item iterated over as a string
+   */
+  getString(attribute: string): string {
```
jsteinich, line 49: Not all `get<Type>` methods are present that we support. One example is `getBoolean`

packages/cdktf/lib/iterator.ts
```diff
@@ -0,0 +1,200 @@
...
+  getBooleanMap(attribute: string): { [key: string]: boolean } {
+    return Token.asBooleanMap(propertyAccess(this._getValue(), [attribute]));
+  }
+
+  /**
+   * @internal
+   */
+  protected _getValue(): any {
+    // uses a Lazy value to be able to access the current TerraformStack and pass it to ref()
+    return Lazy.anyValue(
```
jsteinich, line 114: We haven't done a lot of testing with `Lazy`. One area that seems likely to have issues is with cross stack references. Should make sure to have some test cases with iterators across stacks.

jsteinich (commented):
I think we need some follow up tasks (maybe they already exist) for:
* Generating iterators for complex types
* Iterators that match type for simple types
* Similar support for `count.index`
* Providing a way to reference elements created by iteration as list/map

packages/cdktf/lib/iterator.ts
ansgarm: Can you help me finding out which ones are missing?

packages/cdktf/lib/iterator.ts
ansgarm: Ah, right

packages/cdktf/lib/iterator.ts
ansgarm: Yes, that sounds like a good test case to add

packages/cdktf/lib/iterator.ts
jsteinich: A list of booleans is represented as `Array<boolean | IResolvable> | IResolvable`.
There's also `MapList`.

I think maps have some missing ones as well. For example, `String/Number/Boolean/Any Map`.

ansgarm: added ✅ Thank you for helping!

packages/cdktf/lib/iterator.ts
ansgarm: added `getBoolean, getAny, getAnyMap`

packages/cdktf/lib/iterator.ts
ansgarm: You were right, cross stack references did not resolve properly with iterators (due to Lazy). I'm going to have a time-boxed look at it.

ansgarm: ah wait, it does work 🎉 The test needed to call `stack.prepareStack()` for the cross stack references to be created.

ansgarm: added a test ✅

DanielMSchmidt (approved):

a0s:
Great!! When it will be released?

Issue #1816, opened 2022-05-24
Re-enable windows test for provider add command for csharp


ansgarm:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Somehow the provider add command integration test [did fail](https://github.com/hashicorp/terraform-cdk/runs/6572021428?check_suite_focus=true) on windows

<details>
  <summary>test output</summary>
Note: failing tests are currently retried once (that's why there's two times the output, however, it is still weird that they fail differently)

```
FAIL csharp/provider-add-command/test.ts (150.755 s)
  provider add command
    pre-built
      × installs pre-built provider using nuget (56093 ms)

  ● provider add command › pre-built › installs pre-built provider using nuget

    expect(received).toContain(expected) // indexOf

    Expected substring: "<PackageReference Include=\"HashiCorp.Cdktf.Providers.Random\" Version=\"0.2.55\" />"
    Received string:    "﻿<Project Sdk=\"Microsoft.NET.Sdk\">·
      <PropertyGroup>
        <OutputType>Exe</OutputType>
        <TargetFramework>netcoreapp3.1</TargetFramework>
      </PropertyGroup>·
      <ItemGroup>
        <PackageReference Include=\"HashiCorp.Cdktf\" Version=\"0.10.4\" />
      </ItemGroup>·
      <ItemGroup>·
        <ProjectReference Include=\".gen\\Providers.Null\\Providers.Null.csproj\" />·
      </ItemGroup>·
    </Project>"

      37 |       const proj = driver.readLocalFile("MyTerraformStack.csproj");
      38 |
    > 39 |       expect(proj).toContain(
         |                    ^
      40 |         '<PackageReference Include="HashiCorp.Cdktf.Providers.Random" Version="0.2.55" />'
      41 |       );
      42 |     }, 300_000);

      at csharp/provider-add-command/test.ts:39:20
      at fulfilled (csharp/provider-add-command/test.ts:5:58)

  ● provider add command › pre-built › installs pre-built provider using nuget

    expect(received).toMatchInlineSnapshot(snapshot)

[...]

PR #1831, opened 2022-05-30
fix(cli): Fix package version check on Windows

jsteinich:
For at least c#, the package version lookup is failing on Windows due to line endings. I went ahead and change the other languages as well.
I believe this is at least partially contributing to #1816.

### DIFF
--- a/packages/cdktf-cli/lib/debug.ts
+++ b/packages/cdktf-cli/lib/debug.ts
@@ -133,7 +133,7 @@ async function getPythonPackageVersion(
   }
 
   const versionInfo = output
-    .split("\n")
+    .split(/\r\n|\r|\n/)
     .find((line) => line.startsWith("Version:"));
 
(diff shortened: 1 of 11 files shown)


jsteinich:
I'd like to have a test for this, but I'm not sure the best spot. The unit tests aren't setup very well for checking this and I'd rather not make an entirely new integration test just for this.

ansgarm:
> I'd like to have a test for this, but I'm not sure the best spot. The unit tests aren't setup very well for checking this and I'd rather not make an entirely new integration test just for this.

Hm, I'd be open to both 🤷‍♂️ If it tests, it tests 😁

ansgarm (approved):

ansgarm:
I'm working on this, as our CI is currently broken 🙃 fyi: I'll commit to your branch, hope that is fine with you @jsteinich :)

PR #1840, opened 2022-06-02
feat(docs): Add iterator docs page

ansgarm:
To be merged after https://github.com/hashicorp/terraform-cdk/pull/1830 has been merged and is about to be released.

Is the position in the sidebar (below functions) a good place? Feel free to move to a better fitting position.

### DIFF
--- a/website/docs/cdktf/concepts/iterators.mdx
+++ b/website/docs/cdktf/concepts/iterators.mdx
@@ -1,20 +1,25 @@
 ---
 page_title: Iterators
 description: >-
-  Iterators allow you to iterate over a collection of values that are only available at runtime.
+  Iterators let you loop over a collection of values that are only available at runtime.
 ---
 
 # Iterators
 
-When you need to create multiple resources of the same type, you can usually use loops in your target language to do this.
-However there are certain situations where the number of resources to create is not known at synth time.
-In this case, you can use iterators.
 
-## Using iterators
+Iterators let you loop over a collection of values. You can use them to create multiple resources of the same type that are based on dynamic data that is only known at runtime.
 
-To create an iterator, import the `TerraformIterator` class and call the `.fromList()` or `.fromMap()` static method.
-Pass the iterator to the resource, data source, or module that you want to use it in using using `forEach` property.
-You can now use the iterator in attributes of that resource, data source, or module (see `name` below).
+## When to Use Iterators
+
+Use iterators when you need to reference dynamic data that is not known until after Terraform applies a configuration. For example, instance IDs that cloud providers assign on creation.
+
+When data is static or you know the values before [synthesizing your code](/cdktf/cli-reference/commands#synth), we recommend using loops in your preferred programming language.
+
+## Define Iterators
+
+Import the `TerraformIterator` class and call the `.fromList()` or `.fromMap()` static method. Then use the `forEach` property to pass the iterator to the resource, data source, or module that must use it. This lets you use the iterator in attributes.
+
+The example below uses an iterator to create a unique name for each new S3 bucket.
 
 ```ts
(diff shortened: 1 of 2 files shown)


website/docs/cdktf/concepts/iterators.mdx
```diff
@@ -0,0 +1,58 @@
+---
+page_title: Iterators
+description: >-
+  Iterators allow you to iterate over a collection of values that are only available at runtime.
```
laurapacilio, line 4:
```suggestion
  Iterators let you loop over a collection of values that are only available at runtime.
```

laurapacilio (changes requested):
Some thoughts for you!! great work on this - thank you!

website/docs/cdktf/concepts/iterators.mdx
```diff
@@ -0,0 +1,62 @@
+---
+page_title: Iterators
+description: >-
+  Iterators let you loop over a collection of values that are only available at runtime.
+---
+
+# Iterators
+
+Iterators let you loop over a collection of values. You can use them to create multiple resources of the same type that are based on dynamic data that is only known at runtime.
```
laurapacilio, line 9:
```suggestion
Iterators let you loop over a collection of values. You can use them to create multiple resources of the same type based on dynamic data that is only known at runtime.
```

website/docs/cdktf/concepts/iterators.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+## When to Use Iterators
+
+Use iterators when you need to reference dynamic data that is not known until after Terraform applies a configuration. For example, instance IDs that cloud providers assign on creation.
+
+When data is static or you know the values before [synthesizing your code](/cdktf/cli-reference/commands#synth), we recommend using loops in your preferred programming language.
+
+## Define Iterators
+
+Import the `TerraformIterator` class and call the `.fromList()` or `.fromMap()` static method. Then use the `forEach` property to pass the iterator to the resource, data source, or module that must use it. This lets you use the iterator in attributes.
```
laurapacilio, line 19:
```suggestion
Import the `TerraformIterator` class and call the `.fromList()` or `.fromMap()` static method. Then use the `forEach` property to pass the iterator to a resource, data source, or module. This lets you use the iterator in attributes.
```

website/docs/cdktf/concepts/iterators.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+const iterator = TerraformIterator.fromList(list.listValue);
+
+const s3Bucket = new s3.Bucket(this, "bucket", {
+  forEach: iterator,
+  name: iterator.value,
+});
+```
+
+You cannot access the index of items when iterating over lists. This is because Terraform requires sets for iteration and CDKTF implicitly converts lists to sets when iterating over them. This behavior prevents Terraform from accidentally deleting and recreating resources when their indices change. If you need an index, use `count` and an [escape hatch](/cdktf/concepts/providers-and-resources#escape-hatch) using the [`count.index` property](/language/meta-arguments/count).
```
laurapacilio, line 39:
```suggestion
You cannot access the index of items when iterating over lists. This is because CDKTF implicitly converts lists to sets when iterating over them, but Terraform requires sets for iteration. This behavior prevents Terraform from accidentally deleting and recreating resources when their indices change. If you need an index, use an [escape hatch](/cdktf/concepts/providers-and-resources#escape-hatch) with the [`count.index` property](/language/meta-arguments/count).
```

laurapacilio (approved):
Very nice!! :) A couple of tiny edits, but otherwise, I approve!

ansgarm:
Thank you @laurapacilio 💯

PR #1861, opened 2022-06-10
measure time taken in get

DanielMSchmidt:
- chore: measure get performance
- refactor(provider-generator): improve the speed of cdktf get


jsteinich:
This may end up fixing #1508.
The current provider test that compares jsii output may no longer be stable with this change.

ansgarm (approved):

PR #1870, opened 2022-06-14
fix: Fixed naming collision between resources of different providers

Maed223:
Fixed the unnecessary renaming of resources between different providers in the provider generator.

Closes: #1804


jsteinich:
Seems like this would supersede #1508 (seems fine since we haven't gotten a signed CLA there and the code changes are very minimal).

Will conflict with #1861: just making a note for reference.

packages/@cdktf/provider-generator/lib/get/constructs-maker.ts
```diff
@@ -236,13 +236,12 @@ export class ConstructsMaker {
...
-    if (providerTargets.length > 0) {
-      new TerraformProviderGenerator(
-        this.code,
-        schema.providerSchema,
-        providerTargets
-      );
-    }
+    providerTargets.forEach(
+      (provider) =>
+        new TerraformProviderGenerator(this.code, schema.providerSchema, [
```
DanielMSchmidt, line 241: We should take a look if there is another case where this class is used. If not we should change the interface so it takes a single provider instead of the list of providers it does now

packages/@cdktf/provider-generator/lib/get/generator/resource-parser.ts
```diff
@@ -17,23 +17,27 @@ import {
...
   return ["string"].includes(className.toLowerCase());
 };
 
 class Parser {
   private structs = new Array<Struct>();
 
+  private classNames: string[] = [];
+
+  constructor(unique_classnames: string[]) {
+    this.classNames = unique_classnames;
```
DanielMSchmidt, line 30: `unique_classnames` should be camel cased.

There is also a neat syntactic sugar one can use in cases like this: `constructor(private classNames: string[]){}`, see [the handbook](https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties) for a better explanation

packages/@cdktf/provider-generator/lib/get/generator/resource-parser.ts
```diff
@@ -488,13 +492,15 @@ class Parser {
 }
 
 export class ResourceParser {
+  private unique_classnames: string[] = [];
```
DanielMSchmidt, line 495: This should be camelCased

DanielMSchmidt (approved):
Looks good to me, looking forward to the refactoring PR we talked about in person :)

Issue #1836, opened 2022-06-01
One stack failure when deploying multiple stack in parallel causes other stacks state to remain locked


aadhikary:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### cdktf & Language Versions

$ cdktf --version
0.10.4

$ terraform --version
Terraform v1.2.1
on linux_amd64

Language: typescript

### Affected Resource(s)

Stack state files that did not fail but were running in parallel

### Expected Behavior
The stacks that did not fail should continue and complete. Once this is complete the error message for the stack that failed should be printed at the end. Also no other new stacks then should run. At the end of a failure the state file should not be locked for any of the stacks.

### Actual Behavior

Remote state file gets locked for the stacks that were running in parallel but did not fail

### Steps to Reproduce

1. Setup cdktf project with multiple stacks which can run in parallel
2. Make sure one of the stack fails while other stacks are not complete
3. Run the stacks in parallel
4. Look at the state file of the stacks that did not fail

skorfmann:
Could you elaborate on the problem and its context you're experiencing - I'm not sure what you're describing here.

aadhikary:
As example lets say we have 3 stacks which can run in parallel and the backend for the state file is on Azure storage account blob storage (could probably be any backend but I am using Azure backend):
stack1, stack2, stack3

If we run all these stacks in parallel:
$cdktf deploy 'stack*' --auto-approve

### Timeline: 
stack1 (deploying)---------------------------------------------------------------> (passed)
stack2 (deploying)---------------------------------------------------------------------------------> (failed)
stack3 (deploying)---------------------------------------------------------------------------------> **(in progress)**

**Note stack3 stops in the middle if the stack2 fails even if stack3 would have finished successfully** 
### Condition of state file:
stack1 ----> No locking on state file (In azure blob terms Available)
stack2 ----> No locking on state file (In azure blob terms Available)
stack3 ----> **State file locked** (In azure blob terms in Lease)

aadhikary:
[...]

PR #1881, opened 2022-06-16
chore: replicate deployment bug

DanielMSchmidt:
Replicates the bug described in #1836


aadhikary:
@DanielMSchmidt any updates on when we might get this released.  This is making use of CDKTF quite complicated.  

Thanks

DanielMSchmidt:
Not really, this PR is just checking if the bug exists and not fixing it so the is nothing to release just yet

PR #1884, opened 2022-06-17
chore: first convert and then init

DanielMSchmidt:
This makes retrying easier

### DIFF
--- a/packages/@cdktf/hcl2cdk/lib/index.ts
+++ b/packages/@cdktf/hcl2cdk/lib/index.ts
@@ -428,8 +428,6 @@ type CdktfJson = Record<string, unknown> & {
 };
 export async function convertProject(
   combinedHcl: string,
-  inputMainFile: string,
-  inputCdktfJson: CdktfJson,
   { language, providerSchema }: ConvertOptions
 ) {
   if (language !== "typescript") {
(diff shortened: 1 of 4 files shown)


packages/@cdktf/hcl2cdk/lib/index.ts
```diff
@@ -446,19 +444,22 @@ export async function convertProject(
...
-  cdktfJson.terraformProviders = providers;
-  cdktfJson.terraformModules = tfModules;
 
   return {
-    code: prettier.format(outputMainFile, { parser: "babel" }),
-    cdktfJson,
+    code: (inputMainFile: string) => {
+      const importMainFile = [imports, inputMainFile].join("\n");
+      const outputMainFile = importMainFile.replace(
+        "// define resources here",
```
ansgarm, line 452: Do we have a test that would fail if we ever accidentally change this in our templates (i.e. change the wording?)
The testcase below in this PR brings its own main.ts containing this comment, so that wouldn't fail.
Maybe we could just add a test that checks if the template contains this string?

ansgarm (approved):

packages/@cdktf/hcl2cdk/lib/index.ts
DanielMSchmidt: Added a test 👍

PR #1924, opened 2022-07-14
feat(lib): add support for cloud backend

### LINKED ISSUE #1583: Support Terraform Cloud configuration
### Community Note

### Description

I'm not sure what the difference is, but the Terraform documentation recommends using the Terraform Cloud Configuration rather than the remote backed.

### References

- https://www.terraform.io/language/settings/backends/remote
- https://www.terraform.io/language/settings/terraform-cloud

DanielMSchmidt:
Closes #1583

### DIFF
--- a/examples/typescript/backends/cloud/cdktf.json
+++ b/examples/typescript/backends/cloud/cdktf.json
@@ -1,0 +1,7 @@
+{
+  "language": "typescript",
+  "app": "npx ts-node main.ts",
+  "terraformProviders": [
+    "aws@~> 3.0"
+  ]
+}
--- a/examples/typescript/backends/cloud/cdktf.json
+++ b/examples/typescript/backends/cloud/cdktf.json
@@ -1,7 +1,5 @@
 {
   "language": "typescript",
   "app": "npx ts-node main.ts",
-  "terraformProviders": [
-    "aws@~> 3.0"
-  ]
+  "terraformProviders": []
 }
--- a/packages/cdktf-cli/lib/cdktf-stack.ts
+++ b/packages/cdktf-cli/lib/cdktf-stack.ts
@@ -91,6 +91,20 @@ async function getTerraformClient(
       return tfClient;
     }
   }
+
+  if (parsedStack.terraform?.cloud) {
+    const tfClient = new TerraformCloud(
+      abortSignal,
+      stack,
+      parsedStack.terraform.cloud,
+      isSpeculative,
+      createTerraformLogHandler
+    );
+    if (await tfClient.isRemoteWorkspace()) {
+      return tfClient;
+    }
+  }
+
   return new TerraformCli(abortSignal, stack, createTerraformLogHandler);
 }
 
--- a/packages/cdktf-cli/lib/cdktf-stack.ts
+++ b/packages/cdktf-cli/lib/cdktf-stack.ts
@@ -6,6 +6,7 @@ import { extractJsonLogIfPresent } from "./server/terraform-logs";
 import { TerraformJson } from "./terraform-json";
 import { TerraformCloud } from "./models/terraform-cloud";
 import { TerraformCli } from "./models/terraform-cli";
+import { Errors } from "./errors";
 
 export type StackUpdate =
   | {
@@ -93,10 +94,17 @@ async function getTerraformClient(
   }
 
   if (parsedStack.terraform?.cloud) {
+    const workspaces = parsedStack.terraform.cloud.workspaces;
+    if (!("name" in workspaces)) {
+      throw Errors.Usage(
+        "The Cloud backend can not used with the cdktf-cli unless specified with a workspace name."
+      );
+    }
+
     const tfClient = new TerraformCloud(
       abortSignal,
       stack,
-      parsedStack.terraform.cloud,
+      { ...parsedStack.terraform.cloud, workspaces },
       isSpeculative,
       createTerraformLogHandler
     );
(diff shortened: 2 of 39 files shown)


barrettclark:
giphy

packages/cdktf/lib/backends/cloud-backend.ts
```diff
@@ -0,0 +1,122 @@
...
+  readonly hostname?: string;
+
+  /**
+   * The token used to authenticate with Terraform Cloud.
+   * We recommend omitting the token from the configuration, and instead using terraform login or manually configuring credentials in the CLI config file.
+   */
+  readonly token?: string;
+}
+
+export abstract class CloudWorkspace {
```
jsteinich, line 95: This appears to be unused.

packages/cdktf/lib/terraform-stack.ts
```diff
@@ -333,7 +333,7 @@ export class TerraformStack extends Construct {
   /**
    * Run all validations on the stack.
    */
-  public validate() {
+  public runChildValidations() {
```
ansgarm, line 336: why not:
```suggestion
  public runAllValidations() {
```
that would fit the comment of the method 😁

examples/typescript/backends/cloud/cdktf.json
```diff
@@ -0,0 +1,7 @@
+{
+  "language": "typescript",
+  "app": "npx ts-node main.ts",
+  "terraformProviders": [
+    "aws@~> 3.0"
```
ansgarm, line 5: let's use a smaller one (as it is not used anyway).
wait.. I think we can also remove it completely as `cdktf get` does not fail anymore if there is nothing to do, right?

packages/cdktf/lib/validations/validate-binary-version.ts
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * A validation that can be applied to a construct that will error if the
+ * construct is being used in an environment with a version of a binary lower than the specified version.
+ */
+export class ValidateBinaryVersion implements IValidation {
+  constructor(
+    protected binary: string,
+    protected versionConstraint: string,
+    protected versionCommand: string,
+    protected hint?: string
```
ansgarm, line 14: nice!

packages/cdktf/test/validations.test.ts
```diff
@@ -54,3 +57,78 @@ class CustomConstruct extends Construct {
...
+    const stack = new TerraformStack(app, "MyStack");
+    new TestProvider(stack, "foo", {});
+    const testResource = new TestResource(stack, "testResource", {
+      name: "foo",
+    });
+    testResource.node.addValidation(
+      new ValidateBinaryVersion(
+        "terraform",
+        ">=1.3.0",
+        `echo "Terraform v1.2.0\non darwin_amd64"`
```
ansgarm, line 74: nice 👏👏

packages/cdktf/lib/backends/cloud-backend.ts
```diff
@@ -0,0 +1,136 @@
...
+  /**
+   * The name of the organization containing the workspace(s) the current configuration should use.
+   */
+  readonly organization: string;
+
+  /**
+   * A nested block that specifies which remote Terraform Cloud workspaces to use for the current configuration.
+   * The workspaces block must contain exactly one of the following arguments, each denoting a strategy for how workspaces should be mapped:
+   */
+  readonly workspaces: NamedCloudWorkspace | TaggedCloudWorkspaces;
```
ansgarm, line 89: Does this work via JSII? Or does this result in e.g. `interface{}` or `any`?
If yes, it might make sense to combine them into a single class that validates that either `name` or `tags` is set.

packages/cdktf-cli/lib/terraform-json.ts
```diff
@@ -18,11 +18,22 @@ export interface TerraformJsonConfigBackendRemote {
...
 }
 
+export interface TerraformJsonConfigCloudWorkspace {
+  name?: string;
+}
+export interface TerraformJsonConfigCloud {
+  organization: string;
+  hostname?: string;
+  token?: string;
+  workspaces: TerraformJsonConfigCloudWorkspace;
```
ansgarm, line 28: this type is missing the `tags` property that could be set instead of `name`

packages/cdktf-cli/lib/cdktf-stack.ts
```diff
@@ -91,6 +91,20 @@ async function getTerraformClient(
       return tfClient;
     }
   }
+
+  if (parsedStack.terraform?.cloud) {
+    const tfClient = new TerraformCloud(
+      abortSignal,
+      stack,
+      parsedStack.terraform.cloud,
```
ansgarm, line 99: `cloud` could include "tags" instead of "name" and would probably crash in that case. We should handle the case when users set tags (i.e. we could tell them they need to use the TF CLI in that case – if that's right).

ansgarm (commented):
couple o' small things

packages/cdktf/lib/backends/cloud-backend.ts
DanielMSchmidt: Yes it works with JSII, we do it the same way for remote

packages/cdktf-cli/lib/cdktf-stack.ts
DanielMSchmidt: Yeah good point, I'll add a validation here

ansgarm (approved):
Currently `cdktf init` does not explicitly check the Terraform version but as of this change would fail if run with TF < 1.1 (without the cloud block).
Our [getting started tutorial](https://learn.hashicorp.com/tutorials/terraform/cdktf-install?in=terraform/cdktf) does mention the requirement TF >= 1.1, however.

Let's at least create a follow-up issue to make this more clear, that init requires TF 1.1 now (if using remote state).

PR #1935, opened 2022-07-19
fix(lib): exposed testing matchers for other languages properly

Maed223:
Testing matchers placed in class for jsii exposure


test/csharp/testing-matchers/test.ts
```diff
@@ -0,0 +1,21 @@
...
+    await driver.copyFiles("update-csproj.sh");
+    await driver.exec("dotnet add package HashiCorp.Cdktf.Providers.Aws");
+    await driver.exec("dotnet add package Microsoft.NET.Test.Sdk");
+    await driver.exec("dotnet add package xunit");
+    await driver.exec("dotnet add package xunit.runner.visualstudio");
+    await driver.exec("bash update-csproj.sh");
+  });
+
+  test("run testing suite", async () => {
+    await driver.exec("dotnet test");
```
DanielMSchmidt, line 19: I would put the result in a var (contains stdout and stderr) and check if there is a tests passed successfully output

test/csharp/testing-matchers/test.ts
```diff
@@ -0,0 +1,21 @@
...
+
+describe("csharp testing assertions", () => {
+  let driver: TestDriver;
+
+  beforeAll(async () => {
+    driver = new TestDriver(__dirname);
+    await driver.setupCsharpProject();
+    await driver.copyFiles("Test.cs");
+    await driver.copyFiles("update-csproj.sh");
+    await driver.exec("dotnet add package HashiCorp.Cdktf.Providers.Aws");
```
DanielMSchmidt, line 11: I think we should use manual provider generation here instead of prebuilt providers

DanielMSchmidt (commented):
Some smaller things

jsteinich:
Having to access the `Pass` property feels a bit clunky and I'm not sure exactly what it adds. Can reference the [AWS CDK testing docs](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.assertions-readme.html) for a comparison point.

website/docs/cdktf/test/unit-tests.mdx
```diff
@@ -5,24 +5,32 @@ description: Write assertions and snapshot tests for your CDK for Terraform appl
 
 # Unit Tests
 
-Testing your application can give you faster feedback cycles and guard you against unwanted changes. Currently, testing is only supported in Typescript with jest.
+Testing your application can give you faster feedback cycles and guard you against unwanted changes. Testing is currently supported in Typescript with jest and compatible with any testing framework that supports assertions for all other languages.
 
-We generate all files necessary to run jest when you run `cdktf init` so that you can start writing tests right away. If you want to add jest to an existing project, please [follow their guide](https://jestjs.io/docs/getting-started). Then, you need to add these lines in a [setup file](https://jestjs.io/docs/configuration#setupfiles-array):
+For Typescript we generate all files necessary to run jest when you run `cdktf init` so that you can start writing tests right away. If you want to add jest to an existing Typescript project, please [follow their guide](https://jestjs.io/docs/getting-started). Then, you need to add these lines in a [setup file](https://jestjs.io/docs/configuration#setupfiles-array):
```
DanielMSchmidt, line 10: Could you do a quick follow up and add the testing setup to all of the templates as well? Should be as simple as copying the example tests you already wrote into a file (commented out) and adding the testing helper and testing instructions

DanielMSchmidt (approved):
Nice work 👍

packages/cdktf/lib/testing/matchers.ts
```diff
@@ -115,35 +129,33 @@ function getAssertElementWithProperties(
...
-        pass,
-        message: () =>
-          `Expected no ${
-            itemType.tfResourceType
-          } with properties ${JSON.stringify(
-            properties,
-            jestAsymetricMatcherStringifyReplacer
-          )} to be present in synthesised stack.
+      return new AssertionReturn(
+        `Expected no ${
```
ansgarm, line 133: @DanielMSchmidt this was here before, but why is it "expected **no**"? Shouldn't it be "expected ...  to be present"?

ansgarm: Or is this for the behavior of Jest? as `pass` is true, Jest will use this error message in case a user chained `.not.` when using the matcher?

test/csharp/testing-matchers/update-csproj.sh
```diff
@@ -0,0 +1,17 @@
...
+
+
+prop=$(get_property)
+f_prop=${prop//$'\n'/\\$'\n'}
+
+if [[ "$OSTYPE" == "darwin"* ]]; then
+  sed -i "" "s|<PropertyGroup>|<PropertyGroup>\n${f_prop}|" "MyTerraformStack.csproj"
+else
+  sed -i.bak "s|<PropertyGroup>|<PropertyGroup>\n${f_prop}|" "MyTerraformStack.csproj"
+fi
```
ansgarm, line 17: What does this script do? A comment after the shebang would be nice

test/java/testing-matchers/update-pom.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+
```
ansgarm, line 2: nit: a short comment about this script adding dependencies for java tests would be nice here as well 👍

website/docs/cdktf/test/unit-tests.mdx
```diff
@@ -47,9 +55,127 @@ describe("Unit testing using assertions", () => {
...
+                    put("name", "ubuntu:latest");
+                }
+    }).getPass());
+  }
+}
+```
+
+```csharp
+using Xunit;
+using HashiCorp.Cdktf; // MyApplicationsAbstraction - Could be a class extending from cdktf.Resource
```
ansgarm, line 111:
```suggestion
using HashiCorp.Cdktf; // MyApplicationsAbstraction - Could be a class extending from Construct
```
I think we recommend Construct now (we used to recomment extending Resource though)
What do you think @DanielMSchmidt?

test/go/testing-matchers/main_test.go
```diff
@@ -0,0 +1,121 @@
...
+var stackInvalid = NewMyStackInvalid(app, "stack-invalid")
+var fullSynthInvalid = cdktf.Testing_FullSynth(stackInvalid)
+
+
+func TestToHaveResourcePass(t *testing.T){
+	assertion := cdktf.Testing_ToHaveResource(synth, docker.Container_TfResourceType())
+
+	if !*assertion.Pass() {
+		t.Fail()
+		t.Error(assertion.Message())
```
ansgarm, line 24: The [docs](https://pkg.go.dev/testing#B.Error) for `t.Fail()` state
> Error is equivalent to Log followed by Fail. 

As I understand it, we can remove the calls to `t.Fail()` and only keep `t.Error(…)` making it shorter 🤓

ansgarm (commented):
Nice work, @Maed223 🚀 
Just some nitpicks from me

website/docs/cdktf/test/unit-tests.mdx
DanielMSchmidt: Yeah, I agree we recommend Construct

packages/cdktf/lib/testing/matchers.ts
DanielMSchmidt: Yeah, exactly that is what is happening under the hood. Jest allows you to have specific error messages for the inverse version. You basically return a message in case the result is not the expected one. So in case true is not the expected one a user must have used not (otherwise true would be expected).

website/docs/cdktf/test/unit-tests.mdx
```diff
@@ -5,28 +5,44 @@ description: Write assertions and snapshot tests for your CDK for Terraform appl
 
 # Unit Tests
 
-Testing your application can give you faster feedback cycles and guard you against unwanted changes. Currently, testing is only supported in Typescript with jest.
+Testing your application can give you faster feedback cycles and guard you against unwanted changes. Testing is currently supported in Typescript with jest and compatible with any testing framework that supports assertions for all other languages.
```
laurapacilio, line 8:
```suggestion
Testing your application can give you faster feedback cycles and guard you against unwanted changes. Testing is supported in Typescript with Jest. For all other languages, you can use any existing testing framework that supports assertions. 
```

Maed223:
In the future the creation of a testing adapter similar to the one for jest would be an improvement. Currently Users have to access the `pass` attribute directly, and the `message` attribute is more jest specific. Further refinement in the future could improve usability.

website/docs/cdktf/test/unit-tests.mdx
```diff
@@ -5,28 +5,44 @@ description: Write assertions and snapshot tests for your CDK for Terraform appl
...
 cdktf.Testing.setupJest();
 ```
 
+If you want to add pytest to an existing Python project, please [follow their guide](https://docs.pytest.org/en/7.1.x/getting-started.html). The CDKTF specific assertions can be found in the cdktf package under `Testing`.
+
+If you want to add JUnit to an existing Java project, please [follow their guide](https://maven.apache.org/plugins-archives/maven-surefire-plugin-2.12.4/examples/junit.html). The CDKTF specific assertions can be found in the cdktf package under `Testing`.
+
+If you want to add Xunit to an existing C# project, please [follow their guide](https://xunit.net/docs/getting-started/netcore/cmdline). The CDKTF specific assertions can be found in the cdktf package under `Testing`.
+
+If you want to add testing to an existing Go project, please [follow their guide](https://go.dev/doc/tutorial/add-a-test). The CDKTF specific assertions can be found in the cdktf package under `Testing`.
```
laurapacilio, line 25: Okay how do we feel about making this an actual section? I think it's generally about how to set up testing and integrate it into your project, right? (I realize there wasn't a section before, so this is new!)

Also - when we say that we autogenerate all the files, is that true for other testing frameworks, or is it just true for Jest with TypeScript?

```suggestion
## Add Testing to Your Application

Use the following resources to add testing to your CDKTF application:
- **TypeScript:** Follow TypeScript's [Getting Started guide](https://jestjs.io/docs/getting-started) for testing with Jest. Then, add these lines in a [setup file](https://jestjs.io/docs/configuration#setupfiles-array):

    ```js
    const cdktf = require("cdktf");
    cdktf.Testing.setupJest();
    ```
- **Python:** Follow the [Get Started guide](https://docs.pytest.org/en/7.1.x/getting-started.html) for pytest. The assertions for CDKTF are available in the `cdktf` package under `Testing`.
- **Java:** Follow the [Using JUnit documentation](https://maven.apache.org/plugins-archives/maven-surefire-plugin-2.12.4/examples/junit.html). The assertions for CDKTF specific are available in the `cdktf` package under `Testing`.
- **C#:** Follow the [Getting Started with xUnit.net guide](https://xunit.net/docs/getting-started/netcore/cmdline). The assertions for CDKTF are available in the `cdktf` package under `Testing`.
- **Go:** Follow the [Add a Test guide](https://go.dev/doc/tutorial/add-a-test) in the Go documentation. The assertions for CDKTF specific are available in the `cdktf` package under `Testing`.

After you add testing to your application, run `cdktf init`. CDKTF autogenerates all of the necessary testing files.
```

website/docs/cdktf/test/unit-tests.mdx
```diff
@@ -5,28 +5,44 @@ description: Write assertions and snapshot tests for your CDK for Terraform appl
...
+
 ### Write Assertions
 
-The example below uses `Testing.synthScope` to test a part of the application. This creates a scope to test a subset of the application and returns a JSON string representing the synthesized HCL-JSON. Then it uses custom matchers to verify the code acts as intended.
+The Typescript example below uses `Testing.synthScope` to test a part of the application. This creates a scope to test a subset of the application and returns a JSON string representing the synthesized HCL-JSON. Then it uses custom matchers to verify the code acts as intended.
```
laurapacilio, line 29:
```suggestion
The following Typescript example uses `Testing.synthScope` to test a part of the application. This creates a scope to test a subset of the application and returns a JSON string representing the synthesized HCL-JSON. Then, it uses custom matchers to verify the code acts as intended.
```

website/docs/cdktf/test/unit-tests.mdx
```diff
@@ -5,28 +5,44 @@ description: Write assertions and snapshot tests for your CDK for Terraform appl
...
+
+The other examples use `Testing.synth` to test a part of the application. Given the desired scope to test, a JSON string representing the synthesized HCL-JSON is returned. Then the custom assertions under `Testing` in the cdktf package can be used to verify the code acts as intended.
```
laurapacilio, line 31:
```suggestion
The examples for other languages use `Testing.synth` to test a part of the application. CDKTF returns a JSON string representing the synthesized HCL-JSON is returned. Then, the custom assertions under `Testing` in the `cdktf` package verify the code acts as intended.
```

website/docs/cdktf/test/unit-tests.mdx
```diff
@@ -47,14 +63,130 @@ describe("Unit testing using assertions", () => {
...
+	}
+}
+```
+
+</CodeTabs>
+
 ### Snapshot Testing
 
-Snapshot tests are useful when you want to make sure your infrastructure does not change unexpectedly. You can read more about them in the [Jest docs](https://jestjs.io/docs/snapshot-testing).
+Snapshot tests are useful when you want to make sure your infrastructure does not change unexpectedly. Snapshot Testing is currently only supported in Typescript with jest You can read more about them in the [Jest docs](https://jestjs.io/docs/snapshot-testing).
```
laurapacilio, line 184:
```suggestion
Snapshot tests are useful when you want to make sure your infrastructure does not change unexpectedly. Snapshot Testing is only supported in Typescript with Jest. Refer to the [Jest docs](https://jestjs.io/docs/snapshot-testing) for details.
```

website/docs/cdktf/test/unit-tests.mdx
```diff
@@ -102,6 +238,80 @@ describe("Checking validity", () => {
 });
 ```
```
laurapacilio, line 240: We need an intro comment explaining what the examples do :)

laurapacilio (changes requested):
Excellent work on this! A couple of quick changes and then I'd like to take one final look. Thank you!

website/docs/cdktf/test/unit-tests.mdx
Maed223: So it is true that all language will have all the generated files needed to start testing, great idea to add a separate section for adding testing to existing project 😁

test/csharp/testing-matchers/test.ts
```diff
@@ -0,0 +1,25 @@
...
+    await driver.copyFiles("update-csproj.sh");
+    await driver.exec("dotnet add reference .gen/docker/docker.csproj");
+    await driver.exec(
+      "dotnet add package Microsoft.NET.Test.Sdk --version 17.2.0"
+    );
+    await driver.exec("dotnet add package xunit --version 2.4.1");
+    await driver.exec(
+      "dotnet add package xunit.runner.visualstudio --version 2.4.5"
+    );
+    await driver.exec("bash update-csproj.sh");
```
jsteinich: Won't this fail on windows?

jsteinich: Also, why not just copy in an already build project file?

test/java/testing-matchers/test.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  beforeAll(async () => {
+    driver = new TestDriver(__dirname);
+    await driver.setupJavaProject();
+    await driver.copyFile(
+      "JunitTesting.java",
+      "src/main/java/com/mycompany/app/JunitTesting.java"
+    );
+    await driver.copyFiles("update-pom.sh");
+    await driver.exec("bash update-pom.sh");
```
jsteinich: Why not just copy an updated pom file?

website/docs/cdktf/test/unit-tests.mdx
```diff
@@ -47,14 +63,130 @@ describe("Unit testing using assertions", () => {
...
+  @Test
+  void shouldContainContainer() {
+    assertTrue(Testing.toHaveResource(synthesized, Container.TF_RESOURCE_TYPE).getPass());
+  }
+
+  @Test
+  void shouldUseUbuntuImage() {
+    assertTrue(Testing.toHaveResourceWithProperties(synthesized, Image.TF_RESOURCE_TYPE, new HashMap<String, Object>() {
+                {
+                    put("name", "ubuntu:latest");
```
jsteinich, line 109: Normally recommend against creating an anonymous map class, but otherwise need to pull in additional dependencies to keep code concise so 🤷

website/docs/cdktf/test/unit-tests.mdx
```diff
@@ -47,14 +63,130 @@ describe("Unit testing using assertions", () => {
...
+
+namespace MyCompany.MyApp{
+  public class TestApplication{
+
+    private static TerraformStack stack = new TerraformStack(Testing.app(), "stack");
+    private static MyApplicationsAbstraction appAbstraction = new MyApplicationsAbstraction(stack, "resource");
+    private static string synthesized = Testing.synth(stack);
+
+    [Fact]
+    public void shouldContainContainer(){
```
jsteinich, line 132:
```suggestion
    public void ShouldContainContainer(){
```

website/docs/cdktf/test/unit-tests.mdx
```diff
@@ -47,14 +63,130 @@ describe("Unit testing using assertions", () => {
...
+using HashiCorp.Cdktf; // MyApplicationsAbstraction - Could be a class extending from Construct
+using MyCompany.MyApp;
+using System;
+using System.Collections.Generic;
+using docker;
+
+namespace MyCompany.MyApp{
+  public class TestApplication{
+
+    private static TerraformStack stack = new TerraformStack(Testing.app(), "stack");
```
jsteinich, line 127: I'd opt for putting this in test setup methods rather than just static fields.

website/docs/cdktf/test/unit-tests.mdx
```diff
@@ -102,6 +238,80 @@ describe("Checking validity", () => {
...
+using System;
+
+namespace MyCompany.MyApp{
+  public class TestApplication{
+
+    private static TerraformStack stack = new TerraformStack(Testing.app(), "stack");
+    private static MyApplicationsAbstraction appAbstraction = new MyApplicationsAbstraction(stack, "construct");
+
+    [Fact]
+    public void checkValidity(){
```
jsteinich, line 287:
```suggestion
    public void CheckValidity(){
```

jsteinich (commented):
I mostly focused on the C# and Java pieces since I'm not familiar with unit testing in Python or Go.
Really just minor points about trying to align closer with conventions of the specific languages.

jsteinich:
> In the future the creation of a testing adapter similar to the one for jest would be an improvement. Currently Users have to access the `pass` attribute directly, and the `message` attribute is more jest specific. Further refinement in the future could improve usability.

The problem I see with delaying changes is that we are likely to encounter breaking changes or further suboptimal experience when adding in the future.
Since the additional return information seems to be targeted towards jest at this point, and since jest has an additional adapter, I'd recommend having internal functions that both the jest adapter and public methods make use of. This would still allow the same code and increased jest support to be present, while simplifying the public signatures for other languages to just return a boolean.

packages/cdktf-cli/templates/java/pom.xml
```diff
@@ -24,11 +24,21 @@
       <artifactId>constructs</artifactId>
       <version>{{ constructs_version }}</version>
     </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.13.2</version>
+    </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
```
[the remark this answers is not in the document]
DanielMSchmidt: It seems like junit-jupiter is the new cool thing, so I'm ok with using it for now. I would assume folks who care about their specific version are knowledgeable enough to change it :)

PR #1936, opened 2022-07-19
chore(docs): Update overview.mdx

laurapacilio:
A few tiny updates to the new API reference overview page :)

### DIFF
--- a/website/docs/cdktf/api-reference/overview.mdx
+++ b/website/docs/cdktf/api-reference/overview.mdx
@@ -1,5 +1,5 @@
 ---
-page_title: API Reference
+page_title: API Reference - CDKTF Core Library
 description: >-
   Use the CDKTF core library with generated provider bindings to write your CDKTF program. Get API reference in TypeScript, Python, Java, CSharp, and Go.
 ---
--- a/website/docs/cdktf/api-reference/overview.mdx
+++ b/website/docs/cdktf/api-reference/overview.mdx
@@ -1,7 +1,7 @@
 ---
 page_title: API Reference - CDKTF Core Library
 description: >-
-  Use the CDKTF core library with generated provider bindings to write your CDKTF program. Get API reference in TypeScript, Python, Java, CSharp, and Go.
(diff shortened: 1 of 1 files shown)


website/docs/cdktf/api-reference/overview.mdx
```diff
@@ -1,17 +1,17 @@
 ---
 page_title: API Reference
```
laurapacilio, line 2:
```suggestion
page_title: API Reference - CDKTF Core Library
```

website/docs/cdktf/api-reference/overview.mdx
```diff
@@ -1,17 +1,17 @@
 ---
-page_title: API Reference
+page_title: API Reference - CDKTF Core Library
 description: >-
-  The CDKTF core library is what you use to specify your CDKTF program alongside with the generated provider bindings.
+  Use the CDKTF core library with generated provider bindings to write your CDKTF program. Get API reference in TypeScript, Python, Java, CSharp, and Go.
```
laurapacilio, line 4:
```suggestion
  Use the CDKTF core library with generated provider bindings to write your CDKTF program. It is available in TypeScript, Python, Java, CSharp, and Go.
```

website/docs/cdktf/api-reference/overview.mdx
```diff
@@ -1,17 +1,17 @@
 ---
-page_title: API Reference
+page_title: API Reference - CDKTF Core Library
```
laurapacilio, line 2: I was thinking maybe we should be specific that this is for the core CDK library in the title and not providers, etc. Do you think this makes sense to do, or you think it's confusing?

DanielMSchmidt (approved):

website/docs/cdktf/api-reference/overview.mdx
DanielMSchmidt: Sounds good 👍

laurapacilio:
@DanielMSchmidt could you merge this? I still don't have permissions :)

PR #1937, opened 2022-07-19
chore(docs): Update product maturity and go support statements

schersh:
Update product maturity and go support statements in preparation for GA release.


website/docs/cdktf/index.mdx
```diff
@@ -9,7 +9,7 @@ description: >-
 
 Cloud Development Kit for Terraform (CDKTF) allows you to use familiar programming languages to define and provision infrastructure. This gives you access to the entire Terraform ecosystem without learning HashiCorp Configuration Language (HCL) and lets you leverage the power of your existing toolchain for testing, dependency management, etc.
 
-We currently support TypeScript, Python, Java, C#, and Go (experimental).
+We currently support TypeScript, Python, Java, C#, and Go.
```
laurapacilio, line 12:
```suggestion
We support TypeScript, Python, Java, C#, and Go.
```

laurapacilio (commented):
One language nit :)  I support removing the entire project maturity statement. But I also think it makes sense to do whatever core did. I just know know - is there value in explicitly saying that we're not at 1.0 and what that means? I leave this to the team, but I do think we should make sure that users know what to expect (compatibility promises, support level, etc.) and so I just want to make sure no one is surprised!

jsteinich:
> One language nit :) I support removing the entire project maturity statement. But I also think it makes sense to do whatever core did. I just know know - is there value in explicitly saying that we're not at 1.0 and what that means? I leave this to the team, but I do think we should make sure that users know what to expect (compatibility promises, support level, etc.) and so I just want to make sure no one is surprised!

I agree with this sentiment. Some users will see pre 1.0 and treat it as if that project maturity statement where still in place. Not everyone visiting this project will be familiar with HashiCorp versioning standards, so some hints that there is more support would probably be good. I would expect a GA release to at least strongly adhere to semantic versioning (we've been a bit loose at times).

schersh:
@laurapacilio I confirmed with the Core team that they did not have any type of pre-1.0 statement in their docs, and other HashiCorp pre-1.0 products (like Boundary) don't either. So I think it makes sense to just remove this section.

PR #1947, opened 2022-07-21
fix: use main stack and call class program

DanielMSchmidt:


jsteinich (changes requested):
This will break the remote backend replacement

packages/cdktf-cli/bin/cmds/helper/terraform-login.ts
```diff
@@ -114,12 +114,16 @@ the following file for use by subsequent Terraform commands:
...
+    try {
+      const credentialsFile = JSON.parse(
+        fs.readFileSync(terraformCredentialsFilePath).toString()
+      );
+      const terraformCredentials: TerraformCredentialsFile = credentialsFile;
 
-    return terraformCredentials;
+      return terraformCredentials;
+    } catch (e) {
+      return { credentials: {} };
```
ansgarm, line 125: we should at least add a trace / debug log here to debug this if the future might bring errors with weird setups once more 😅

ansgarm (approved):

Issue #1855, opened 2022-06-08
Generating Java code for the akamai provider results in ENAMETOOLONG error


erikgull:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### cdktf & Language Versions

```
$ cdktf --version
0.10.4
$ terraform --version
Terraform v1.2.0
```

### Affected Resource(s)

Akamai Terraform Provider

### Debug Output

https://gist.github.com/erikgull/bfca2c4441921bd2d78d114d069fda26

### Expected Behavior

We expected the command `cdktf get` to generate the Java classes for the akamai provider. 

### Actual Behavior

The command `cdktf get` exits with non-zero code and a message mentioning ENAMETOOLONG.

### Steps to Reproduce

```
$ cd tfcdk-first-test/
$ cdktf init --local --template="java"
$ cat <<< $(jq '.terraformProviders = [ {"name":"akamai", "source":"akamai/akamai", "version":"2.0.0"} ]' cdktf.json) > cdktf.json
$ cdktf get
⠹ downloading and generating modules and providers...

This command fails with 
Error: ENAMETOOLONG: name too long, open '/private/var/folders/yz/1p6r636n27n9sjjmj5v251d40000gp/T/temp-uR0UKb/dist/java/src/main/java/imports/akamai/DataAkamaiIamGroupsGroupsSubGroupsSubGroupsSubGroupsSubGroupsSubGroupsSubGroupsSubGroupsSubGroupsSubGroupsSubGroupsSubGroupsSubGroupsSubGroupsSubGroupsSubGroupsSubGroupsSubGroupsSubGroupsSubGroupsSubGroupsSubGroupsSubGroupsSubGroupsSubGroupsOutputReference.java'
```

It appears that there is something causing the code generator to recurse very deep with the Akamai IAMGroups when generating the Java class/file name before it fails with the above error.

### Important Factoids

Environment is MacOS 12.4.

### References

N/A

jsteinich:
[...]

PR #1951, opened 2022-07-22
fix(provider-generator): remove "recursion" from generated code

DanielMSchmidt:
The provider schema can have a logical recursion that is due to JSONs staticness
transcribed by repeating the recursive struct n times. This blows up our generated provider
bindings to a degree where we can no longer built go in github action runners.

To fix this we now detect the recursion and use recursive references in Typescript.

### DIFF
--- a/packages/@cdktf/provider-generator/lib/get/generator/loop-detection.ts
+++ b/packages/@cdktf/provider-generator/lib/get/generator/loop-detection.ts
@@ -1,0 +1,77 @@
+import { AttributeModel } from "./models/attribute-model";
+
+type AttributeIdentifier = [string, Set<string>];
+const equalAttributeIdentifiers = (
+  x: AttributeIdentifier,
+  y: AttributeIdentifier
+) =>
+  x[0] === y[0] &&
+  x[1].size === y[1].size &&
+  [...x[1]].every((x) => y[1].has(x));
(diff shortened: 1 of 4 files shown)


ansgarm (approved):
image

packages/@cdktf/provider-generator/lib/get/generator/loop-detection.ts
```diff
@@ -0,0 +1,83 @@
...
+  x[0] === y[0] &&
+  x[1].size === y[1].size &&
+  [...x[1]].every((x) => y[1].has(x));
+
+function typeStructure(model: AttributeModel) {
+  if (model.type.isPrimitive) {
+    return model.type.name;
+  }
+
+  return "<complex>";
```
jsteinich, line 17: This seems like it has the potential for a false match if attribute names on different levels match, but they have different value types.

jsteinich:
This should resolve #1855

packages/@cdktf/provider-generator/lib/get/generator/loop-detection.ts
DanielMSchmidt: I completely agree with you, will improve it in a follow up, I want to check this against the prebuilt provider

PR #1952, opened 2022-07-22
feat(docs): add docs for block level iteration

ansgarm:
- feat(docs): add docs for block level iteration

### DIFF
--- a/website/docs/cdktf/concepts/iterators.mdx
+++ b/website/docs/cdktf/concepts/iterators.mdx
@@ -61,9 +61,13 @@ const s3Bucket = new s3.Bucket(this, "bucket", {
 });
 ```
 
-## Using Iterators for list attributes
+## Using Iterators for List Attributes
 
-Besides using the iterator to create multiple resources at the top level, it is also possible to create a list of objects based on each item in the list and assign the result as a value to a property of a resource. The equivalent in native TypeScript code would be using `Array.map` and when using Terraform with HCL it would be using dynamic blocks. Use iterators for list attributes if the length of the list is not known before deploying, else use native functions that are available in your language, e.g. `Array.map` in TypeScript.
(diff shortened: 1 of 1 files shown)


ansgarm:
Preview: https://terraform-735bn3rry-hashicorp.vercel.app/cdktf/concepts/iterators#using-iterators-for-list-attributes

website/docs/cdktf/concepts/iterators.mdx
```diff
@@ -60,3 +60,141 @@ const s3Bucket = new s3.Bucket(this, "bucket", {
   tags: iterator.getStringMap("tags"),
 });
 ```
+
+## Using Iterators for list attributes
+
+Besides using the iterator to create multiple resources at the top level, it is also possible to create a list of objects based on each item in the list and assign the result as a value to a property of a resource. The equivalent in native TypeScript code would be using `Array.map` and when using Terraform with HCL it would be using dynamic blocks. Use iterators for list attributes if the length of the list is not known before deploying, else use native functions that are available in your language, e.g. `Array.map` in TypeScript.
```
laurapacilio, line 66: Let me know if my explanation of the example is wrong!

```suggestion
You can also use iterators to create a list of objects based on each item in a list and assign the result as a value to a property of a resource. This is equivalent to using `Array.map` in TypeScript and using [dynamic blocks](/language/expressions/dynamic-blocks) in a Terraform HCL configuration. 

Use iterators for list attributes if the length of the list is not known before deploying. Otherwise, use native functions that are available in your language (e.g., `Array.map` in TypeScript).

The examples below use an iterator to create a list of members for each organization.
```

laurapacilio (approved):
Approving so I don't block - a couple of updates for you!

ansgarm:
As always, thank you so much @laurapacilio!

I changed the last sentence a tiny bit to reflect what the example is doing exactly to:

> The examples below use an iterator to create a team containing each member of an organization.

PR #1959, opened 2022-07-26
chore(docs): comment on tfc execution mode

DanielMSchmidt:


website/docs/cdktf/create-and-deploy/terraform-cloud.mdx
```diff
@@ -47,6 +47,7 @@ For new projects, do the following:
 1. Choose whether to send crash reports to the CDKTF team.
 
 CDKTF creates both a scaffolded project in your chosen language and a new Terraform Cloud workspace. Your project is connected to the workspace and you can begin using CDKTF and Terraform Cloud together.
+By default, the workspace is set to Local Execution mode, which means Terraform continues to run on your local machine. If you want Terraform to run remotely, you need to set the workspace to Remote Execution mode.
```
ansgarm, line 50: You could also link to the Terraform Cloud docs, something along the lines of: "To learn more about remote operations, visit the [Terraform Cloud docs](https://www.terraform.io/cloud-docs/run/remote-operations)."

But I'm sure that @laurapacilio could phrase this way better 😁

laurapacilio, line 50: I like it! :) How about linking like this (thank you, ansgar!)

```suggestion
By default, the workspace is set to Local Execution mode, which means Terraform continues to run on your local machine. If you want Terraform to run remotely, you need to set the workspace to [Remote Execution mode](/cloud-docs/workspaces/settings#execution-mode).
```

laurapacilio: Also @DanielMSchmidt one thing I noticed is that the numbering after the code blocking in our "existing projects" section isn't 3. Can we fix that? I think it's probably because the code block isn't indented properly. You may need to try it out and run the site locally to make sure the formatting is correct!

Thank you! 

Screen Shot 2022-07-26 at 11 09 26 AM

laurapacilio (approved):
Pre-approving, but @DanielMSchmidt see my note about the formatting of the code block in the "Existing Projects" section. Thank you both!! Excited to see this go out!

PR #1960, opened 2022-07-26
chore: ensure we have license header

DanielMSchmidt:
- chore: add copyright notice to license
- chore: add license headers to all files


tools/add-copyright-header.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -ex
+scriptdir="$(cd $(dirname $0) && pwd)"
+
+for i in $(git ls-files | grep '.ts$'); do
+  if ! grep -q Copyright $i
```
ansgarm, line 7: Is there a way to tell grep to only look at the first line? As this would ignore files that contain the string `Copyright` somewhere else too, right?

examples/typescript/aws-cloudfront-proxy/__tests__/main-test.ts
```diff
@@ -1,3 +1,5 @@
+// Copyright (c) HashiCorp, Inc
+// Licensed under the Mozilla Public License 2.0
```
ansgarm, line 2: the RFC about this requirement was updated and now requires the following it seems:
```suggestion
// SPDX-License-Identifier: MPL-2.0
```

ansgarm (changes requested):

ansgarm (approved):

PR #1966, opened 2022-07-27
chore: add sizing info on the examples so e2e ones are easier to identify

DanielMSchmidt:


website/docs/cdktf/examples.mdx
```diff
@@ -26,72 +26,76 @@ Follow these hands-on tutorials from HashiCorp Learn:
...
-| [ucloud](https://github.com/hashicorp/terraform-cdk/tree/main/examples/typescript/ucloud)                                | Provisions a Linux base image on UCloud.                                                                                                                                                         |
-| [vault](https://github.com/hashicorp/terraform-cdk/tree/main/examples/typescript/vault)                                  | Creates a Mount in Vault.                                                                                                                                                                        |
+| Example                                                                                                                  | Description                                                                                                                                                                                      | Size  |
```
laurapacilio, line 29: I like this and the way that you've changed the names to have "end-to-end"! 

Instead of size, I wonder if instead we should say "Complexity" and say that it's either High or Low. I wonder if that will communicate the meaning a little bit better than size (which I guess works, but I feel like maybe complexity is more precise?)

PR #1973, opened 2022-07-28
fix: provider add uses local provider for google in go

DanielMSchmidt:
google and google-beta can not build go bindings yet
the GH Action runners don't have enough RAM to support this


DanielMSchmidt:
We manually build the go prebuilt provider for now

PR #1974, opened 2022-07-28
chore: add a script for manually publishing go prebuilt providers for google

DanielMSchmidt:


tools/build-prebuilt-golang-provider.sh
```diff
@@ -0,0 +1,47 @@
...
+# due to OOM errors in some instances, we need to build and push the provider manually.
+
+PROVIDERS=(google googlebeta azurerm kubernetes)
+
+if [ -z "$GITHUB_TOKEN" ]; then
+    echo "Set the GITHUB_TOKEN environment variable to a personal github token with enough permissions."
+    exit 1;
+fi
+
+# Some of use use gpg signing, but that does not work with the automation, so we need to disable it first.
```
ansgarm, line 15:
```suggestion
# Some of us use gpg signing, but that does not work with the automation, so we need to disable it first.
```

tools/build-prebuilt-golang-provider.sh
```diff
@@ -0,0 +1,47 @@
...
+    echo "Installing provider dependencies"
+    yarn
+
+    echo "Set version in package.json"
+
+    CURRENT_VERSION=$(git -c "versionsort.suffix=-" tag --sort="-version:refname" --list "v*" | head -n 1 | sed 's/^v//')
+    jq -M -r ".version = \"$CURRENT_VERSION\"" package.json > package.json.new
+    mv package.json.new package.json
+
+    # We don't use built to get around packaging for all the languages.
```
ansgarm, line 39:
```suggestion
    # We don't use "build" to get around packaging for all the languages.
```

tools/build-prebuilt-golang-provider.sh
```diff
@@ -0,0 +1,47 @@
...
+    echo "Compiling ${PROVIDER} provider"
+    yarn compile
+
+    echo "Packaging ${PROVIDER} provider"
+    yarn package:go
+
+    GIT_USER_NAME="CDK for Terraform Team" GIT_USER_EMAIL="github-team-tf-cdk@hashicorp.com" GITHUB_TOKEN="$GITHUB_TOKEN" npx --yes -p publib@latest publib-golang
```
ansgarm, line 46: Using a specific version of publib could be safer if we don't want this to accidentally break in the future.
That said, we could also add a comment with the current version that works, so that we can use that as soon as it starts breaking 😁

tools/build-prebuilt-golang-provider.sh
```diff
@@ -0,0 +1,47 @@
...
+    cd "$tmp_dir"
+    echo "Cloning ${PROVIDER} provider into $tmp_dir"
+    git clone "git@github.com:hashicorp/cdktf-provider-$PROVIDER.git" .
+
+    echo "Installing provider dependencies"
+    yarn
+
+    echo "Set version in package.json"
+
+    CURRENT_VERSION=$(git -c "versionsort.suffix=-" tag --sort="-version:refname" --list "v*" | head -n 1 | sed 's/^v//')
```
ansgarm, line 35: So this has to be run after a successful release of the pre-built provider, right?

ansgarm (approved):
Looks good.
If we don't run this for every release, we need to add a bit more graceful error handling to the `cdktf provider add` command as that one expects pre-built provider versions for any language to exist (if they exist for NPM).
We could add a check for Go and print a warning if that pre-built provider exists but not for Go, as Go needs a workaround at the moment.

That said, there could always be cases where a package might be missing due to various reasons in any of the languages, so we could also add such graceful handling to all languages at once.

tools/build-prebuilt-golang-provider.sh
DanielMSchmidt: Not sure if this is worth the change, I mean we are probably removing the script sooner than later

PR #1980, opened 2022-07-28
keep *.ts in npm

erikwoo:
make debug easier


ansgarm:
Hi @erikwoo 👋

Shipping *`.ts` files alongside the code introduces a variety of problems (as you can see with the failed tests) – I agree that it is easier to read TypeScript code when something fails in the `cdktf` lib. If you want to debug that, I'd recommend cloning this repository and linking the source locally – that's the way we debug and iterate ([CONTRIBUTING.md](https://github.com/hashicorp/terraform-cdk/blob/main/CONTRIBUTING.md) should help with getting that set up).

However, in the long-term we aim for fixing the bugs you might encounter and to make our core library more robust and our error messages more explanatory. This will also help users that use CDKTF with other languages than TypeScript.

I hope you can understand our reasons here and are okay with me closing this PR.

PR #1984, opened 2022-08-01

team-tf-cdk:


ansgarm (approved):

ansgarm:
Hm, with the update of JSII, this does not seem to work anymore in Python:
image
It fails with
> TypeError: type of argument singlereq must be one of (imports.edge.ListBlockResourceSinglereq, Dict[str, Any]); got imports.edge.ListBlockResourceSinglereqOutputReference instead

(Note: this also does not work in Go for example)

ansgarm:
It was probably introduced by this change in JSII: https://github.com/aws/jsii/pull/3660

However, our example does not feel very real-worldy there. Because usually if you want to reuse what you passed to some resource in another resource of the same type, you would be able to use e.g. `tagsInput` (for `tags`) to access the input type. So our example here more or less tests the very rare case, where an argument of some resource (or data source) accidentally matches the input of a different resource. And while this kind of type overlap does work in TypeScript, it may not necessarily work in other languages (such as Go, or – as of this change – Python). If users still need to do this, they could still explicitly use an Iterator (which would also work for assigning block type attributes).

What do you think, @DanielMSchmidt @jsteinich, should we change that example and accept that JSII does not accept this kind of passthrough anymore? (I'm leaning towards yes, but wanted to know, what you two think.)

jsteinich:
> What do you think, @DanielMSchmidt @jsteinich, should we change that example and accept that JSII does not accept this kind of passthrough anymore? (I'm leaning towards yes, but wanted to know, what you two think.)

I'm fine with changing / removing the test case. This already only worked in TypeScript and Python, and as you stated, was probably not a real / safe use case.

PR #1987, opened 2022-08-02
fix(cli): Wait for other stacks to complete if one of them failed

### LINKED ISSUE #1836: One stack failure when deploying multiple stack in parallel causes other stacks state to remain locked

ansgarm:
- chore: replicate deployment bug
- fix(cli): fix test as planning steps have no ensured order
- fix(cli): Wait for other work promises to complete even if one of them failed, which causes .race and .all to short circuit

Closes #1836

### DIFF
--- a/packages/cdktf-cli/lib/cdktf-project.ts
+++ b/packages/cdktf-cli/lib/cdktf-project.ts
@@ -550,11 +550,15 @@ export class CdktfProject {
             "Encountered an error while awaiting stack to finish",
             e
           );
-          logger.debug("Waiting for still running stacks to finish");
+          const openStacks = this.stacksToRun.filter(
+            (ex) => ex.currentWorkPromise
+          );
+          logger.debug(
+            "Waiting for still running stacks to finish:",
+            openStacks
+          );
           await Promise.allSettled(
-            this.stacksToRun
-              .filter((ex) => ex.currentWorkPromise)
-              .map((ex) => ex.currentWorkPromise)
+            openStacks.map((ex) => ex.currentWorkPromise)
           );
           logger.debug(
             "Done waiting for still running stacks. All pending work finished"
--- a/packages/cdktf-cli/lib/cdktf-project.ts
+++ b/packages/cdktf-cli/lib/cdktf-project.ts
@@ -8,6 +8,7 @@ import { NestedTerraformOutputs } from "./output";
 import { logger } from "./logging";
 import minimatch from "minimatch";
 import { createEnhanceLogMessage } from "./execution-logs";
+import { ensureAllSettledBeforeThrowing } from "./util";
 
 type MultiStackApprovalUpdate = {
   type: "waiting for approval";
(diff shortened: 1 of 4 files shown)


packages/cdktf-cli/lib/cdktf-project.ts
```diff
@@ -527,24 +534,49 @@ export class CdktfProject {
...
+            ? nextRunningExecutor.deploy(refreshOnly)
+            : nextRunningExecutor.destroy();
+        } catch (e) {
+          // await next() threw an error because a stack failed to apply/destroy
+          // wait for all other currently running stacks to complete before propagating that error
+          logger.debug(
+            "Encountered an error while awaiting stack to finish",
+            e
+          );
+          logger.debug("Waiting for still running stacks to finish");
```
DanielMSchmidt, line 553: Nit: Maybe we can add the currently open stacks here for a better debugging log line

packages/cdktf-cli/lib/cdktf-project.ts
```diff
@@ -527,24 +534,49 @@ export class CdktfProject {
...
     }
 
     // We end the loop when all stacks are started, now we need to wait for them to be done
-    await Promise.all(
-      this.stacksToRun
-        .filter((ex) => ex.currentWorkPromise)
-        .map((ex) => ex.currentWorkPromise)
-    );
+    // We wait for all work to finish even if one of the promises threw an error.
+    const currentWork = this.stacksToRun
```
DanielMSchmidt, line 569: I wonder if we can make this part a method, seems like we have a core function to execute and wait for all promises to settle in the event of an error

packages/cdktf-cli/test/lib/cdktf-project.test.ts
```diff
@@ -350,6 +368,66 @@ describe("CdktfProject", () => {
...
+        throw new Error("This error should not be thrown");
+      } catch (e) {
+        expect(e).toMatchInlineSnapshot(`[Error: non-zero exit code 1]`);
+      }
+
+      const relevantEvents = events
+        .filter((e) => !e.type.includes("update"))
+        .map((e) => `${e.stackName || "global"}: ${e.type}`);
+
+      // the first 5 events have a stable order
```
DanielMSchmidt, line 399: Do we even need these?

DanielMSchmidt (approved):

packages/cdktf-cli/test/lib/cdktf-project.test.ts
ansgarm: I think it still makes sense to test that these events correctly occur. This helps spotting issues that cause the test to fail, that might not be directly related to the other events which we test. E.g. if one of those first five events was missing, the other assertions would fail and it would be hard to spot why everything shifted in the array by one element (due to the missing event that is never mentioned in the test).

packages/cdktf-cli/lib/cdktf-project.ts
ansgarm: done

packages/cdktf-cli/test/lib/cdktf-project.test.ts
DanielMSchmidt: Fair point 👍

PR #1990, opened 2022-08-02
fix: don't lock state in tests

danielBreitlauch:
The test matcher `toPlanSuccessfully` executes a full Terraform plan which needs to look at potentially locked remote state.
This locks the state and interferes with tests running in parallel etc.

On the other hand, disabling the locking in tests should have no negative affect.
Hence the PR.


ansgarm (approved):
Thank you @danielBreitlauch 👍 Sounds sensible to me.

ansgarm:
(force pushed to fix the commit message to please our linter)

PR #2014, opened 2022-08-10
chore: start linting for jsdoc strings in core library

DanielMSchmidt:
This approach will help us document new parts directly, while we can gradually improve our documentation step by step.

Successor of #1946 (instead of trying to do it in one step, we can do it incrementally)

### DIFF
--- a/packages/cdktf/lib/terraform-functions.ts
+++ b/packages/cdktf/lib/terraform-functions.ts
@@ -12,7 +12,7 @@ type TFValueValidator = (value: any) => TFValue;
 type ExecutableTfFunction = (...args: any[]) => IResolvable;
 
 /**
- * Determines if given str had unescaped double quotes
+ * Determines if given str has unescaped double quotes
  * @param str String to test
  * @returns Boolean
  */
--- a/packages/cdktf/lib/terraform-functions.ts
+++ b/packages/cdktf/lib/terraform-functions.ts
@@ -21,14 +21,17 @@ function hasUnescapedDoubleQuotes(str: string) {
 }
 
 // Validators
+// eslint-disable-next-line jsdoc/require-jsdoc
 function anyValue(value: any): any {
   return value;
 }
 
+// eslint-disable-next-line jsdoc/require-jsdoc
 function mapValue(value: any): any {
   return value;
 }
 
+// eslint-disable-next-line jsdoc/require-jsdoc
 function stringValue(value: any): any {
   if (typeof value !== "string" && !Tokenization.isResolvable(value)) {
     throw new Error(`'${value}' is not a valid string nor a token`);
@@ -43,6 +46,7 @@ function stringValue(value: any): any {
   return value;
 }
 
+// eslint-disable-next-line jsdoc/require-jsdoc
 function numericValue(value: any): any {
   if (typeof value !== "number" && !Tokenization.isResolvable(value)) {
     throw new Error(`${value} is not a valid number nor a token`);
@@ -50,6 +54,7 @@ function numericValue(value: any): any {
   return value;
 }
 
+// eslint-disable-next-line jsdoc/require-jsdoc
 function listOf(type: TFValueValidator): TFValueValidator {
   return (value: any) => {
     if (Tokenization.isResolvable(value)) {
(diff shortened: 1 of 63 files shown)


packages/cdktf/lib/terraform-functions.ts
```diff
@@ -11,19 +11,27 @@ type TFValueValidator = (value: any) => TFValue;
 
 type ExecutableTfFunction = (...args: any[]) => IResolvable;
 
+/**
+ * Determines if given str had unescaped double quotes
```
ansgarm, line 15:
```suggestion
 * Determines if given str has unescaped double quotes
```

packages/cdktf/lib/testing/adapters/__tests__/index.test.ts
```diff
@@ -37,6 +37,9 @@ describe("#synthScope", () => {
   });
 
   test("using resource", () => {
+    /**
+     *
+     */
```
ansgarm, line 42: could we ignore `*.test.ts` files for the jsdoc rules?
Else I'd rather but an eslint-disable comment at the root of each test and get rid of these empty comments

packages/cdktf/lib/testing/matchers.ts
```diff
@@ -29,8 +31,13 @@ export function returnMatcherToJest(
   };
 }
 
-// All expected properties are matched and considered equal if
-// There can be more properties in the received object than in the expected object while still returning true
+/**
+ * Compares expected and recieved. All expected properties are matched and considered equal if
```
ansgarm, line 35:
```suggestion
 * Compares expected and received. All expected properties are matched and considered equal if
```

ansgarm: also: the sentence just stops after "if" 😅

packages/cdktf/lib/tokens/resolvable.ts
```diff
@@ -151,6 +154,9 @@ export class StringConcat implements IFragmentConcatenator {
  * @experimental
  */
 export class DefaultTokenResolver implements ITokenResolver {
+  /**
+   * Resolves token
```
ansgarm, line 158:
```suggestion
   * Resolves tokens
```

packages/cdktf/package.json
```diff
@@ -71,12 +72,14 @@
       "@typescript-eslint/no-use-before-define": 0,
       "@typescript-eslint/no-empty-interface": 0,
       "@typescript-eslint/explicit-module-boundary-types": 0,
-      "no-sequences": "error"
+      "no-sequences": "error",
+      "jsdoc/require-jsdoc": [2, { "contexts": ["ClassDeclaration"] }]
```
ansgarm, line 76: I'd be more explicit here. Using `0` for turning off rules feels intuitive enough, but using `2` never really felt good to me 😁
```suggestion
      "jsdoc/require-jsdoc": ["error", { "contexts": ["ClassDeclaration"] }]
```

ansgarm (commented):
Love this! Just some small nits here

packages/cdktf/lib/testing/matchers.ts
DanielMSchmidt: The sentence goes on in the next line

packages/cdktf/lib/testing/matchers.ts
```diff
@@ -29,8 +31,13 @@ export function returnMatcherToJest(
   };
 }
 
-// All expected properties are matched and considered equal if
-// There can be more properties in the received object than in the expected object while still returning true
+/**
+ * Compares expected and received. All expected properties are matched and considered equal if
+ * there can be more properties in the received object than in the expected object while still returning true.
```
ansgarm, line 36: Ah, okay but the sentence does not sound right. "…considered equal if there can be more…" 😅

```suggestion
 * Compares expected and received. All expected properties are matched and considered equal even if
 * there are more properties in the received object than in the expected object in which case it will still return true.
```

ansgarm (approved):

packages/cdktf/lib/testing/adapters/__tests__/index.test.ts
ansgarm: nice!

PR #2016, opened 2022-08-10
chore: add vercel config for cdk.tf redirects page

ansgarm:
Migrates the configuration of cdk.tf redirects into our main repository (into a new root level dir called `cdk.tf` which we could e.g. use for a landing page in the future as well).

Example link: https://terraform-cdk-redirects.vercel.app/docs

### DIFF
--- a/cdk.tf/README.md
+++ b/cdk.tf/README.md
@@ -1,3 +1,7 @@
 # cdk.tf
 
 This domain currently only is used for short urls linking to various docs and places and are used in e.g. error messages or across social media.
+
+## Deployment
+
+This service is deployed via Vercel. The Vercel Github integration will build previews for pull requests and the `main` branch will be deployed to production automatically. Refer to Vercel for deployment logs in case an error is posted via the Github integration: https://vercel.com/hashicorp/terraform-cdk-redirects
--- a/cdk.tf/README.md
+++ b/cdk.tf/README.md
@@ -1,0 +1,3 @@
+# cdk.tf
+
+This domain currently only is used for short urls linking to various docs and places and are used in e.g. error messages or across social media.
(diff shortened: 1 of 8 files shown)


cdk.tf/README.md
```diff
@@ -0,0 +1,3 @@
+# cdk.tf
+
+This domain currently only is used for short urls linking to various docs and places and are used in e.g. error messages or across social media.
```
DanielMSchmidt, line 3: Can you expand this please? I'd like to see

- How does it go from here to production?
- What can go wrong? Where do I see if a rollout failed?

cdk.tf/README.md
```diff
@@ -0,0 +1,3 @@
+# cdk.tf
```
DanielMSchmidt, line 1: I wonder if we can make this live anywhere nested, like having a "websites" folder

DanielMSchmidt (changes requested):

cdk.tf/README.md
ansgarm: I'd refrain from having `website` and `websites` 😅 But I'm also not happy with it polluting the root of our repository 😁 However, it just didn't feel right in any of the other directories.
Ideally I'd like to put this into `website` but that directory already is a Vercel root 😅

Do you have any other idea where to put this? I mean, it could be buried as it is not that important to intuitively discover.

cdk.tf/README.md
ansgarm: added a bit more content below. Does that suffice?

cdk.tf/README.md
DanielMSchmidt: Ah I see, I thought we could haven website, then this and the docs under it

cdk.tf/README.md
DanielMSchmidt: Ok, yeah works for me

DanielMSchmidt (approved):

PR #2025, opened 2022-08-15

team-tf-cdk:


ansgarm (approved):

ansgarm:
@DanielMSchmidt you already tested the latest JSII update (which is part of this upgrade) – was it okay to go?
(If yes, merge this PR 😁)

DanielMSchmidt:
@ansgarm We did not get a substantial performance improvement from this change, but the bug with the booleans seems fixed so I'll merge this :) No need to not update from my point of view

PR #2035, opened 2022-08-19
feat: add Go port of TypeScript Google CloudRun example

a-h:


examples/go/google-cloudrun/cdktf.json
```diff
@@ -0,0 +1,13 @@
+{
+  "language": "go",
+  "app": "go run main.go",
+  "codeMakerOutput": "generated",
+  "projectId": "1259b194-7cf9-49c0-aeb2-5e537d3c06da",
```
ansgarm, line 5: We usually remove the project ids from our examples to leave users the choice of whether they'd like to opt into telemetry or not :)

```suggestion
```

examples/go/google-cloudrun/README.md
```diff
@@ -0,0 +1,3 @@
+# google-cloudrun
+
+A port of the TypeScript example at: https://github.com/hashicorp/terraform-cdk/tree/main/examples/typescript/google-cloudrun
```
ansgarm, line 3: I think it would be fine to just copy & paste the README from the other example here.
Saves readers a click 😁

examples/go/google-cloudrun/go.mod
```diff
@@ -0,0 +1,11 @@
+module github.com/hashicorp/terraform-cdk/examples/go/google-cloudrun
+
+go 1.16
+
+require github.com/aws/constructs-go/constructs/v10 v10.1.59
+
+require (
+	github.com/aws/jsii-runtime-go v1.63.1
+	github.com/hashicorp/cdktf-provider-google-go/google v0.0.0
+	github.com/hashicorp/terraform-cdk-go/cdktf v0.12.0
```
ansgarm, line 10: Our other (Go) examples use the repository local version of cdktf and a pre-built version of the provider. This way we can run examples in CI and ensure they still work with the latest changes.

I can adjust this example for you, if you need help getting everything to run this way 🙂

ansgarm:
Thank you for the PR, @a-h 👏

examples/go/google-cloudrun/go.mod
a-h: Thanks. I think it might be better to update the CI process to add the `replace` directive, rather than have the code include it, because with the `replace` directive you can't just download the example code and use it, but I've put it in there in the latest commit.

Just to document the process for anyone else who reads this... all of the examples have this replace directive in the `go.mod` file:

```
replace github.com/hashicorp/terraform-cdk-go/cdktf => ../../../packages/cdktf/dist/go/cdktf
```

Following that through, in the `packages/cdktf` directory I tried to run `npm run package:go` which I assumed would create the `dist` directory.

This produced the `sh: line 1: jsii-pacmak: command not found` error. I installed that with `npm install --global jsii-pacmak`. I installed all of the missing JS packages with `yarn install` then I was able to run `npm run package:go` to create the `dist` directory and be able to run `cdktf synth` again.

examples/go/google-cloudrun/README.md
a-h: Thanks, I've done that. The README wasn't particularly relevant, so I've updated it too.

examples/go/google-cloudrun/go.mod
ansgarm: Yeah, you're right. Having an example that can be used as is would be preferable.

That said, I think we still need to switch to locally built providers as they have different imports and only locally built bindings will contain whatever changes we might have made to the code generation 😇 But the example should still work for users that copy it, even if it uses locally built providers.

Thank you for sharing that! We're using yarn workspaces, so a `yarn install` in the root of the project should've done the trick (and install jsii-pacmak).

ansgarm: Added a follow-up issue to remove the `replace` directive from all our Go examples: https://github.com/hashicorp/terraform-cdk/issues/2041

DanielMSchmidt (approved):

PR #2042, opened 2022-08-23
fix: support provisioners in JSII languages

DanielMSchmidt:


examples/python/provisioner/README.md
```diff
@@ -0,0 +1,25 @@
+# Python-Provisioner
+
+This example shows how one can use provisioners in python.
```
ansgarm, line 3:
```suggestion
This example shows how one can use provisioners in Python.
```

examples/python/provisioner/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+You can install Pipenv by visiting the [website](https://pipenv.pypa.io/en/latest/).
+
+Install project dependencies
+
+```shell
+pipenv install
+```
+
+Generate CDK for Terraform constructs for Terraform provides and modules used in the project.
```
ansgarm, line 21:
```suggestion
Generate CDK for Terraform constructs for Terraform providers and modules used in the project.
```

ansgarm (approved):
Two typos and a question: Do we need to add a small note to our changelog that the names changed? I think a patch release is fine, but documenting dropping the leading `I` feels worth it.

ansgarm (approved):

PR #2044, opened 2022-08-23
chore(lib): deprecate Resource in favor of Construct

### LINKED ISSUE #1273: chore(lib): Deprecate Resource in favour of Construct
### Community Note

### Description

There are practically 2 ways to create abstractions within cdktf right now:

1. [Resource](https://github.com/hashicorp/terraform-cdk/blob/f4dc243658f181aea92aed7519ef63a0f4f778ff/packages/cdktf/lib/resource.ts)
1. [Construct](https://github.com/aws/constructs/blob/cd5aadbaa45cece2697d4f284fee6b885401aecd/src/construct.ts)

`Resource` is just a very thin layer on top fo `Construct` and is not doing anything particularly useful and I don't recall why it got introduced in the first place. Since it's regularily causing confusion, we should deprecated and eventually remove `Resource` in favour of `Construct`.

### Example

The following example shows that both versions are possible and there's no practical difference.

```ts
import { Construct } from "constructs";
import { App, Resource, TerraformStack } from "cdktf";
import { AwsProvider, S3 } from './.gen/providers/aws'

class MyResourceConstruct extends Resource {
  public readonly foo: string;

    super(scope, id);

    this.foo = "bar";

    new S3.S3Bucket(this, "MyBucket");
  }
}

class MyConstruct extends Construct {
  public readonly foo: string;

    super(scope, id);

    this.foo = "bar";

    new S3.S3Bucket(this, "MyBucket");
  }
}


class MyStack extends TerraformStack {
  constructor(scope: Construct, name: string) {
    super(scope, name);

    // define resources here

    new AwsProvider(this, "aws", {
      region: "us-east-1"
    });

    new MyResourceConstruct(this, "my-resource");
    new MyConstruct(this, "my-construct");
  }
}

const app = new App();
new MyStack(app, "foobar");
app.synth();
```

DanielMSchmidt:
Closes #1273


packages/cdktf/lib/resource.ts
```diff
@@ -14,6 +14,7 @@ export interface IResourceConstructor<T extends IResource> {
 
 /**
  * A construct which represents a resource.
+ * @deprecated - Please use Construct instead.
```
ansgarm, line 17: Maybe mention the constructs package as it cannot be imported from cdktf as Resource could?

jsteinich:
We should probably stop using the class internally if it is meant to be deprecated.

ansgarm (approved):

PR #2047, opened 2022-08-24
refactor: port example script to JS

DanielMSchmidt:
This will help us to measure and send metrics


tools/build-example.js
```diff
@@ -0,0 +1,32 @@
...
+function run(command) {
+  // TODO: look into https://www.npmjs.com/package/pidusage to get the Memory usage of the process.
+  const start = performance.now();
+  exec(command, {
+    stdio: "inherit",
+    env: {
+      ...process.env,
+      CI: "true", // Disable spinner even when we have a TTY
+    },
+  });
```
ansgarm, line 16: The old script makes sure to always call the command from the repo root: https://github.com/hashicorp/terraform-cdk/blob/65e7adb207d540298a53be7622f4fd00610f24c6/tools/build-examples.sh#L5
I think we could supply a `cwd` here to do the same?

ansgarm (approved):

PR #2062, opened 2022-08-25
feat(cli): throw an error if a user tries to create a new project with TFC on an old TF version

DanielMSchmidt:


packages/cdktf-cli/bin/cmds/helper/init.ts
```diff
@@ -77,13 +79,21 @@ export async function runInit(argv: Options) {
...
   // Check if template was specified by the user
   let template = "";
   if (argv.template) {
     template = argv.template;
   }
 
+  const tfVersion = await getTerraformVersion();
+  if (tfVersion && semver.lt(tfVersion, "1.1.0") && isRemote) {
+    throw Errors.Usage(
+      `Terraform version ${tfVersion} is not supported for remote configuration. We use the CloudBackend to configure the connection to Terraform Cloud, which is only supported in Terraform 1.1 and higher. Please upgrade to version 1.1.0 or higher or use the local mode by passing --local (in this case you can use the RemoteBackend to work with Terraform Cloud.`
```
ansgarm, line 93:
```suggestion
      `Terraform version ${tfVersion} is not supported for remote configuration. We use the CloudBackend to configure the connection to Terraform Cloud, which is only supported in Terraform 1.1 and higher. Please upgrade to version 1.1.0 or higher or use the local mode by passing --local (afterwards you can use the RemoteBackend to work with Terraform Cloud without upgrading your Terraform version.`
```

packages/cdktf-cli/bin/cmds/helper/init.ts
```diff
@@ -113,7 +123,7 @@ This means that your Terraform state file will be stored locally on disk in a fi
 
     // Check if token is set so we can set up Terraform Cloud workspace
     // only set with the '--local' option is specified the user.
```
ansgarm, line 125: That comment seems vastly outdated now (and incorrect even before that 😅)
```suggestion
    // Set up a Terraform Cloud workspace if the user opted-in
```

packages/cdktf-cli/bin/cmds/helper/terraform-check.ts
```diff
@@ -4,19 +4,13 @@ import { SynthesizedStack } from "../../../lib/synth-stack";
 import { existsSync } from "fs-extra";
 import * as path from "path";
 import { AbortController } from "node-abort-controller"; // polyfill until we update to node 14
+import { logger } from "../../../lib/logging";
 
-const MIN_SUPPORTED_VERSION = "0.13.0";
+const MIN_SUPPORTED_VERSION = "1.0.0";
```
ansgarm, line 9: Where is this used? We should add at least a line to our changelog (about init requiring TF 1.1 now if using TFC).
If there is no change for existing projects, I'd be fine to release this as a patch even. Else we should only include this changed requirement with the next minor (i.e. 0.13).

DanielMSchmidt: We run this as on init, login and whenever a react component executes. I can also remove this, but I think we already require users to be on TF 1.0 or above, don't we?

ansgarm: Yeah, but enforcing this is new.

DanielMSchmidt: Then let's remove this commit (I made a separate one for this line)

ansgarm: I checked the code and it does only warn but not fail, so keeping it would be fine with me

ansgarm: However, we should document that >1.1 is forced for init now, if users use remote state.

ansgarm (approved):

PR #2063, opened 2022-08-25
chore(docs): Check whether links work :) 

laurapacilio:
We need to migrate the links to exclude the terraform.io part --> that will no longer exist with the devdot migration. This PR is where we can test to make sure the new links we merged into the `dev-portal` branch will work!

### DIFF
--- a/cdk.tf/vercel.json
+++ b/cdk.tf/vercel.json
@@ -185,7 +185,7 @@
     },
     {
       "source": "/crash-reporting",
-      "destination": "https://developer.hashicorp.com/cdktf/telemetry#crash-reporting"
+      "destination": "https://www.terraform.io/cdktf/telemetry#crash-reporting"
     }
   ]
 }
--- a/cdk.tf/vercel.json
+++ b/cdk.tf/vercel.json
@@ -182,6 +182,10 @@
     {
       "source": "/cdk-day-2022-hybrid-modules",
       "destination": "https://www.youtube.com/watch?v=7Dt_zJbDmso"
+    },
+    {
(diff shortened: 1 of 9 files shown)


laurapacilio:
hm @ansgarm it's interesting that the Vercel deployment failed - i wonder if that's expected? I don't really know what that check is doing but maybe it makes sense since we changed links to something that don't actually exist yet? Do you have thoughts?

cdk.tf/vercel.json
```diff
@@ -180,6 +180,9 @@
       "destination": "https://registry.terraform.io/browse/providers"
     },
     {
+      "source": "/crash-reporting",
+      "destination": "https://developer.hashicorp.com/cdktf/telemetry#crash-reporting"
+     },
```
ansgarm, line 185: JSON seems to be invalid 😅 That's probably why Vercel failed.

```suggestion
     },
     {
```

cdk.tf/vercel.json
```diff
@@ -180,6 +180,10 @@
       "destination": "https://registry.terraform.io/browse/providers"
     },
     {
+      "source": "/crash-reporting",
+      "destination": "https://developer.hashicorp.com/cdktf/telemetry#crash-reporting"
```
BRKalow, line 184:
```suggestion
      "destination": "https://terraform.io/cdktf/telemetry#crash-reporting"
```

ansgarm (approved):

ansgarm:
I just rebased the branch (it had a few conflicts), if everything still looks correct, we can merge this @laurapacilio 💯

PR #2068, opened 2022-08-26
fix: let create changelog script use commit hashes from PRs instead of finding merge commit messages that may not always exist

ansgarm:


DanielMSchmidt (approved):

jsteinich:
The same tooling that we use for handling pre-release versioning could also be used in the release process. I tested it some when adding, but haven't fully confirmed the changelog side since I'm not involved in the release process.
Take a look [here](https://github.com/conventional-changelog/standard-version) for some more information on it.

PR #2069, opened 2022-08-26
chore(examples): clean up examples

ansgarm:


ansgarm:
This is how it looks like if it fails now:

image

DanielMSchmidt (approved):

Issue #2097, opened 2022-09-06
Support building with go 1.19


mutahhir:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Expected Behavior
Running `yarn build` succeeds.

### Actual Behavior
Running `yarn build` fails with the following errors:

<img width="1140" alt="image" src="https://user-images.githubusercontent.com/573531/188669991-6ba90745-bfce-41b8-b40b-afc89142469d.png">

### Steps to Reproduce

- Install Go 1.19
- run `yarn build`

### References

- https://github.com/golang/go/commit/680caf15355057ca84857a2a291b6f5c44e73329

PR #2098, opened 2022-09-06
chore: Support building with Go 1.19

mutahhir:
`@cdktf/hcl2json` doesn't work when built with Go 1.19, as [Go changed wasm_exec to expect polyfills to be available by environment](https://github.com/golang/go/commit/680caf15355057ca84857a2a291b6f5c44e73329).

This PR creates a new wrapper javascript file that provides the required polyfills. In theory, this changes should not affect Go versions prior to 1.19. This is because the older version of `wasm_exec.js` was doing the polyfilling itself and it no longer would need to do it.

Closes: #2097 

> **Note**
> 
> I didn't know about `globalThis` before, it's nice: 
>
>> Historically, accessing the global object has required different syntax in different JavaScript environments. On the web you can use [window](https://developer.mozilla.org/en-US/docs/Web/API/Window/window), [self](https://developer.mozilla.org/en-US/docs/Web/API/Window/self), or [frames](https://developer.mozilla.org/en-US/docs/Web/API/Window/frames) - but in [Web Workers](https://developer.mozilla.org/en-US/docs/Web/API/Worker) only self will work. In Node.js none of these work, and you must instead use global. The this keyword could be used inside functions running in non–strict mode, but this will be undefined in Modules and inside functions running in strict mode. You can also use Function('return this')(), but environments that disable [eval()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval), like [CSP](https://developer.mozilla.org/en-US/docs/Glossary/CSP) in browsers, prevent use of [Function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function) in this way.
>>
>> The globalThis property provides a standard way of accessing the global this value (and hence the global object itself) across environments. Unlike similar properties such as window and self, it's guaranteed to work in window and non-window contexts. In this way, you can access the global object in a consistent manner without having to know which environment the code is being run in. To help you remember the name, just remember that in global scope the this value is globalThis.
> -- [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis)


packages/@cdktf/hcl2json/wasm/.npmignore
```diff
@@ -1,2 +1,3 @@
 !wasm_exec.js
+!wasm_exec_node.js
```
ansgarm, line 2:
```suggestion
!bridge_wasm_exec.js
```

ansgarm (commented):
Great work 🎉 Only one small fix.

packages/@cdktf/hcl2json/wasm/.npmignore
mutahhir: Thank you! Good catch!

ansgarm (approved):

PR #2102, opened 2022-09-07
Remove empty code block

cwiggs:
This commit removes an empty code block that was at the bottom of the file.


ansgarm (approved):
Thanks! Good catch @cwiggs!

PR #2104, opened 2022-09-08
chore: Update diagram with new provider count

danbarr:
Minor update to the main CDKTF graphic to bump "1000+ providers" to "2000+ providers"


ansgarm (approved):
💯(×20)

Issue #734, opened 2021-05-26
Support terraform cli options


karthikns16:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description
 
I am working on setting up OKTA application using OKTA provider https://registry.terraform.io/providers/oktadeveloper/okta/latest/docs

This provider seems to have a bug https://github.com/okta/terraform-provider-okta/issues/162 and are unable to handle parallel request sent by terraform. We have to use ```-parallelism=1```. cdktf deploy doesn't seem support terraform cli options

Hence instead of using cdktf deploy, I have to resort to using ```terraform apply --auto-approve -parallelism=1``` after cdktf synth

Would be great if terraform cli options honored by cdktf cli

RomeroGaliza:
I have a very similar use case with SignalFX provider, where it turns out that `-parallelism=1` can be used workaround.

jsteinich:
It might be interesting to explore allowing something like `cdktf deploy stack -- -parallelism=1 -target=mymodule`.
I've worked with a few tools that have this idea of a separator after which all options are directly passed to underlying tools. In this example an empty `--` is treated as the separator, but could come up with something else.

mutahhir:
I think this will be solved between #2107 and the move to Terraform CLI

DanielMSchmidt:
Closed by #2468

jsteinich:
While a few cli flags are supported, there still isn't general support for them.

PR #2107, opened 2022-09-13
feat(cli): Add support for terraform's parallelism flag 

### LINKED ISSUE #1715: Issue with @cdktf/provider-postgresql Grant (Add terraform-parallelism flag)
### Community Note

### cdktf & Language Versions
"@cdktf/provider-aws": "^7.0.13"
"@cdktf/provider-postgresql": "^0.4.12"
"@cdktf/provider-random": "^0.2.12"
"cdktf": "^0.10"

"ts-node": "^10.7.0",
"typescript": "^4.6.2"

### Affected Resource(s)

PostgresqlProvider Grant Resource

### Debug Output

`Error: could not execute revoke query: pq: tuple concurrently updated`

### Expected Behavior

Successfully created each role grant for each role.

### Actual Behavior

Concurrency error in spite of trying both the `--parallelism=0` and `--parallelism=1` flags

### Steps to Reproduce

```
roles.forEach((role) => {

     ...doStuff

      role.grants.forEach((grant) => {
        new Grant(this, `${this.deploymentPrefix}-${role.name}-${grant.objectType}-grant`, {
          role: role.name,
          database: role.database,
          objectType: grant.objectType,
          privileges: grant.privileges,
        });
      });

})
```

### Important Factoids

I've done some research and tried the concurrency flag which doesn't seem to be fixing anything. My assumption is that either something is wrong with the flag implementation in CDKTF or the Grant object itself.

mutahhir:
Fixes #1715 

This PR adds the `-terraform-parallelism` flag to the CLI. This flag is only available for `deploy` and `destroy`. One note here is that due the nature of `deploy`, the parallelism flag will be used for both `plan` and `apply` within the Terraform CLI during the `deploy` operation.

One thing to note: The `parallelism` flag is not supported by Terraform Cloud, so setting that for remote backends will be ignored.

Also, now now have two `parallelism` flags. One is `--parallelism` and the other is `--terraform-parallelism`. The former only affects how `cdktf` processes multiple stacks, while the latter is passed on to the Terraform CLI and changes how many concurrent operations Terraform starts as it walks the graph.

### DIFF
--- a/packages/cdktf-cli/bin/cmds/deploy.ts
+++ b/packages/cdktf-cli/bin/cmds/deploy.ts
@@ -56,6 +56,8 @@ class Command extends BaseCommand {
         default: false,
       })
       .option("parallelism", {
+        // Note: This parallelism doesn't affect the underlying Terraform traversal parallelism
+        // That is done by `terraform-parallelism`
         type: "number",
         required: false,
         desc: "Number of concurrent CDKTF stacks to run. Defaults to infinity, denoted by -1",
@@ -70,7 +72,8 @@ class Command extends BaseCommand {
       .option("terraform-parallelism", {
         type: "number",
         required: false,
-        desc: "Customize number of parallel graph traversals by Terraform. By default, the parallelism flag is not forwarded to Terraform.",
+        desc: "Customize number of parallel graph traversals by Terraform. By default, the parallelism flag is not forwarded to Terraform. Note: This flag is not supported by remote / cloud backend",
+        // Setting value to negative will prevent it from being forwarded to terraform as an argument
         default: -1,
       })
       .showHelpOnFail(true);
--- a/packages/cdktf-cli/bin/cmds/deploy.ts
+++ b/packages/cdktf-cli/bin/cmds/deploy.ts
@@ -72,7 +72,7 @@ class Command extends BaseCommand {
       .option("terraform-parallelism", {
         type: "number",
         required: false,
-        desc: "Customize number of parallel graph traversals by Terraform. By default, the parallelism flag is not forwarded to Terraform. Note: This flag is not supported by remote / cloud backend",
+        desc: "Forwards value as the `-parallelism` flag to Terraform. By default, the this flag is not forwarded to Terraform. Note: This flag is not supported by remote / cloud backend",
         // Setting value to negative will prevent it from being forwarded to terraform as an argument
         default: -1,
       })
--- a/packages/cdktf-cli/bin/cmds/deploy.ts
+++ b/packages/cdktf-cli/bin/cmds/deploy.ts
@@ -67,6 +67,12 @@ class Command extends BaseCommand {
         boolean: true,
         desc: 'Select the "refresh only" planning mode, which checks whether remote objects still match the outcome of the most recent Terraform apply but does not propose any actions to undo any changes made outside of Terraform.',
       })
+      .option("terraform-parallelism", {
+        type: "number",
+        required: false,
+        desc: "Customize number of parallel graph traversals by Terraform. By default, the parallelism flag is not forwarded to Terraform.",
+        default: -1,
+      })
       .showHelpOnFail(true);
 
   public async handleCommand(argv: any) {
--- a/packages/cdktf-cli/bin/cmds/destroy.ts
+++ b/packages/cdktf-cli/bin/cmds/destroy.ts
@@ -51,7 +51,7 @@ class Command extends BaseCommand {
       .option("terraform-parallelism", {
         type: "number",
         required: false,
-        desc: "Customize number of parallel graph traversals by Terraform. By default, the parallelism flag is not forwarded to Terraform. Note: This flag is not supported by remote / cloud backend",
+        desc: "Forwards value as the `-parallelism` flag to Terraform. By default, the this flag is not forwarded to Terraform. Note: This flag is not supported by remote / cloud backend",
         // Setting value to negative will prevent it from being forwarded to terraform as an argument
         default: -1,
       })
--- a/packages/cdktf-cli/bin/cmds/destroy.ts
+++ b/packages/cdktf-cli/bin/cmds/destroy.ts
@@ -48,6 +48,12 @@ class Command extends BaseCommand {
         desc: "Number of concurrent CDKTF stacks to run. Defaults to infinity, denoted by -1",
         default: -1,
       })
+      .option("terraform-parallelism", {
+        type: "number",
+        required: false,
+        desc: "Customize number of parallel graph traversals by Terraform. By default, the parallelism flag is not forwarded to Terraform.",
+        default: -1,
+      })
       .showHelpOnFail(true);
 
   public async handleCommand(argv: any) {
--- a/packages/cdktf-cli/bin/cmds/destroy.ts
+++ b/packages/cdktf-cli/bin/cmds/destroy.ts
@@ -51,7 +51,8 @@ class Command extends BaseCommand {
       .option("terraform-parallelism", {
         type: "number",
         required: false,
-        desc: "Customize number of parallel graph traversals by Terraform. By default, the parallelism flag is not forwarded to Terraform.",
+        desc: "Customize number of parallel graph traversals by Terraform. By default, the parallelism flag is not forwarded to Terraform. Note: This flag is not supported by remote / cloud backend",
+        // Setting value to negative will prevent it from being forwarded to terraform as an argument
         default: -1,
       })
       .showHelpOnFail(true);
--- a/packages/cdktf-cli/lib/cdktf-project.ts
+++ b/packages/cdktf-cli/lib/cdktf-project.ts
@@ -279,17 +279,21 @@ export type MultipleStackOptions = {
   stackNames?: string[];
 };
 
-export type ExecutionOptions = MultipleStackOptions & {
+export type AutoApproveOptions = {
   autoApprove?: boolean;
-  ignoreMissingStackDependencies?: boolean;
-  parallelism?: number;
-  refreshOnly?: boolean;
-  terraformParallelism?: number;
 };
 
 export type DiffOptions = SingleStackOptions & {
   refreshOnly?: boolean;
 };
+
+export type MutationOptions = MultipleStackOptions &
+  AutoApproveOptions & {
+    refreshOnly?: boolean;
+    ignoreMissingStackDependencies?: boolean;
+    parallelism?: number;
+    terraformParallelism?: number;
+  };
 
 type LogMessage = {
   stackName: string;
@@ -453,7 +457,7 @@ export class CdktfProject {
 
   public getStackExecutor(
     stack: SynthesizedStack,
-    opts: ExecutionOptions = {}
+    opts: AutoApproveOptions = {}
   ) {
     const enhanceLogMessage = createEnhanceLogMessage(stack);
     const onLog = this.bufferWhileWaitingForApproval(this.onLog);
@@ -520,16 +524,15 @@ export class CdktfProject {
   private async execute(
     method: "deploy" | "destroy",
     next: () => Promise<CdktfStack | undefined>,
-    parallelism: number,
-    refreshOnly?: boolean,
-    terraformParallelism?: number
+    opts: MutationOptions
   ) {
     // We only support refresh only on deploy, a bit of a leaky abstraction here
-    if (refreshOnly && method !== "deploy") {
+    if (opts.refreshOnly && method !== "deploy") {
       throw Errors.Internal(`Refresh only is only supported on deploy`);
     }
 
-    const maxParallelRuns = parallelism === -1 ? Infinity : parallelism;
+    const maxParallelRuns =
+      !opts.parallelism || opts.parallelism < 0 ? Infinity : opts.parallelism;
     while (this.stacksToRun.filter((stack) => stack.isPending).length > 0) {
       const runningStacks = this.stacksToRun.filter((stack) => stack.isRunning);
       if (runningStacks.length >= maxParallelRuns) {
@@ -544,8 +547,11 @@ export class CdktfProject {
           break;
         }
         method === "deploy"
-          ? nextRunningExecutor.deploy(refreshOnly, terraformParallelism)
-          : nextRunningExecutor.destroy(terraformParallelism);
+          ? nextRunningExecutor.deploy(
+              opts.refreshOnly,
+              opts.terraformParallelism
+            )
+          : nextRunningExecutor.destroy(opts.terraformParallelism);
       } catch (e) {
         // await next() threw an error because a stack failed to apply/destroy
         // wait for all other currently running stacks to complete before propagating that error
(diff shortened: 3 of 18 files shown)


packages/cdktf-cli/lib/cdktf-project.ts
```diff
@@ -279,17 +279,22 @@ export type MultipleStackOptions = {
...
-  ignoreMissingStackDependencies?: boolean;
-  parallelism?: number;
-  refreshOnly?: boolean;
 };
 
 export type DiffOptions = SingleStackOptions & {
   refreshOnly?: boolean;
 };
 
+export type MutationOptions = MultipleStackOptions &
```
mutahhir, line 290: I'm not in love with the name, and after all the modifications, I wonder if the name change is even required. I chose mutation because it was going to be consumed by both `deploy` and `destroy`, which are both operations that mutate state, as opposed to Synth.

packages/cdktf-cli/lib/cdktf-project.ts
```diff
@@ -519,51 +524,47 @@ export class CdktfProject {
...
-            openStacks
-          );
-          await Promise.allSettled(
-            openStacks.map((ex) => ex.currentWorkPromise)
-          );
-          logger.debug(
-            "Done waiting for still running stacks. All pending work finished"
-          );
-          throw e;
+        continue;
```
mutahhir, line 540: Flattening the if-else a bit. As far as I can see, it shouldn't have any impact, but maybe I've missed something

packages/cdktf-cli/test/lib/cdktf-project.test.ts
```diff
@@ -6,13 +6,24 @@ import * as os from "os";
...
 import { getMultipleStacks } from "../../lib/cdktf-project";
+import { exec } from "cdktf-cli/lib/util";
 
 function eventNames(events: any[]) {
   return events
     .map((event) => event.type)
     .filter((name) => !name.includes("update"));
 }
 
+jest.mock("cdktf-cli/lib/util", () => {
```
mutahhir, line 17: It's been a while since I've done this, so maybe there's a better way to do this?

ansgarm: Seems to be the way to go with Jest 😁 I don't know a better one 🤷‍♂️

packages/cdktf-cli/test/lib/cdktf-project.test.ts
```diff
@@ -211,6 +230,60 @@ describe("CdktfProject", () => {
       ]);
       return expect(eventTypes.includes("waiting for approval")).toBeFalsy();
     });
+
+    it("runs without terraform parallelism", async () => {
```
ansgarm, line 234: The test seems to run with that flag, update the test name? Or am I missing something?

ansgarm:
```suggestion
    it("passes the terraform parallelism flag to terraform", async () => {
```

ansgarm (commented):
Nice work, @mutahhir 🎉 Good job cleaning up some old code as well 👍
Only a couple small nits / questions 🙂

packages/cdktf-cli/lib/cdktf-project.ts
ansgarm: If we go with that, we should rename the `execute` method as well, as I think that one was the original reason for it to be named `ExecutionOptions`. I personally would also be fine with a verbose `DeployDestroyOptions` 😁 But Mutation seems like a valid umbrella term for both of those.

What do you think, @DanielMSchmidt? You came up with this name originally, is that wording maybe already used in the Terraform world for apply/destroy?

packages/cdktf-cli/bin/cmds/deploy.ts
```diff
@@ -67,6 +69,13 @@ class Command extends BaseCommand {
         boolean: true,
         desc: 'Select the "refresh only" planning mode, which checks whether remote objects still match the outcome of the most recent Terraform apply but does not propose any actions to undo any changes made outside of Terraform.',
       })
+      .option("terraform-parallelism", {
```
DanielMSchmidt, line 72: Was there a reason to decide for the CLI-flag approach versus the construct approach? To me, the reasoning of per-stack configuration made sense

packages/cdktf-cli/bin/cmds/destroy.ts
```diff
@@ -48,6 +48,13 @@ class Command extends BaseCommand {
         desc: "Number of concurrent CDKTF stacks to run. Defaults to infinity, denoted by -1",
         default: -1,
       })
+      .option("terraform-parallelism", {
```
DanielMSchmidt, line 51: Wouldn't we also need this for plan?

packages/cdktf-cli/bin/cmds/deploy.ts
```diff
@@ -56,6 +56,8 @@ class Command extends BaseCommand {
         default: false,
       })
       .option("parallelism", {
+        // Note: This parallelism doesn't affect the underlying Terraform traversal parallelism
```
DanielMSchmidt, line 59: We need to add this to the CLI documentation as well, alongside with the explanation in the PR description

packages/cdktf-cli/bin/cmds/deploy.ts
ansgarm: My opinion on this:

I like the general idea of putting such flags into configuration and we should think about this general idea for other flags as well.
The downside I see to such an approach is that this can't be added to the `cdk.tf.json` that we output and therefore would stop users from using the Terraform CLI in the output directory without knowing that they might need to set such a setting on a certain stack.

For this particular flag, we should confirm its common use-case. My guess is that it is usually used on large stacks when running into limits with some cloud provider APIs – to me personally this would not warrant putting it in the stack (but rather might be something folks would put into their CI).
A different angle: If this is something that could be part of the config, of e.g. a TF provider setting, why is it a flag in Terraform and not a setting in HCL itself? Maybe digging into that direction helps us to uncover more here.

mutahhir: My bad, I started with the construct approach, and then switched mid-way. I was planning on explaining it, but lost track. Here's how I was thinking about it:

The reason why the construct approach appealed to me in the start was due to multi-stack deploys. From my limited understanding, parallelism would be used in advanced cases, which leads me to think that it's not something that would be needed across multiple stacks.

However, when I looked at the CLI design, it seemed like we always wanted a certain stack to be specified. Initially, as I was starting out by investigating the CLI, it looked like the deploy command would be deploying a single stack, and nothing else. Then, having a flag would look better, and easier to use.

However, looking at the code, it seems like we would also start planning dependencies of the stack we're deploying. Is that correct? If so, indeed, having a construct is probably better because then we might have a single stack which is a dependency that requires parallelism to be set. At first I thought this would be a blocker for the flag approach, since parallelism would be set for all the dependant stacks, but after some thinking, it seemed like the customer could always deploy the dependant stack with the flag set separately.

In the end, however, the main reason for going back to the CLI flag approach was 'where would the customer try to look for it', and it seems that for people who are accustomed to Terraform, they would look at the CLI.

Sorry for the long-winded comment, but that's my story on how I thought about it and why I ended up with this approach.

DanielMSchmidt: > However, when I looked at the CLI design, it seemed like we always wanted a certain stack to be specified

Not really, we let users pass a stack selector and use minimatch glob syntax to find all stacks relevant, build a dependency graph and deploy as many in parallel as possible if not restricted. `cdktf deploy '*'` deploys all stacks specified.

> In the end, however, the main reason for going back to the CLI flag approach was 'where would the customer try to look for it', and it seems that for people who are accustomed to Terraform, they would look at the CLI.

I can see that point and it's probably okay for now

> My guess is that it is usually used on large stacks when running into limits with some cloud provider APIs – to me personally this would not warrant putting it in the stack (but rather might be something folks would put into their CI).

From my point of view you contradict yourself here a bit: the flag is used for some cloud providers and large stacks, so if you have a stack with another cloud provider that is not effected that you want to deploy through the same deploy command you now have no way to differentiate and every stacks "suffers" from limited parallelism. 

I'm not saying we have to do this stack-wise, but we discussed this offline to be a construct solution so I was curious on why we drifted away from that approach.

ansgarm: I was not part of that discussion, unfortunately 😅 I can see the reasoning with different stacks, though 👍

packages/cdktf-cli/bin/cmds/deploy.ts
mutahhir: Quick questions:
- What's the process of updating the CLI documentation
- Which PR description should we update? This one, or the one that added the flag?

packages/cdktf-cli/bin/cmds/deploy.ts
DanielMSchmidt: > I was not part of that discussion

Yeah it was just an ad-hoc one while pairing, let's all make sure to document stuff like this in a public channel :) 

> I didn't see a lot of possible flags that we would like to expose in the potential construct. What do you think is another good potential flag?

I would agree, seems to be like the first thing that falls into this category

packages/cdktf-cli/bin/cmds/deploy.ts
DanielMSchmidt: > What's the process of updating the CLI documentation

This is the file you will need to change within this PR: https://github.com/hashicorp/terraform-cdk/blob/main/website/docs/cdktf/cli-reference/commands.mdx

> Which PR description should we update? This one, or the one that added the flag?

I meant adding the context within this PR description to the explanation you'll add to the documentation

packages/cdktf-cli/test/lib/cdktf-project.test.ts
mutahhir: Ah yeah, sorry, the wording here is confusing. I just used it because without parallelism would mean passing it a value of 1. By default parallelism is on (with a default value of 10). I like yours much better.

ansgarm: It was actually yours (from the destroy test further down) 😁

website/docs/cdktf/cli-reference/commands.mdx
```diff
@@ -130,27 +130,30 @@ $ cdktf deploy --help
...
+      --ignore-missing-stack-dependencies       Don't check if all stacks specified in the command have their dependencies included as well                                                                                                                              [boolean] [default: false]
+      --parallelism                             Number of concurrent CDKTF stacks to run. Defaults to infinity, denoted by -1                                                                                                                                                [number] [default: -1]
+      --refresh-only                            Select the "refresh only" planning mode, which checks whether remote objects still match the outcome of the most recent Terraform apply but does not propose any actions to undo any changes made outside of Terraform.  [boolean] [default: false]
+      --terraform-parallelism                   Customize number of parallel graph traversals by Terraform. By default, the parallelism flag is not forwarded to Terraform. Note: This flag is not supported by remote / cloud backend                                       [number] [default: -1]
```
mutahhir, line 152: @laurapacilio I've added a new flag with some documentation, followed by a warning to differentiate between `parallelism` and the new `terraform-parallelism` flag. Can I ask for you to take a look over this change please?

packages/cdktf-cli/bin/cmds/deploy.ts
mutahhir: > Personally I'm in favor of a more general solution to arbitrary Terraform cli arguments. See #734

I like the `--` separator idea, but I have two thoughts:

- Having two `parallelism` flags mean different things based on position is not a great experience. The current way seems clearer to me.
- Do you have a proposal for another TF CLI flag that we should try the `--` separator out with?

website/docs/cdktf/cli-reference/commands.mdx
laurapacilio: yes! I will put it on my calendar to look at this week! I just need a few days because everything is tight schedule-wise before HashiConf :)

website/docs/cdktf/cli-reference/commands.mdx
```diff
@@ -130,27 +130,30 @@ $ cdktf deploy --help
...
+  -h, --help                                    Show help                                                                                                                                                                                                                                 [boolean]
+```
+
+~> **Note:** The `parallelism` flag has a different behavior than the [terraform parallelism flag](https://www.terraform.io/cli/commands/apply#parallelism-n). To set the custom terraform parallelism flag, please use the `--terraform-parallelism` flag instead.
```
laurapacilio, line 156: A couple of edits for style. But one thing I actually want to recommend is --> Can we tell users how the behavior is different so they can decide which flag they want to use? We're noting that it IS different, which is great, but if the user isn't super familiar they might need help understanding the best path for them

```suggestion
-> **Note:** The `parallelism` flag has a different behavior than the [Terraform `parallelism` flag](https://www.terraform.io/cli/commands/apply#parallelism-n). Use `--terraform-parallelism` to set the custom terraform parallelism flag.
```

website/docs/cdktf/cli-reference/commands.mdx
laurapacilio, line 152: hmm is there a way that we can explain what we mean by "parallel graph traversals"? I think this might tie into my comment below a bit, but for someone who isn't super familiar with Terraform, we may want to explain a bit more in plain language.

packages/cdktf-cli/bin/cmds/deploy.ts
jsteinich: I agree that two arguments of the same name only differing by position isn't a great experience; however, having a special case for an argument isn't great either.
I could see supporting `terraform` prefix generically for any Terraform CLI argument, though just using a separator seems easier to use.
Maybe `parallelism` is more common than I realize, but I've personally never needed to use it so I'm hesitant to give it special consideration. Does the Terraform core team perhaps have data on argument usage?

One argument that I've run into needing is `target`.

mutahhir: Yeah, I agree, both aren't great ideas, and I too haven't had the need for using parallelism either. 

`target`, is indeed an interesting flag, but I wonder if we'd want to incorporate it within the cdktf cli as a first-class flag instead of just forwarding it to Terraform instead.

website/docs/cdktf/cli-reference/commands.mdx
mutahhir: Yeah. I totally get what you're saying here. I mostly tried to summarise [this documentation](https://developer.hashicorp.com/terraform/internals/graph#walking-the-graph), but overall this is hard to explain.

A few options could be:
- Say something like: `Forwarded to the terraform cli's '-parallelism' flag when set, read up more on the terraform cli documentation` and not go into the graph traversals at all?
- Keep the CLI output short, and maybe as-is, but add a separate section about parallelism underneath usage in this doc?

packages/cdktf-cli/lib/cdktf-project.ts
DanielMSchmidt: I'm fine with any option here, naming things is hard and mutation fits quite ok. My thought was execution but for no deeper reason than I needed a term that fits these two

DanielMSchmidt (approved):

PR #2111, opened 2022-09-13
feat(cli): support generating provider bindings in parallel

DanielMSchmidt:
This should speed up cdktf get for all languages except TS 🎉

### DIFF
--- a/packages/@cdktf/provider-generator/lib/get/constructs-maker.ts
+++ b/packages/@cdktf/provider-generator/lib/get/constructs-maker.ts
@@ -280,6 +280,77 @@ export class ConstructsMaker {
     return filePath;
   }
 
+  private async generateJsiiLanguage(target: ConstructsMakerTarget) {
+    // these are the module dependencies we compile against
+    const deps = ["@types/node", "constructs", "cdktf"];
+    const opts: srcmak.Options = {
+      entrypoint: target.fileName,
+      deps: deps.map((dep) =>
+        path.dirname(require.resolve(`${dep}/package.json`))
+      ),
+      moduleKey: target.moduleKey,
+    };
+
+    // used for testing.
+    if (this.options.outputJsii) {
+      opts.jsii = { path: this.options.outputJsii };
+    }
+
+    if (this.isPythonTarget) {
+      opts.python = {
+        outdir: this.codeMakerOutdir,
+        moduleName: target.srcMakName,
+      };
+    }
+
+    if (this.isJavaTarget) {
+      opts.java = {
+        outdir: ".", // generated java files aren't packaged, so just include directly in app
+        package: `imports.${target.srcMakName}`,
+      };
+    }
+
+    if (this.isCsharpTarget) {
+      opts.csharp = {
+        outdir: this.codeMakerOutdir,
+        namespace: target.srcMakName,
+      };
+    }
+
+    if (this.isGoTarget) {
+      // TODO: check if needed for modules somehow
+      // const targetType = target.isProvider ? 'provider' : 'module';
+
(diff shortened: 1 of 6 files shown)


packages/@cdktf/provider-generator/lib/get/constructs-maker.ts
```diff
@@ -294,79 +366,24 @@ export class ConstructsMaker {
...
-          // TODO: check if needed for modules somehow
-          // const targetType = target.isProvider ? 'provider' : 'module';
-
-          // jsii-srcmac will produce a folder inside this dir named after "packageName"
-          // so this results in e.g. .gen/hashicorp/random
-          const outdir = path.join(
-            this.codeMakerOutdir,
-            target.namespace ?? ""
+      const numberOfWorkers =
+        this.options.jsiiParallelism === -1
```
ansgarm, line 370: Maybe change this (or instead use `Math.max()` for the value below in line 372) to ensure passing `-2` or `0` does not cause an endless loop or crash?
```suggestion
        this.options.jsiiParallelism <= 0 || typeof this.options.jsiiParallelism === "undefined"
```

packages/@cdktf/provider-generator/lib/get/constructs-maker.ts
```diff
@@ -294,79 +366,24 @@ export class ConstructsMaker {
...
+          ? this.targets.length
+          : this.options.jsiiParallelism;
+
+      const work = [...this.targets];
+      const workers = new Array(numberOfWorkers).fill(async () => {
+        let target: ConstructsMakerTarget | undefined;
+        while ((target = work.pop())) {
```
ansgarm, line 377: I really like this pattern! Even if it felt a bit weird at first 😁

packages/@cdktf/provider-generator/lib/get/constructs-maker.ts
ansgarm: Oh, just noticed: passing `undefined` would create an array with the length `undefined`, wouldn't it? (before my suggested change)

packages/cdktf-cli/bin/cmds/handlers.ts
```diff
@@ -418,6 +423,10 @@ export async function providerAdd(argv: any) {
     console.log(
       "Local providers have been updated. Running cdktf get to update..."
     );
-    await get({ language: language, output: config.codeMakerOutput });
+    await get({
+      language: language,
+      output: config.codeMakerOutput,
+      parallelism: -1,
```
ansgarm, line 429: Users that have not enough memory would not be able to use the `provider add` command anymore if they can't set this option as get recreates all provider bindings.

website/docs/cdktf/cli-reference/commands.mdx
```diff
@@ -325,6 +329,12 @@ $ cat cdktf.json
 $ cdktf get
 ```
 
+If you run into issues generating a lot of bindings you can use the `--parallelism` option to limit the number of bindings generated in parallel.
+
+```bash
+$ cdktf get --parallelism 1
+```
+
```
ansgarm, line 337: Good catch to add this here 👍

packages/cdktf-cli/bin/cmds/handlers.ts
DanielMSchmidt: Then let's default to 1 here 👍

packages/@cdktf/provider-generator/lib/get/constructs-maker.ts
ansgarm: Nice! Looks good

ansgarm (approved):

PR #2117, opened 2022-09-14
fix: Change url of `pipenv` in log

kination:
Hello
Currently, following message are shown if `pipenv` is not install, to use python

```
$ cdktf init --template python --local
Note: By supplying '--local' option you have chosen local storage mode for storing the state of your stack.
This means that your Terraform state file will be stored locally on disk in a file 'terraform.<STACK NAME>.tfstate' in the root of your project.
...
Unable to find "pipenv". Install from https://pipenv.kennethreitz.org
```
But I couldn't access to `https://pipenv.kennethreitz.org`(is there someone who can?). I think `https://pipenv.pypa.io/` is the correct one. Could you check?

### DIFF
--- a/packages/cdktf-cli/templates/python/.hooks.sscaff.js
+++ b/packages/cdktf-cli/templates/python/.hooks.sscaff.js
@@ -12,7 +12,7 @@ exports.pre = () => {
       execSync('which pipenv')
     }
   } catch {
-    console.error(`Unable to find "pipenv". Install from https://pipenv.kennethreitz.org`)
+    console.error(`Unable to find "pipenv". Follow the instruction from https://pipenv.pypa.io/`)
     process.exit(1);
   }
 };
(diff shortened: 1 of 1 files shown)


packages/cdktf-cli/templates/python/.hooks.sscaff.js
```diff
@@ -12,7 +12,7 @@ exports.pre = () => {
       execSync('which pipenv')
     }
   } catch {
-    console.error(`Unable to find "pipenv". Install from https://pipenv.kennethreitz.org`)
+    console.error(`Unable to find "pipenv". Follow the instruction from https://pipenv.pypa.io/`)
```
ansgarm, line 15:
```suggestion
    console.error(`Unable to find "pipenv". Follow the instructions from https://pipenv.pypa.io/`)
```

ansgarm (approved):
Good catch, @kination! That was probably a quite old link there..

PR #2118, opened 2022-09-16
fix(cli): Update require package in template

kination:
Hello,
When we generate `python-pip` project via template, requirements.txt only includes `pytest` though it also needs `cdktf` package. Looks like this is being overwritten.

Please take a look for this change 🙏


DanielMSchmidt (approved):

DanielMSchmidt:
@kination Thanks for your help! Could you make your commit message match the PR title? Otherwise our release tooling will have problems picking this PR up

kination:
@DanielMSchmidt updated. Could you check?

kination:
@ansgarm looks like it is synced with main branch. Could you check?

Issue #1132, opened 2021-10-11
Show warning when usage of `npm link` causes `instanceof TerraformElement` to fail


ansgarm:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

After extracting commonly used logic (which instantiates constructs) into a separate npm package, `npm link` is commonly used in local development to test the library with an cdktf project.

However the way `npm link` works breaks the `instanceof` [check](https://github.com/hashicorp/terraform-cdk/blob/0732e9f5bbf2e85f7eb2cd131855fe6bf0e8b641/packages/cdktf/lib/terraform-stack.ts#L118-L121) in `TerraformStack`. It creates a symlink from e.g. `<cdktf-project>/node_modules/<custom-library>` pointing to the custom library directory. And as the custom library also has the `cdktf` package installed, that one is instead used for resolving dependencies to `cdktf.TerraformStack` (for constructs created in the custom library code). This causes the check in `TerraformStack.allocateLogicalId` to fail and produces the following error: `TypeError: node.tryGetContext is not a function`.

Instead of this hard to debug error, we could e.g. probe the passed argument and check whether it probably is in fact a `TerraformStack` and print a more useful warning message (e.g. recommending `yalc` – see workaround section).

We could also get rid of `instanceof` for the check but that could only defer the underlying issue to a later (possibly even harder to debug) place in the code.

### Workaround for the issue caused by `npm link`
Using [yalc](https://github.com/wclr/yalc) to link packages for local development solves the issue as it works differently.

### References

* https://github.com/hashicorp/terraform-cdk/blob/0732e9f5bbf2e85f7eb2cd131855fe6bf0e8b641/packages/cdktf/lib/terraform-stack.ts#L118-L121

DanielMSchmidt:
We should check if using symbols solves this problem, e.g. like we do in https://github.com/hashicorp/terraform-cdk/blob/main/packages/cdktf/lib/terraform-stack.ts#L43-L45

cprice404:
[...]

PR #2125, opened 2022-09-16
chore(lib)!: Replace all instanceof checks with Symbols and add lint rule against accidentally adding new ones

ansgarm:
**BREAKING CHANGE** - as we learned the hard way in 0.12.2 this is a breaking change, as JSII bindings for Python are affected by changes to methods of parent classes and this could be the case for static methods as well. This is not a breaking change for users but it is internally as such that pre-built provider bindings need to be generated against this version to work.

Related to #1132

### DIFF
--- a/packages/cdktf/lib/terraform-stack.ts
+++ b/packages/cdktf/lib/terraform-stack.ts
@@ -24,7 +24,6 @@ import { App } from "./app";
 import { TerraformBackend } from "./terraform-backend";
 
 // eslint-disable-next-line @typescript-eslint/ban-types
-type Constructor<T> = Function & { prototype: T };
 type StackIdentifier = string;
 type OutputIdMap =
   | { [constructId: string]: string }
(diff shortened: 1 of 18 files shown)


ansgarm:
(converted to draft to not accidentally merge it before we're ready to release a next major version; as this can't be released as patch)

packages/cdktf/lib/terraform-stack.ts
```diff
@@ -119,24 +118,14 @@ export class TerraformStack extends Construct {
...
-
-    if (!predicate) {
-      throw new Error("Either byConstructor or byPredicate must be provided");
-    }
-
     const items: T[] = [];
 
     const visit = async (node: IConstruct) => {
-      if (predicate(node)) {
+      if (byPredicate(node)) {
```
DanielMSchmidt, line 128: Why don't we just have the predicate be the only positional argument? It seems unnecessary to me in this case to have an object deconstruction

packages/cdktf/package.json
```diff
@@ -59,7 +59,8 @@
     "parser": "@typescript-eslint/parser",
     "plugins": [
       "@typescript-eslint",
-      "jsdoc"
+      "jsdoc",
+      "no-instanceof"
```
DanielMSchmidt, line 63: Nice :)

DanielMSchmidt (approved):

PR #2130, opened 2022-09-19
chore(cli): Move things to src/ folder

mutahhir:
> **Warning**
> This is an experiment, a bit of a RFC

While working on the CLI, I found it a bit annoying that the *.js and *.d.ts files show up alongside our *.ts files. I also realize that we're compiling everything twice because we do `tsc` and `esbuild`. So, this PR is kind of a proposal / rfc for:

- Moving all typescript sources within the src/ folder
- Since `esbuild` doesn't validate typescript, we still need to use `tsc`, but can we enable `--noEmit` to just perform type-checking.

#### Incremental builds
Unfortunately, we lose the ability to use incremental builds because of this. `--noEmit` and `--incremental` [work together in Typescript 4.0+](https://devblogs.microsoft.com/typescript/announcing-typescript-4-0-beta/#noemit-and-incremental), but we're restricted in that regard due to JSII.

#### CDKTF Application Entrypoint
One of the challenges we have with this change is the `cdktf` application file. It has the following code:

```
#!/usr/bin/env node
require("./cdktf.js");
```
We reference this file to run some [integration-like tests](https://github.com/hashicorp/terraform-cdk/blob/69192b10f0165ed34b07af4a373e9a79c645bc0b/packages/cdktf-cli/src/test/cmds/convert.test.ts#L8) which then start failing because the `cdktf.js` file is no longer emitted by tsc.

In this PR, I've moved that to using `bundle/bin/cdktf` since that's also being generated anyway. However, I'm unsure if this will work well with watching tests.


packages/cdktf-cli/package.json
```diff
@@ -6,9 +6,11 @@
     "cdktf": "bundle/bin/cdktf"
   },
   "scripts": {
-    "build": "tsc",
+    "compile-build-config": "tsc build.ts",
+    "prebuild": "yarn compile-build-config",
+    "build": "tsc --noEmit",
```
ansgarm, line 11: Why still invoke this if we also have esbuild? For type checking?

packages/cdktf-cli/src/lib/checkpoint.ts
```diff
@@ -142,7 +142,7 @@ information on how to disable it.`
 
 export async function ReportRequest(reportParams: ReportParams): Promise<void> {
   // we won't report when checkpoint is disabled.
-  if (CHECKPOINT_DISABLE) {
+  if (CHECKPOINT_DISABLE()) {
```
ansgarm, line 145: Why does this need to be a function now? just curious

packages/cdktf-cli/tsconfig.json
```diff
@@ -24,20 +24,19 @@
...
     "esModuleInterop": true
   },
   "files": [
-    "lib/ambient.d.ts"
+    "src/lib/ambient.d.ts"
   ],
   "include": [
-    "**/*.ts",
+    "src/**/*.ts",
     "test/ui/deploy.test.tsx"
```
ansgarm, line 35: That file now is in `src/test/ui...`
Might need to be adjusted

packages/cdktf-cli/package.json
```diff
@@ -6,9 +6,11 @@
...
     "postbuild": "node build.js",
-    "watch": "tsc -w",
+    "watch": "yarn compile-build-config && node build.js --watch",
     "watch-preserve-output": "tsc -w --preserveWatchOutput",
```
ansgarm, line 14: This line needs to be adjusted as well, as this is invoked by the "global" watch command (i.e. when running `yarn watch` in the root of this repository).

ansgarm (commented):
Nice work! I like where this is going.

One thing we need to add to the Changelog is a note that users might need to update their shell completion script. For example, I added the output from `cdktf completion` to my shell config which contained an old path to the cdktf binary. So, if users copied that code instead of using something like `eval $(cdktf completion)`, they need to update it.

Also: Running `yarn watch` in the root of the repository does not work if I use `bundle/bin/cdktf` (it does still work when using `src/bin/cdktf` because the `watch-preserve-output` script still creates js files in all the directories as before – see my comment on that script, we should adjust that one as well).

packages/cdktf-cli/package.json
mutahhir: Yep! Esbuild doesn't typecheck IIRC

packages/cdktf-cli/src/lib/checkpoint.ts
mutahhir: I'll take this out, as it doesn't make sense in this PR, but I was a bit confused about this test working:

https://github.com/hashicorp/terraform-cdk/blob/e05a7c999cd38d20553f3c9daac26fd4a1f47da2/packages/cdktf-cli/src/test/checkpoint.test.ts#L36

It does seem to succeed, but I'm not sure why :/ since the process.env is read and set in this variable, and doesn't get re-read later.

DanielMSchmidt (approved):
Nice, seems like esbuild is also quite fast so watch mode feels even more performant although it's already bundling

CHANGELOG.md
```diff
@@ -151,6 +151,8 @@ A very minor change in the UX of the `cdktf get` command now generates the provi
 - chore: measure and print memory consumption of build and synth steps for examples in CI [\#2060](https://github.com/hashicorp/terraform-cdk/pull/2060)
 - chore: start running tests against TF 1.2.8 and drop running against 1.0.7 [\#2058](https://github.com/hashicorp/terraform-cdk/pull/2058)
 
+Only for developers of `terraform-cdk`: we modified how the `cdktf` cli is built. This might have an effect on the generated completion script. It is recommended to build the `cdktf-cli` package and update your completion script.
```
ansgarm, line 154: That's not true – users of the CDKTF could've copied this script as well 🙈 They also don't have build the CLI – running `cdktf completion` after updating will give them the latest script.

ansgarm: Just checked, and our docs even describe a way of setting up completions that requires an update (as it echoes the completion script at the time of running that command into the shell config):
https://github.com/hashicorp/terraform-cdk/blob/87170fd3e46f62f62e26037ba158cc4ef52a3fa3/website/docs/cdktf/cli-reference/commands.mdx#L26-L64

We need to update this docs page as well, as it probably contains the old script.

mutahhir: Can you explain a bit more about how this PR would modify normal users behavior? They'd continue to use it with the homebrew (or npm) version of cdktf, no? I'm happy to change the text, but I just wanted to understand a bit more about how this breaks. Shell autocompltion is a bit of a mystery to me still 🙈 😅

ansgarm: Hi, sorry for all the fuss. When I ran your PR I must've misled myself due to the way my completions script was set-up 😅
I just ran `cdktf completion` and it outputted a script containing `cdktf` as the binary (which indicates no change and `CHANGELOG` entry is needed as it won't affect users or developers).
When the script is generated (i.e. when running `cdktf completion`) by default it's rendering referencing the `cdktf` command (which is even visible in the docs I linked above 🤦). See this line:
https://github.com/hashicorp/terraform-cdk/blob/87170fd3e46f62f62e26037ba158cc4ef52a3fa3/website/docs/cdktf/cli-reference/commands.mdx#L46

I must've adjusted this line in my local completions script manually to:
```
' reply=($(COMP_CWORD="$((CURRENT-1))" COMP_LINE="$BUFFER" COMP_POINT="$CURSOR" /Users/ansgar/projects/hashicorp/terraform-cdk/packages/cdktf-cli/bundle/bin/cdktf --get-yargs-completions "${words[@]}"))
```
(I probably did this so that my completions support the latest flags I built locally and forgot about it)

And this path (`bundle/bin/cdktf`) has been changed by this PR, but as explained, this is only my problem and not a problem for our users 😬

So, we can actually remove that line 😅
```suggestion
```

ansgarm (approved):
Nice work, @mutahhir 🚀

As I was asking this myself (again) while looking at the build steps in the `cdktf-cli/package.json`: Can we document somewhere that we're running `tsc` because `esbuild` doesn't do any type checking? I don't have a good proposal for a good location sadly – maybe at the top of `build.ts`?

mutahhir:
Added note to CONTRIBUTING.md

PR #2138, opened 2022-09-23
allow resources named object

ansgarm:
- fix(provider-generator): Add special case in case resource is named 'object'
- test(provider-generator): Add test for resources named object or string


DanielMSchmidt (approved):
Please add a follow up ticket to adjust convert to handle this case

Issue #1815, opened 2022-05-24
Cache calls to the NPM Registry


ansgarm:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

The `cdktf provider add` command added in [#1761](https://github.com/hashicorp/terraform-cdk/pull/1761) queries package information from NPM. As that information does not change after a package has been released, we could and should cache those calls. At least we should cache calls within a single run of the command. I think we currently even query the same data twice.

PR #2145, opened 2022-09-28
chore(cli): Use the npm package's repository field for generating go package name

mutahhir:
Related to the move of our pre-built providers to the `cdktf` github organization.
Also fixes https://github.com/hashicorp/terraform-cdk/issues/1815

The core intention of the PR is to use some of the information published by projen on the pre-built provider's package.json, especially the 'repository' field, to generate the package module path for Go. This will prevent `cdktf provider add` to be broken as we move the providers to the `cdktf` organization.

### DIFF
--- a/packages/cdktf-cli/lib/dependencies/dependency-manager.ts
+++ b/packages/cdktf-cli/lib/dependencies/dependency-manager.ts
@@ -9,6 +9,7 @@ import { CdktfConfigManager } from "./cdktf-config-manager";
 import { PackageManager } from "./package-manager";
 import {
   getNpmPackageName,
+  getPrebuiltProviderRepositoryName,
   getPrebuiltProviderVersion,
 } from "./prebuilt-providers";
 import { getLatestVersion } from "./registry-api";
@@ -197,7 +198,6 @@ export class DependencyManager {
       );
     }
 
-    const packageName = this.convertPackageName(npmPackageName);
     const prebuiltProviderVersion = await getPrebuiltProviderVersion(
       constraint,
       this.cdktfVersion
@@ -209,6 +209,13 @@ export class DependencyManager {
     }
 
     const packageVersion = prebuiltProviderVersion;
+    const prebuiltProviderRepository = await getPrebuiltProviderRepositoryName(
+      npmPackageName
+    );
+    const packageName = this.convertPackageName(
+      npmPackageName,
+      prebuiltProviderRepository
+    );
 
     await this.packageManager.addPackage(packageName, packageVersion);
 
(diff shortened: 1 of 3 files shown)


packages/cdktf-cli/lib/dependencies/dependency-manager.ts
```diff
@@ -241,10 +248,14 @@ export class DependencyManager {
   /**
    * Converts an NPM package name of a pre-built provider package to the name in the target language
    */
-  private convertPackageName(name: string): string {
+  private convertPackageName(name: string, repository: string): string {
     const providerName = name.replace("@cdktf/provider-", "");
     switch (this.targetLanguage) {
       case Language.GO: // e.g. github.com/hashicorp/cdktf-provider-opentelekomcloud-go/opentelekomcloud
```
ansgarm, line 254: While we're at it we could update the comment 😁
```suggestion
      case Language.GO: // e.g. github.com/cdktf/cdktf-provider-opentelekomcloud-go/opentelekomcloud
```

ansgarm: (even if it is not yet true, but it soon will be)

packages/cdktf-cli/lib/dependencies/prebuilt-providers.ts
```diff
@@ -21,15 +21,68 @@ type ProvidersMap = {
   [name: string]: string;
 };
 
+const fetchCache = new Map<string, any>();
```
ansgarm, line 24: Nice! This closes https://github.com/hashicorp/terraform-cdk/issues/1815 already 😁

packages/cdktf-cli/test/lib/dependencies/prebuilt-providers.test.ts
```diff
@@ -0,0 +1,231 @@
+// Copyright (c) HashiCorp, Inc
+// SPDX-License-Identifier: MPL-2.0
+import nock from "nock";
```
ansgarm, line 3: Love the tests you added 🙇

packages/cdktf-cli/lib/dependencies/dependency-manager.ts
```diff
@@ -241,10 +248,14 @@ export class DependencyManager {
...
+        if (repository) {
+          return `${repository}-go/${providerName}`;
+        }
+
         return `github.com/hashicorp/cdktf-provider-${providerName}-go/${providerName}`;
```
ansgarm, line 259: I'm torn on whether we should fallback to the old one here. My hunch is that in the future Projen might introduce a change that changes the repository url format in the package.json which might cause this to break – a bit far fetched, I know 😁
But I can also see the opposite where we don't know whether we always had the repository in the package.json in the right format 🤔
I have no idea here, feel free to just discard this – I think yours is safer for today 😅
```suggestion
        return `github.com/cdktf/cdktf-provider-${providerName}-go/${providerName}`;
```

ansgarm (approved):
👏 Stellar PR, @mutahhir 🚀 Loved going through it commit by commit 💪

packages/cdktf-cli/lib/dependencies/dependency-manager.ts
mutahhir: Yeah, this is mostly just a 'if something's gone _really_ wrong' kinda thing. That being said, we're probably safe that it won't happen right now, so your change makes more sense.

packages/cdktf-cli/lib/dependencies/dependency-manager.ts
mutahhir: :/ Github didn't let me merge these in a single commit, so decided to do it manually.

ansgarm: No worries!

PR #2153, opened 2022-09-30
feat(provider-generator): Move generated files into their own directories

### LINKED ISSUE #2147: Add docs to new resource namespaces
### Community Note

### Description

Follow-up issue for https://github.com/hashicorp/terraform-cdk/pull/2087

- [x]  Add doc string for each of the namespaces with a link to the TF docs

More context from [Daniels comment](https://github.com/hashicorp/terraform-cdk/pull/2087#issuecomment-1256361841):

> RE Add doc string for each of the namespaces with a link to the TF docs: The docs are only picked up if you put them into a README.md in the folder of the namespace. The thing is right now we don't have a folder for the namespace, so this is not doable right now without changing this. Should have no effect besides adding the documentation to the namespace, but it's more work than I can finish today.

ansgarm:
Closes #2147

### Todos
- [x] Put generated code into sub-directory
- [x] Update tests
- [x] Add Markdown file to newly generated directory with link to Terraform Registry
- [x] Add new test to test Markdown file existence & content (if not already part of other snapshots)


packages/@cdktf/provider-generator/lib/get/generator/provider-generator.ts
```diff
@@ -168,20 +169,35 @@ export class TerraformProviderGenerator {
...
       providerResource.providerVersion = providerVersion;
       files.push(this.emitResource(providerResource));
+      this.emitResourceReadme(providerResource);
     }
 
     this.emitIndexFile(name, files);
   }
 
+  private emitResourceReadme(resource: ResourceModel): void {
+    const filePath = `${resource.namespaceFolderPath}/README.md`;
```
mutahhir, line 179: Should we use `path.join` here too?

packages/@cdktf/provider-generator/lib/get/generator/provider-generator.ts
```diff
@@ -168,20 +169,35 @@ export class TerraformProviderGenerator {
...
+      ? resource.provider
+      : resource.terraformType;
+    this.code.line(
+      `Refer to the Terraform Registory for docs: [\`${type}\`](${resource.linkToDocs}).`
+    );
+    this.code.closeFile(filePath);
+  }
+
   private emitIndexFile(provider: string, files: string[]): void {
     const folder = `providers/${provider}`;
```
mutahhir, line 193: I haven't worked with Windows machines in a while, so I don't know if these hardcoded `/` will break on windows?

mutahhir (approved):
Overall, good to go! Just one question in general about joining paths using `path.join` instead of templated strings.

packages/@cdktf/provider-generator/lib/get/generator/provider-generator.ts
ansgarm: I think they work for Nodejs – at least we already use them that way.

packages/@cdktf/provider-generator/lib/get/generator/provider-generator.ts
ansgarm: I briefly thought about it, but refrained from doing so, because the rest of this file is already only doing it manually 😄 And I didn't want to accidentally break things (by refactoring those bits as well then).

ansgarm: But I can add this improvement if you want?

packages/@cdktf/provider-generator/lib/get/generator/provider-generator.ts
mutahhir: I think `require` and `import` fixes that internally, but I'm not sure if `fs` fixes the paths too. At least, I can't find any documentation around it.

Issue #1677, opened 2022-03-30
Test existence of Providers with properties


coen-hyde:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description
I have a stack that generates a lot of providers. I would like to write tests to ensure they exist and were created with specific properties. This doesn't appear possible atm. It would be good to have something similar to `toHaveResourceWithProperties` but for providers. Please correct me if this is indeed possible.

skorfmann:
To clarify: You want to test the generated provider bindings itself, or the instantiated provider in a stack? I'm assuming you're using Typescript?

aayushharwani-aidash:
@coen-hyde  @ansgarm  I am new to community. Can you just throw some light on this issue ? I am interested in working on it.

jsteinich:
@aayushharwani-aidash This would be adding additional testing capabilities. Take a look [here](https://www.terraform.io/cdktf/test/unit-tests) for more information on testing.

Technically this involves adding another function similar to this one, but for looking at the `provider` and maybe `terraform.required_providers` sections of the synthesized output:
https://github.com/hashicorp/terraform-cdk/blob/a3338f49abc001dafe35833511392f45553b1a61/packages/cdktf/lib/testing/matchers.ts#L151

We would also want to update the [testing docs](https://github.com/hashicorp/terraform-cdk/blob/main/website/docs/cdktf/test/unit-tests.mdx) to include the new functions.

aayushharwani-aidash:
I understood issue. I am interested to work on it but not getting where to start. Will anyone like to help me ? @jsteinich @ansgarm @skorfmann

jsteinich:
> I understood issue. I am interested to work on it but not getting where to start. Will anyone like to help me ? @jsteinich @ansgarm @skorfmann

Are you in the [cdk dev slack](https://join.slack.com/t/cdk-dev/shared_invite/zt-xtpfmrqt-6ormYTA0hLdpMSAtTkM_2A)? You could ask specific questions in the `terraform-cdk` channel.

PR #2154, opened 2022-09-30
feat(lib): Added testing matchers for Providers

Maed223:
Closes: #1677


DanielMSchmidt (commented):
Looks good function wise 👍  There are some documentation related things missing:

- [x] Please add docstrings to the testing class (Ideally to all if you have the time) so our documentation generation gets better
- [ ] Let's add these new functions with an example to the docs
- [ ] We should also add an example to the generated demo test classes we have in our init templates / examples

ansgarm (approved):

PR #2158, opened 2022-10-03
fix: Prevent any release jobs from triggering if no release is required

mutahhir:
This updates the `release` workflow, to:

- Short-circuit the workflow is no new release needs to be made
- Enable publishing version to homebrew when new release is made

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,6 +24,7 @@ jobs:
     outputs:
       tests: ${{ steps.build-test-matrix.outputs.tests }}
       version: ${{ steps.get_version.outputs.version }}
+      release_status: ${{ steps.get_release_status.outputs.release }}
     container:
       image: docker.mirror.hashicorp.services/hashicorp/jsii-terraform
     env:
@@ -82,6 +83,7 @@ jobs:
   integration_test:
     needs:
       - prepare-release
+    if: jobs.prepare-release.outputs.release_status == 'unreleased'
     uses: ./.github/workflows/integration.yml
     with:
       skip_setup: true
@@ -102,6 +104,7 @@ jobs:
 
   unit_test:
     uses: ./.github/workflows/unit.yml
+    if: jobs.prepare-release.outputs.release_status == 'unreleased'
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -121,6 +125,7 @@ jobs:
       - integration_test
       - unit_test
     runs-on: ubuntu-latest
+    if: jobs.prepare-release.outputs.release_status == 'unreleased'
```
ansgarm, line 128: While in the same workflow (which had a previously unreleased Sentry version) we can re-run this step without it short-circuiting here, right?
While I like the simplicity of re-using the Sentry state of the release – I'm wondering whether there are edge cases where we could lock ourselves out of retrying a release – but I suppose that might happen with any way of storing this "released?" state 🤔

ansgarm (approved):
Nice! A bit unsure about certain edge cases, but that's a feeling and not a fact 😄

.github/workflows/release.yml
mutahhir: I think that's a reasonable issue with the current way. Unfortunately, I don't know how to best solve it :/. I think what we have right now is the ability to release and let npm / pypi etc. deal with the duplication. 

The alternative would be to go into sentry and delete the release before retrying? I don't know if that's possible or how to do it though.

mutahhir: @ansgarm since you had a recent experience with this, do you think we can document the sentry release deletion process to unblock this?

ansgarm: Yeah!
It's basically this:
```
# Install Sentry CLI and login
npm i -g @sentry/cli
sentry-cli login
# List all releases (optional)
sentry-cli releases list --org hashicorp
# Delete the release, Note: there will be no confirmation for deleting the release!
sentry-cli releases delete --org hashicorp <release> # e.g. cdktf-cli-0.14.0
```

mutahhir: Actually, I was doing some further research on the sentry-cli commands that we use, and I realize that we don't need this. I've added it anyway since it's good context for situations like we had earlier, but it's not relevant to the current PR. 

image

PR #2168, opened 2022-10-04
fix(provider-generator): change Static resource class to StaticResource as their namespace 'static' is not valid in TypeScript re-exports

ansgarm:
e.g. happens in the time Terraform Provider


mutahhir (approved):
Looks good!

ansgarm:
CI fails because the constructs package couldn't be released to NuGet: https://github.com/aws/constructs/actions/runs/3178412027/jobs/5180232659 for its latest version apparently

jsteinich:
Do we want to try and be more proactive here? What I mean is that there are plenty of other names which would cause issues (for example access modifiers).

ansgarm:
@jsteinich yeah, makes sense. Do you happen to know a good source for works in TypeScript that can't be used in these kinds of expressions?
```ts
export * as static from './xy'
```

ansgarm:
I found this list of keywords: https://github.com/microsoft/TypeScript/blob/503604c884bd0557c851b11b699ef98cdb65b93b/src/compiler/types.ts#L114-L197

And tested them using:
<details>
<summary>
this snippet
</summary>

```ts
export * as break from "./index";
export * as case from "./index";
export * as catch from "./index";
export * as class from "./index";
export * as const from "./index";
export * as continue from "./index";
export * as debugger from "./index";
export * as default from "./index";
export * as delete from "./index";
export * as do from "./index";
export * as else from "./index";
export * as enum from "./index";
export * as export from "./index";
export * as extends from "./index";
export * as false from "./index";
export * as finally from "./index";
export * as for from "./index";
export * as function from "./index";
export * as if from "./index";
export * as import from "./index";
export * as in from "./index";
export * as instanceof from "./index";
export * as new from "./index";
export * as null from "./index";
export * as return from "./index";
export * as super from "./index";
export * as switch from "./index";
export * as this from "./index";
export * as throw from "./index";
export * as true from "./index";
export * as try from "./index";
export * as typeof from "./index";
export * as var from "./index";
export * as void from "./index";
export * as while from "./index";
export * as with from "./index";
export * as implements from "./index";
export * as interface from "./index";
export * as let from "./index";
export * as package from "./index";
export * as private from "./index";
export * as protected from "./index";
export * as public from "./index";
export * as static from "./index";
export * as yield from "./index";
export * as abstract from "./index";
export * as accessor from "./index";
export * as as from "./index";
export * as asserts from "./index";
export * as assert from "./index";
export * as any from "./index";
export * as async from "./index";
export * as await from "./index";
export * as boolean from "./index";
export * as constructor from "./index";
export * as declare from "./index";
export * as get from "./index";
export * as infer from "./index";
export * as intrinsic from "./index";
export * as is from "./index";
export * as keyof from "./index";
export * as module from "./index";
export * as namespace from "./index";
export * as never from "./index";
export * as out from "./index";
export * as readonly from "./index";
export * as require from "./index";
export * as number from "./index";
export * as object from "./index";
export * as satisfies from "./index";
export * as set from "./index";
export * as string from "./index";
export * as symbol from "./index";
export * as type from "./index";
export * as undefined from "./index";
export * as unique from "./index";
export * as unknown from "./index";
export * as from from "./index";
export * as global from "./index";
export * as bigint from "./index";
export * as override from "./index";
export * as of from "./index";
```

</details>

which showed errors for:
```
implements
interface
let
package
private
protected
public
static
yield
await
```

So, I'll push a commit to add those as well 👍

Issue #1641, opened 2022-03-22
Resource FQN tokens aren't resolved in a user friendly way


jsteinich:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### cdktf & Language Versions

0.9.0. Any language.

### Affected Resource(s)

Attempting to use `.fqn` on any resource when appending to it or using in an array (probably other cases as well).
This mainly comes up when needing to use escape hatches.

### Expected Behavior

Synthesized output is valid.

### Actual Behavior

- `Expected the start of an expression, but found an invalid expression token`
- jsii.errors.JSIIError: Expected array type, got "<unresolved-token>"
- X is object with Y attributes

### Important Factoids

Support for cross stack references caused a change in behavior. Previously `.fqn` actually return just a raw string (string token of string is the input string).

### References

- https://discuss.hashicorp.com/t/python-escape-hatching-in-0-9-x/37202

armandosorianopfs:
@DanielMSchmidt you made my day!! thanks!!

If anyone reaches this issue with similar issue note that you need to use `lookup` to resolve object properties and `element` to resolve arrays. For my case, I've reached to this:
```
private getIpFromStatusFqn(fqn: string): string {
    const resolvedFqn = Fn.element(fqn as unknown as any[], 0);
    const loadBalancer = Fn.lookup(resolvedFqn, 'load_balancer', []);
    const loadBalancerFirstElement = Fn.element(loadBalancer, 0);
    const ingress = Fn.lookup(loadBalancerFirstElement, 'ingress', []);
    const ingressFirstElement = Fn.element(ingress, 0);

    return Fn.lookup(ingressFirstElement, 'ip', 'no-ip');
  }
```

mutahhir:
@paymog I've been looking into this issue, and I was hoping to get some more information.

### Why is the syntax error happening
From what I understand, the root of the problem is this pattern:

```tf
${length(${data.foo}.names)}
```

Our token system assumes that we're going to be used within a string and adds interpolation syntax `${}` around the token. However, since you're already wrapping it within a `${}`, that becomes a syntax error in Terraform.

[...]

PR #2179, opened 2022-10-12
feat: [Proposal] Enable non-escaping mode for tokens

mutahhir:
Relates to: https://github.com/hashicorp/terraform-cdk/issues/1641

This PR proposes to use the HCL syntax for string interpolation (`${}`) to alter how CDKTF outputs [Tokens](https://developer.hashicorp.com/terraform/cdktf/concepts/tokens). What I'm looking for right now is some feedback around what kinds of potential problems this introduces that I may not have thought about and the relative usefulness of such things pertaining to its ease of use with [escape hatches](https://developer.hashicorp.com/terraform/cdktf/concepts/resources#escape-hatch) and overrides.

### Problem
As the above issue mentions, working with FQN is hard currently, and the root cause seems to be the resolution of tokens. The easiest example is from the above issue, where a user tried the following:

```typescript
efsMountTarget.addOverride(
      "count",
      `\${length(${dataAwsAvailabilityZonesAll.fqn}.names)}`
    );
```
Here's the expectation is that the value of `count` in the output would be: `${length(data.allAvailableZones.name)}`, however, earlier to this PR, we would output:

```tf
"${length(${data.allAvailableZones}.name)}"
```

This then trips up Terraform because `${}` is only valid within a string and not within normal HCL, causing errors like: `Error: Invalid character` or `Error: Invalid expression` etc.

### Current workarounds / preferred ways
As already mentioned in the above issue, a few work arounds for this would be to use CDKTF's [Terraform functions](https://developer.hashicorp.com/terraform/cdktf/concepts/functions). If we were to use those, the above example could become:

```typescript
   const zoneNames = Fn.lookup(dataAwsAvailabilityZonesAll.fqn, "names", undefined)
   efsMountTarget.addOverride(
     "count",
     Fn.lengthOf(zoneNames)
   );
```

While this is a pretty nice example where I think using Terraform functions is recommended, there are times when using them is a bit more difficult / cumbersome. From the original issue:

```
const webappService = new Service(this, 'k8s-service-webapp', {
  ...
  waitForLoadBalancer: true,
});

new DnsRecordSet(this, `webapp-dns-zone-record`, {
  ...
  rrdatas: [(webappService?.status.fqn as any)[0]?.load_balancer?.ingress?.ip], // won't be resolved properly
});
```

The solution using Terraform functions is:

```typescript
Fn.lookup(Fn.lookup(Fn.lookup(Fn.element(webappService.status.fqn, 0), "load_balancer"), "ingress"), "ip")
```

which can be a bit complicated to figure out.

### Proposal
With this PR, I'd like to make a proposal with adding something like a 'raw' mode where the user can signal CDKTF to stop adding any additional string interpolation syntax (`${}`) in it's output. Here's how this would look like for the first example:

```typescript
efsMountTarget.addOverride(
      "count",
      `\${length(${dataAwsAvailabilityZonesAll.fqn}.names)}`
    );
```

This example would no longer create an error because the output will be: 

```tf
"${length(data.allAvailableZones.name)}"
```

The second example becomes:

```typescript
const webappService = new Service(this, 'k8s-service-webapp', {
  ...
  waitForLoadBalancer: true,
});

new DnsRecordSet(this, `webapp-dns-zone-record`, {
  ...
  rrdatas: [`$\{${webappService?.status.fqn}[0].load_balancer[0].ingress[0].ip}`],
});
```

> **Warning**
> Within the top-level interpolation sequence, CDKTF will continue suppressing the interpolation syntax, so things like `` `$\{"test-${resource.fqn}"}` `` would not have the intended effect. However, there can be multiple top-level interpolation sequences, like: `` `$\{${resource1.fqn}.name}-${Fn.lookup(resource2.fqn, "names", undefined)}` `` that would do the right thing.

### TODOs
- [ ] Update [functions.mdx documentation](https://github.com/hashicorp/terraform-cdk/blob/main/website/docs/cdktf/concepts/functions.mdx)
- [ ] More tests for edge cases

### DIFF
--- a/packages/cdktf/lib/tfExpression.ts
+++ b/packages/cdktf/lib/tfExpression.ts
@@ -58,7 +58,11 @@ class TFExpression extends Intrinsic implements IResolvable {
     }
 
     // Only a token reference
-    if (tokenList.literals.length === 0 && numberOfTokens === 1) {
+    if (
+      tokenList.literals.length === 0 &&
+      tokenList.escapes.length === 0 &&
+      numberOfTokens === 1
+    ) {
       return resolvedArg;
     }
 
--- a/packages/cdktf/lib/tfExpression.ts
+++ b/packages/cdktf/lib/tfExpression.ts
@@ -6,7 +6,6 @@ import { Tokenization, Token } from "./tokens/token";
 import { App } from "./app";
 import { TerraformStack } from "./terraform-stack";
 import { ITerraformDependable } from "./terraform-dependable";
-import { TokenString } from "./tokens/private/encoding";
 
 // eslint-disable-next-line jsdoc/require-jsdoc
 class TFExpression extends Intrinsic implements IResolvable {
@@ -322,17 +321,11 @@ class FunctionCall extends TFExpression {
   public resolve(context: IResolveContext): string {
     const suppressBraces = context.suppressBraces;
     context.suppressBraces = true;
+    context.ignoreEscapes = true;
 
     const serializedArgs = this.args
       .map((arg) => this.resolveArg(context, arg))
       .join(", ");
-
-    const escapeRegex = TokenString.forEscape(serializedArgs);
-    if (escapeRegex.test()) {
-      throw new Error(
-        "Cannot use terraform escape sequences within CDKTF Functions. Please use the escape sequence at the beginning and use terraform functions directly"
-      );
-    }
 
     const expr = `${this.name}(${serializedArgs})`;
 
--- a/packages/cdktf/lib/tfExpression.ts
+++ b/packages/cdktf/lib/tfExpression.ts
@@ -85,13 +85,23 @@ class TFExpression extends Intrinsic implements IResolvable {
         }
 
         // if left is only a token, needs to be wrapped as terraform expression
-        if (leftTokenList.literals.length === 0 && leftTokenCount === 1) {
+        if (
+          leftTokenList.literals.length === 0 &&
+          leftTokenList.escapes.length === 0 &&
+          leftTokenCount === 1
+        ) {
           leftValue = `\${${leftTokens[0]}}`;
+        }
+
+        if (leftTokenList.escapes.length === 1 && leftTokenCount === 0) {
+          leftValue = `${leftTokenList.escapes[0]}`;
         }
 
         const rightValue =
           rightTokens.length === 0
             ? this.escapeString(right)
+            : leftTokenList.escapes.length > 0
+            ? rightTokens[0]
             : `\${${rightTokens[0]}}`;
 
         return `${leftValue}${rightValue}`;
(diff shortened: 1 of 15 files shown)


jsteinich:
Preface: I haven't looked at the code changes yet; just going off of the proposal text.

Having a raw mode is a reasonable idea and one that many programming languages have used. I can definitely see a path for it working; however, there's a few reasons that I would recommend against it:

* The particular syntax suggested is generally difficult to use in Typescript as it requires escapes. Perhaps a different syntax would solved that, but we support 5 different languages, so likely will have issues in at least one language.
* We likely would end up losing the reference tracking that was the impetus for needing this.
* An opt-in model is yet an extra step for users to learn about and remember to use.

A possible alternative route is to implicitly remove excess interpolations with an opt out option. Conceptually the idea is similar to the resolve context that we already have which removes the interpolation when the context says there is already an outer one present. Practically, this may be better suited as an after resolve step where we detect an invalid nested interpolation and simply remove it. Opting out would require the same syntax as is used in hcl to escape. In this regard, any special syntax would rarely be required and reference tracking would be left intact.

jsteinich:
> Can you make a proposal of how this looks like, I'm not sure I fully follow.

I realized that what I was proposing is actually the same thing as this proposal. Ultimately it still was a detect if within an hcl interpolation block and if so remove extra interpolations.
It might be simpler to implement as a check on the synthesized output (can be before it's written to disk). At that stage tokens have already been resolved so it should become just checking for nested interpolations.

mutahhir:
I'm currently stuck with dealing with escapes and Terraform functions using `Fn.*`. Will circle back as I think about it more.

team-tf-cdk:

team-tf-cdk:

team-tf-cdk:

team-tf-cdk:

DanielMSchmidt (approved):
I think this is a good change in general! It improves the current situation so I see no need to block it. I would maybe add a follow up issue to create a warning / throw an error if `Fn.*` is mixed with this interpolation, so that we can mitigate situations where users would be surprised and educate them instead :)

ansgarm (approved):
Looks good (as far as I can tell at least 😄) 👍

Just out of curiosity: Does this example need to be updated? Or does it still work the way it is written currently?
https://github.com/hashicorp/terraform-cdk/blob/47a562d4fbafd77e6e9a2e26783e1bc1188cd9bf/examples/typescript/google-cloudrun/main.ts#L64

mutahhir:
> Looks good (as far as I can tell at least 😄) 👍
> 
> Just out of curiosity: Does this example need to be updated? Or does it still work the way it is written currently?
> 
> https://github.com/hashicorp/terraform-cdk/blob/47a562d4fbafd77e6e9a2e26783e1bc1188cd9bf/examples/typescript/google-cloudrun/main.ts#L64

I don't think this example would've worked before this change (or at least after we introducted cross-stack references) 😅

team-tf-cdk:

team-tf-cdk:

Issue #2173, opened 2022-10-06
Provider Add Command should not depend on provider.json file in Github directly


ansgarm:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Currently the `provider add` command depends on the `provider.json` file in the repository-manager:
https://github.com/hashicorp/terraform-cdk/blob/b23fc173715e90c0a5b8c8633d9ec7f71edf9ed4/packages/cdktf-cli/lib/dependencies/prebuilt-providers.ts#L19

We should change this (e.g. by using a redirect via `cdk.tf` or similar) to ensure that we have full control over the location of this information in case we refactor pre-built providers in the future for example.

PR #2180, opened 2022-10-12
feat: add redirect link to pre-built providers json file

ansgarm:
In preparation for #2173, blocks #2181

### DIFF
--- a/cdk.tf/vercel.json
+++ b/cdk.tf/vercel.json
@@ -190,6 +190,10 @@
     {
       "source": "/crash-reporting",
       "destination": "https://www.terraform.io/cdktf/telemetry#crash-reporting"
+    },
+    {
+      "source": "/.well-known/prebuilt-providers.json",
+      "destination": "https://raw.githubusercontent.com/hashicorp/cdktf-repository-manager/main/provider.json"
     }
   ]
 }


cdk.tf/vercel.json
```diff
@@ -190,6 +190,10 @@
     {
       "source": "/crash-reporting",
       "destination": "https://www.terraform.io/cdktf/telemetry#crash-reporting"
+    },
+    {
+      "source": "/.well-known/prebuilt-providers.json",
```
DanielMSchmidt, line 195: any reason for the `.well-known`? I couldn't find the prebuilt-providers.json in the [wiki](https://en.wikipedia.org/wiki/Well-known_URI#:~:text=A%20well%2Dknown%20URI%20is,well%2Dknown%20locations%20across%20servers.)

cdk.tf/vercel.json
```diff
@@ -190,6 +190,10 @@
     {
       "source": "/crash-reporting",
       "destination": "https://www.terraform.io/cdktf/telemetry#crash-reporting"
+    },
+    {
```
DanielMSchmidt, line 194: Can we add more "this is not for leisure but depended on" context or does that break the build?
```suggestion
    { "//": "This is used by the cdktf provider subcommands"
```

cdk.tf/vercel.json
ansgarm: I tried to use that to suggest that this is depended on. E.g. the TF Registry uses this as well with something not found in the wiki `https://registry.terraform.io/.well-known/terraform.json` 😁

cdk.tf/vercel.json
ansgarm: Yeah, such a comment seems reasonable – let's see whether Vercel allows for the `//` attribute.

ansgarm:
^ @DanielMSchmidt adding the comment is not supported by Vercel :/

cdk.tf/vercel.json
DanielMSchmidt: ok, I see :)

Issue #1929, opened 2022-07-15
Add multi-language code examples to docs 


schersh:
Most of our docs only include TypeScript examples. We should add multi-language code blocks to examples in our documentation to highlight that there are five supported languages to use with CDKTF.

Scanning the codebase shows that there are currently 26 code blocks in our docs, so this work could be handled incrementally. 

Reference: [Guidance from the Education team on using CodeTabs to add multiple languages to code examples](https://docs.google.com/document/d/1fmpSviM2pQUWm6y3QnVQ3QAdq5i7RjeWopzNXbMH2Io/edit#heading=h.sjgr1uoclelk)

pchaganti:
+1. Please provide examples in Python.

PR #2186, opened 2022-10-12
chore(docs): translate all examples to python

Maed223:
re: #1929


website/docs/cdktf/concepts/constructs.mdx
```diff
@@ -51,7 +51,7 @@ You can import any [CDKTF-compatible](#available-constructs) construct that is a
 The following TypeScript example instantiates a construct called `KubernetesWebAppDeployment` and uses the available arguments to specify that the deployment will have two replicas.
```
laurapacilio, line 51: I'm so excited to see this! One thing I want to say is, we probably want to remove "TypeScript" from the text above all the examples where we have more than one language. so it should say "The following example" instead :)

Nice work!

packages/cdktf/lib/terraform-local.ts
```diff
@@ -40,6 +40,22 @@ export class TerraformLocal
     return Token.asList(this.interpolation());
   }
 
+  public get asStringMap(): { [key: string]: string } {
```
DanielMSchmidt, line 43: We missed the feature, so one of our TS examples was just impossible to achieve in Python.

ansgarm: Good catch! As the fix won't show up in the generated changelog, can we add a small manual section to the changelog file that notes that this has been added (I don't think we need to call it breaking, as the 0.13->0.14 change will already "break" older pre-built providers.

DanielMSchmidt: Yeah makes sense 👍

website/docs/cdktf/concepts/iterators.mdx
```diff
@@ -98,29 +177,29 @@ new TeamMembers(this, "members", {
...
-        "role": "maintainer"
-    })
-)
+        github.team_members.TeamMembers(self, "members",
+                                        team_id=github_team.id,
+                                        members=org_member_iterator.dynamic({
+                                            "username": Token().as_string(org_member_iterator.value),
+                                            "role": "maintainer"
+                                        })
+                                        )
```
ansgarm, line 202: As commented above, we might want to get rid of all this indentation in a follow-up PR.
As opposed to the other comment, for this sample it would be enough to remove all leading indentation that all lines share.
The script should already do that [here](https://github.com/hashicorp/terraform-cdk/blob/39241c4640da6116485d8fb6ea243f596080dc36/tools/documentation-code-snippets/update-documentation-code-snippets.ts#L222), but that logic seems to have a bug (it should ignore empty lines, as they currently set the min value to 0 😅)

website/docs/cdktf/concepts/stacks.mdx
```diff
@@ -191,10 +361,30 @@ To make the dependency explicit, run`stackD.addDependency(stackC)`.
...
 If you want to keep the result of the function attached to one stack and save its state, create a [Terraform Local value](/cdktf/concepts/variables-and-outputs#local-values) and expose it. The following example creates a `TerraformLocal`.
 
+<CodeTabs>
+
 ```typescript
 this.allResources = TerraformLocal(this, "merged_items", Fn.mergeLists(resourceFromStackA.items, resourceFromStackB.items)
 ```
 
+```python
+
```
ansgarm, line 371: missing? 😅

ansgarm (commented):
Nice work, @Maed223 👏

Added a couple of small comments / nits, the bigger ones can go into follow-up issues (i.e. the whitespace stuff).

ansgarm:
Also: Vercel seems to be failing 😅

website/docs/cdktf/concepts/aspects.mdx
```diff
@@ -48,8 +50,28 @@ export class TagsAddingAspect implements IAspect {
...
+
+```
+
+```csharp
+
+```
+
+```go
+
+```
```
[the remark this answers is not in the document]
Maed223: Ah yes, it's a remnant from before I split up the initial PR into PR's based on language. Good catch on why the build is failing.

examples/python/documentation/data_sources.py
```diff
@@ -0,0 +1,67 @@
...
+                    region="us-east-1"
+                    )
+
+        # .....
+        remoteState = DataTerraformRemoteState(self, "vpc-prod-remote-state",
+                                               organization="hashicorp",
+                                               workspaces=NamedRemoteWorkspace(name='vpc-prod')
+                                               )
+
+        Instance(self, "foo",
```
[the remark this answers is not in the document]
Maed223: Coming from the original Typescript docs there doesn't seem to be consistent reasoning as to what imports are included in the snippets. Don't think this case matters too much since `Instance` could be interchangeable with another resource. Though more generally the selection of what to include could be helpful for important aspects of the snippet whose locations aren't immediately obvious.

website/docs/cdktf/concepts/constructs.mdx
Maed223: Good catch! Updated 😁

website/docs/cdktf/concepts/assets.mdx
```diff
@@ -72,36 +72,32 @@ app.synth();
...
-class MyStack(cdktf.TerraformStack):
+class MyAssetStack(cdktf.TerraformStack):
     def __init__(self, scope: Construct, name: str):
         super().__init__(scope, name)
 
-        AwsProvider(self, 'aws', region='eu-central-1')
+        aws.provider.AwsProvider(self, 'aws', region='eu-central-1')
 
-        bucket = S3Bucket(self, "bucket", bucket="demo")
+        bucket = aws.s3_bucket.S3Bucket(self, "bucket", bucket="demo")
```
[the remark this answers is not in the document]
Maed223: Good point I'd imagine it would impact performance especially consider the size of the aws provider. Updated 😁

website/docs/cdktf/concepts/iterators.mdx
```diff
@@ -61,6 +99,45 @@ const s3Bucket = new S3Bucket(this, "bucket", {
 });
 ```
 
+```python
+import imports.aws as aws
+from cdktf import TerraformIterator, TerraformVariable, TerraformLocal
+        list = TerraformLocal(self, "my-list", [
```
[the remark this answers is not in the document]
Maed223: Updated 😁

website/docs/cdktf/concepts/stacks.mdx
Maed223: Updated 😁

website/docs/cdktf/concepts/iterators.mdx
```diff
@@ -36,12 +40,46 @@ const s3Bucket = new S3Bucket(this, "bucket", {
 });
 ```
 
+```python
+import imports.aws as aws
+from cdktf import TerraformIterator, TerraformVariable, TerraformLocal
+        list = TerraformVariable(self, "list",
```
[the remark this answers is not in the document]
Maed223: Updated 😁

website/docs/cdktf/concepts/iterators.mdx
Maed223: 👍

DanielMSchmidt (approved):
🎉🎉🎉

website/docs/cdktf/concepts/functions.mdx
```diff
@@ -37,3 +41,35 @@ new TerraformOutput(this, "first-zone", {
...
+                                         state="available",
+                                         )
+
+        TerraformOutput(self, 'first-zone',
+                        value=Fn.element(zones.names, 0)
+                        )
+
+app = App()
+FunctionsStack(app, "functions")
+app.synth()
```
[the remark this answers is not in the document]
Maed223: 👍

PR #2199, opened 2022-10-18
fix(provider-generator): don't require arrays wrapping variable / output declarations

DanielMSchmidt:
While terraform gives us this format, a hand-written or generated .tf.json file might not have
extra arrays around the values of variables and outputs. This leads to errors when parsing, so
when cdktf get is invoked.


ansgarm:
I don't get it 😅 Where's the array in the HCL / HCL JSON test input? 😅

DanielMSchmidt:
There is no array in the test input. It fails currently in cases where variable / output values are not wrapped in json. The implementation for parsing HCL puts these parts in an array, TFModuleStack does not

ansgarm:
Ah 💡 Thanks for clarifying that!

ansgarm (approved):

PR #2213, opened 2022-10-21
chore: cache examples and terraform plugins

DanielMSchmidt:
Using caches always and also for terraform plugins should improve stability and reduces the test runtime of the unit tests by 7 minutes

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -68,3 +68,4 @@ jobs:
           TEST_TARGET: "${{ matrix.target }}"
           TERRAFORM_BINARY_NAME: "terraform${{ matrix.terraform }}"
           MAVEN_OPTS: "-Xmx3G"
+          TF_PLUGIN_CACHE_DIR: ${{ steps.yarn-cache-dir-path.outputs.dir }}/terraform-plugins
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -40,6 +40,16 @@ jobs:
     timeout-minutes: 60
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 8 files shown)


.github/workflows/examples.yml
```diff
@@ -40,6 +40,16 @@ jobs:
     timeout-minutes: 60
     steps:
       - uses: actions/checkout@v2
+      - name: Get yarn cache directory path
+        id: global-cache-dir-path
+        run: echo "::set-output name=dir::$(yarn cache dir)"
```
mutahhir, line 45: Nit: This is already getting deprecated now, should we find an alternative?

https://docs.github.com/en/actions/using-jobs/defining-outputs-for-jobs

```suggestion
        run: echo "dir=$(yarn cache dir)" >> $GITHUB_OUTPUT
```

.github/workflows/integration.yml
```diff
@@ -37,12 +37,12 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Get yarn cache directory path
-        id: yarn-cache-dir-path
+        id: global-cache-dir-path
         run: echo "::set-output name=dir::$(yarn cache dir)"
```
mutahhir, line 41: While we're here, maybe convert these too?

mutahhir (approved):
I don't have a detailed understanding of how the github action caching works, but this looks like a really nice improvement overall. Just some minor nits if we can improve while we're here.

.github/workflows/integration.yml
DanielMSchmidt: I'm doing this in another PR, so I'll leave it for now

PR #2214, opened 2022-10-21
chore(deps): Update docker image to use new jsii/superchain tag

mutahhir:
The old `node14` docker image tag is [now deprecated](https://hub.docker.com/r/jsii/superchain). This PR replaces it with the new recommended docker image tag.

### Background
We're seeing errors like the following image in our PRs:

image

From [this issue](https://github.com/golang/go/issues/51436) it seems like it's related to Go 1.16, which our `hashicorp/jsii-terraform` image was using. Since `jsii/superchain` was already on Go 1.18 already, we realized something's wrong with our docker image, leading us to the discontinued tags.


ansgarm (approved):
🐳

PR #2215, opened 2022-10-21
chore: add mergify

DanielMSchmidt:
Also made some CI changes (collecting success jobs, paths instead of paths-ignore) to make life easier with mergify

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -38,6 +38,13 @@ queue_rules:
       - -label~=(blocked/do-not-merge )
       - and: *ci_checks
 pull_request_rules:
+  - name: Trigger CI once the PR is not in draft
+    actions:
+      label:
+        add: ci/run
+    conditions:
+      - -conflict
+      - -draft
   - name: Automatic merge on approval and successful build
     actions:
       delete_head_branch: {}
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -1,0 +1,52 @@
+shared:
+  # sharing this configuration: https://docs.mergify.com/configuration/#shared
+  ci: &ci_checks
+    # We need to replicate the ignore-paths here as well
+    # examples.yml
+    - or:
+        # Either none of the files are modified
+        - -files~=^(examples/|packages/|.terraform.versions.json|package.json)
+        # Or we need the check to succeed
+        - status-success=examples_done
+    # unit.yml
+    - or:
+        # Either none of the files are modified
+        - -files~=^(packages/|.terraform.versions.json|package.json|tools/build-unit-test-matrix.sh)
+        # Or we need the check to succeed
+        - status-success=examples_done
+    # integration.yml
+    - or:
+        # Either none of the files are modified
+        - -files~=^(test/|packages/|.terraform.versions.json|package.json|tools/build-test-matrix.sh)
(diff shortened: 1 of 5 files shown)


.github/workflows/examples.yml
```diff
@@ -58,3 +58,10 @@ jobs:
           TEST_TARGET: "${{ matrix.target }}"
           TERRAFORM_BINARY_NAME: "terraform${{ matrix.terraform }}"
           MAVEN_OPTS: "-Xmx3G"
+
+  examples_done:
```
ansgarm, line 62: Maybe name this `examples_success`? Done sounds a bit ambiguous in that regard?

.github/workflows/integration.yml
```diff
@@ -112,12 +112,19 @@ jobs:
           TERRAFORM_BINARY_NAME: "terraform${{ matrix.terraform }}"
           NODE_OPTIONS: "--max-old-space-size=7168"
 
+  linux_integration_done:
```
ansgarm, line 115: nit: Same proposal regarding "done" wording as above 😇

.github/workflows/integration.yml
```diff
@@ -157,3 +164,10 @@ jobs:
           TEST_TARGET: ${{ matrix.target }}
           TERRAFORM_CLOUD_TOKEN: ${{ secrets.TERRAFORM_CLOUD_TOKEN }}
           NODE_OPTIONS: "--max-old-space-size=7168"
+
+  windows_integration_done:
```
ansgarm, line 168: nit: We could combine this in a single "integration_success" job and depend on both linux and windows. No hard feelings either in favor or against it, though 😅

.github/workflows/examples.yml
```diff
@@ -1,19 +1,19 @@
 name: Examples Integration
 on:
   pull_request:
-    paths-ignore:
-      - "**.md"
-      - "docs/**"
-      - "website/**"
-      - "cdk.tf/**"
-      - "tools/update-github-project-board/**"
+    paths:
```
ansgarm, line 4: maybe add an info / link here (and in the other workflow files) that notes that changes here also need to be done to the Mergify config so CI does not accidentally block on this (because Mergify waits on success for a job that is newly ignore by changes here).

.github/workflows/integration.yml
ansgarm: @DanielMSchmidt any comment on why two is better than one? 😁

.github/mergify.yml
```diff
@@ -0,0 +1,119 @@
...
+    conditions:
+      - -draft
+      - conflict
+      - and: *hashi_contributor # Restrict to Hashicorp folks for now
+
+  - name: Automatic merge on approval and successful build
+    actions:
+      delete_head_branch: {}
+      queue:
+        method: rebase
```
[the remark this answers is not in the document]
ansgarm: Just out of curiosity: Why?
We currently build our changelog based on PR titles, so this wouldn't have a practical effect. Or is this to improve cosmetics or similar?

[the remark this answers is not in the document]
DanielMSchmidt: Oh right, forgot that we base it on the PR titles and not the commits

ansgarm (approved):

PR #2219, opened 2022-10-24
chore: Update Go to version 1.18 on windows workflows

mutahhir:
Follow up of: https://github.com/hashicorp/terraform-cdk/pull/2214

Upgrades the version of Go in the windows platform integration tests.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -135,7 +135,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.16.x
+          go-version: 1.18.x
       - name: Download dist
         uses: actions/download-artifact@v2
         with:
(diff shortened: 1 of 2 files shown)


ansgarm (approved):
👍

(Do you know whether our users also need Go `>=1.18` now?)

mutahhir:
> Do you know whether our users also need Go `>=1.18` now?

If they're using Go 1.16, they'll have to upgrade, but I worry about setting it as a minimum version as a patch release. Should we do that as part of version 0.14 release?

ansgarm:
Yeah, sounds like an 0.14 change 👍
Do you still know which change required that? A JSII upgrade?

mutahhir:
No, I'm unsure what caused it. Wasn't an upgrade because all PRs were experiencing this.

PR #2224, opened 2022-10-25
chore(lib): telemetry track tfc usage

### LINKED ISSUE #2206: Track which cloud backend is being used in telemetry
### Community Note

### Description

We'd like to better understand how much of our userbase is using CDKTF with TFE vs. TFC, so let's add a signal to our telemetry. In case of privacy concerns, we can just make this binary (user is either using `'app.terraform.io'` = TFC vs. anything else = TFE).

### References

From @ansgarm:

> We'd "only" need to make it part of the stack metadata which we already send for synth events here: [https://github.com/hashicorp/terraform-cdk/blob/411f47a6fef9ce829363a17496a9c1393a1a2d3c/packages/cdktf-cli/lib/synth-stack.ts#L176](https://www.google.com/url?q=https://github.com/hashicorp/terraform-cdk/blob/411f47a6fef9ce829363a17496a9c1393a1a2d3c/packages/cdktf-cli/lib/synth-stack.ts%23L176&sa=D&source=docs&ust=1666273034990508&usg=AOvVaw2pmUVvix39Uj8893MCxXa7)

ansgarm:
- chore(lib): add cloud = tfc | tfe to synth telemetry events to track usage of either
  BREAKING CHANGE: This changes the signatures of some exported methods and thus is probably a breaking change (at least for Python)
- chore(docs): update docs

Resolves #2206

### DIFF
--- a/packages/cdktf/lib/backends/cloud-backend.ts
+++ b/packages/cdktf/lib/backends/cloud-backend.ts
@@ -40,6 +40,13 @@ export class CloudBackend extends TerraformBackend {
     };
   }
 
+  public toMetadata() {
+    const cloud = [undefined, "app.terraform.io"].includes(this.props.hostname)
+      ? "tfc"
+      : "tfe";
+    return { ...super.toMetadata(), cloud };
+  }
+
   protected synthesizeAttributes(): { [name: string]: any } {
     return keysToSnakeCase({
       ...this.props,
--- a/packages/cdktf/lib/backends/cloud-backend.ts
+++ b/packages/cdktf/lib/backends/cloud-backend.ts
@@ -9,6 +9,17 @@ import { TerraformBackend } from "../terraform-backend";
 import { ValidateBinaryVersion } from "../validations";
 
 /**
+ * checks whether the given hostname belongs to tfc or (else) to tfe
+ * If no hostname is given, it will return tfc, as that's the default in backends.
+ * @param hostname e.g. app.terraform.io, app.terraform.io:80, tfe.myorg.org
+ * @returns "tfc" or "tfe"
+ */
+export function getHostNameType(hostname?: string): "tfc" | "tfe" {
+  if (!hostname) return "tfc"; // default is tfc when not passing a hostname to backends
+  return hostname.startsWith("app.terraform.io") ? "tfc" : "tfe";
+}
+
+/**
  * The Cloud Backend synthesizes a {@link https://www.terraform.io/cli/cloud/settings#the-cloud-block cloud block}.
  * The cloud block is a nested block within the top-level terraform settings block.
  * It specifies which Terraform Cloud workspaces to use for the current working directory.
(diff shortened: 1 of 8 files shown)


packages/cdktf/lib/backends/cloud-backend.ts
```diff
@@ -40,6 +40,13 @@ export class CloudBackend extends TerraformBackend {
     };
   }
 
+  public toMetadata() {
+    const cloud = [undefined, "app.terraform.io"].includes(this.props.hostname)
```
DanielMSchmidt, line 44: Are you sure the value can only be exactly `app.terraform.io` and `undefined` here? What happens if `null` is passed or `https://app.terraform.io` or `app.terraform.io/`?

ansgarm: Good call! I didn't check that..
Seems like they are invalid:
image
image

`app.terraform.io:80` is valid though 👀 Adding that one just in case, hehe.

ansgarm: added ✅

DanielMSchmidt (approved):

packages/cdktf/lib/backends/remote-backend.ts
```diff
@@ -18,6 +18,17 @@ export class RemoteBackend extends TerraformBackend {
     return keysToSnakeCase({ ...this.props });
   }
 
+  public toMetadata() {
+    const cloud = [
+      undefined,
+      "app.terraform.io",
+      "app.terraform.io:80",
```
DanielMSchmidt, line 25: Is :443 invalid?

packages/cdktf/lib/backends/cloud-backend.ts
```diff
@@ -40,6 +40,17 @@ export class CloudBackend extends TerraformBackend {
     };
   }
 
+  public toMetadata() {
+    const cloud = [
```
DanielMSchmidt, line 44: Nit: Can we dedupe this into a function?

DanielMSchmidt (commented):
Is there a reason why we don't check for a string containing a substring?

packages/cdktf/lib/backends/remote-backend.ts
ansgarm: Good catch 😁
And yeah, with all those additions a substring might be good enough.
It would catch also `app.terraform.io.ansgar.dev` or [test.ansgar.dev/app.terraform.io](https://test.ansgar.dev/app.terraform.io) though 😁 But I agree that this will be "too rare to care" 😄

packages/cdktf/lib/backends/cloud-backend.ts
ansgarm: we just might 😁
Are you fine with putting this into a function in the cloud backend and reusing it in the old remote backend?

DanielMSchmidt: yeah seems like a good idea

DanielMSchmidt (approved):

PR #2227, opened 2022-10-27

team-tf-cdk:


mutahhir:
Will take another look on the failing checks today

mutahhir:
Just noticed that this is weird. It's opening a pull request against a branch other than main. I wonder how that happened? /cc @team-tf-cdk

mutahhir:
Did some initial investigation. The python typing failures are a bit complicated to figure out, so this will take a bit more time.

PR #2239, opened 2022-10-31
feat(examples): add upcloud-server python example

kangasta:
This PR adds first example for UpCloud provider.

See also:
- [upcloud.com](https://upcloud.com/)
- [Terraform provider](https://registry.terraform.io/providers/UpCloudLtd/upcloud)
- [Pre-built cdktf provider repo](https://github.com/cdktf/cdktf-provider-upcloud)

### DIFF
--- a/examples/python/upcloud-server/main.py
+++ b/examples/python/upcloud-server/main.py
@@ -1,0 +1,76 @@
+#!/usr/bin/env python
+
+from os import getenv
+
+from constructs import Construct
+from cdktf import App, TerraformOutput, TerraformStack
+
+from imports.upcloud.provider import UpcloudProvider
+from imports.upcloud.server import Server, ServerLogin, ServerNetworkInterface, ServerTemplate
+
+
+def get_server_login(user=None):
+    """Get ServerLogin with ssh-key from PUBLIC_KEY_PATH environment variable
+
+    As a fallback, create one-time password, if PUBLIC_KEY_PATH is undefined or its value is an empty string.
+    """
+    public_key_path = getenv("PUBLIC_KEY_PATH")
+
+    if public_key_path:
+        with open(public_key_path) as f:
+            ssh_key = f.read()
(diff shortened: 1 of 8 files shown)


examples/python/upcloud-server/main.py
```diff
@@ -0,0 +1,74 @@
...
+    keys under ~/.ssh directory.
+    """
+    user_ssh_key_paths = glob(os.path.expanduser("~/.ssh/*.pub"))
+
+    if user_ssh_key_paths:
+        with open(user_ssh_key_paths[0]) as f:
+            ssh_key = f.read()
+        return ServerLogin(user=user, keys=[ssh_key])
+    else:
+        return ServerLogin(user=user, create_password=True, password_delivery="email")
```
ansgarm, line 27: I like that this function tries to give a good default for users by reading their first (and oftentimes only) SSH key – But as we synthesize our examples in CI, I'm not too fond of having code that contains read access to the SSH directory of the Github Action runner that this is running in (even if it doesn't actually do something with it, other than writing it to the output `cdk.tf.json` template).

I'd prefer this to default to the password based `ServerLogin` and only try to read from `~/.ssh` if either a boolean flag, or even the name of the ssh public key file has been passed.

ansgarm (commented):
Hi @kangasta 👋
Thank you for the PR! Looks good, except for one small thing: I'd prefer that it tries not to read the public SSH keys of our CI 😁

examples/python/upcloud-server/main.py
kangasta: Thanks for the quick review! I'll modify this part so that SSH key is only read if user gives the name of the ssh key as variable

ansgarm (approved):

ansgarm:
Closing in favour of https://github.com/hashicorp/terraform-cdk/pull/2366 (fixed lint and rebased there as I couldn't push to your branch here)

PR #2240, opened 2022-11-02
feat(cli): add providers in parallel

### LINKED ISSUE #2110: cdktf provider add should add all providers in parallel and not abort if one of them fails
### Community Note

### Description

I tried running `cdktfl provider add aws google azurerm cloudflare kubernetes --force-local` and if failed with cloudflare since that provider was not in the hashicorp namespace. all of the providers were added one after another. I think we should try to add them in parallel and at least not stop once one failed, the kubernetes provider could have been added successfully already.

DanielMSchmidt:
Closes #2110

### DIFF
--- a/packages/cdktf-cli/src/bin/cmds/handlers.ts
+++ b/packages/cdktf-cli/src/bin/cmds/handlers.ts
@@ -400,7 +400,10 @@ export async function debug(argv: any) {
   }
 }
 
-export async function providerAdd(argv: any) {
+export async function providerAdd(argv: {
+  provider: string[];
+  forceLocal?: boolean;
+}) {
   const config = CdktfConfig.read();
 
   const language = config.language;
@@ -417,21 +420,29 @@ export async function providerAdd(argv: any) {
     config.projectDirectory
   );
 
-  let needsGet = false;
-
-  for (const provider of argv.provider) {
+  const providerAddPromises = argv.provider.map(async (provider: string) => {
     const constraint = ProviderConstraint.fromConfigEntry(provider);
 
     if (argv.forceLocal) {
-      needsGet = true;
       await manager.addLocalProvider(constraint);
+      return true;
     } else {
       const { addedLocalProvider } = await manager.addProvider(constraint);
-      if (addedLocalProvider) {
-        needsGet = true;
-      }
+      return addedLocalProvider;
     }
-  }
+  });
+
+  // We need to use allSettled, because in case of an error we don't want the first one, but all of them
+  const results = await Promise.allSettled(providerAddPromises);
+  const errors = results.reduce((errorMessage, item) => {
+    if (item.status === "rejected") {
+      return `${errorMessage}\n\n- ${item.reason.message}`;
+    }
+    return errorMessage;
+  }, "");
+  const needsGet = results.some(
+    (item) => item.status === "fulfilled" && item.value === true
+  );
 
   if (needsGet) {
     console.log(
@@ -442,6 +453,10 @@ export async function providerAdd(argv: any) {
       output: config.codeMakerOutput,
       parallelism: 1,
     });
+  }
+
+  if (errors !== "") {
+    throw Errors.Usage(`Error adding one or more providers: \n ${errors}`);
   }
 }
 
--- a/packages/cdktf-cli/src/bin/cmds/handlers.ts
+++ b/packages/cdktf-cli/src/bin/cmds/handlers.ts
@@ -49,6 +49,7 @@ import {
 } from "../../lib/dependencies/dependency-manager";
 import { CdktfConfig, ProviderDependencySpec } from "../../lib/cdktf-config";
 import { logger } from "../../lib/logging";
+import { PackageConstraint } from "../../lib/dependencies/dependency-manager";
 
 const chalkColour = new chalk.Instance();
 const config = cfg.readConfigSync();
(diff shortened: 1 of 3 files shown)


packages/cdktf-cli/src/bin/cmds/handlers.ts
```diff
@@ -417,21 +420,29 @@ export async function providerAdd(argv: any) {
...
     if (argv.forceLocal) {
-      needsGet = true;
       await manager.addLocalProvider(constraint);
+      return true;
     } else {
       const { addedLocalProvider } = await manager.addProvider(constraint);
-      if (addedLocalProvider) {
-        needsGet = true;
-      }
+      return addedLocalProvider;
```
ansgarm, line 431: As discussed offline, we should confirm that we don't run into race conditions like e.g. running `npm install` twice at the same time.

packages/cdktf-cli/src/bin/cmds/handlers.ts
```diff
@@ -440,21 +445,41 @@ export async function providerAdd(argv: {
     }
     return errorMessage;
   }, "");
+
   const needsGet = results.some(
-    (item) => item.status === "fulfilled" && item.value === true
+    (item) =>
+      item.status === "fulfilled" && item.value.addedLocalProvider === true
   );
+  const packagesToInstall = results.reduce((packages, item) => {
```
ansgarm, line 453: nit: I think changing the `reduce` to a chained `filter` and `flatMap` would make this more readable 😁
#no-reduce 😄

packages/cdktf-cli/src/bin/cmds/handlers.ts
```diff
@@ -440,21 +445,41 @@ export async function providerAdd(argv: {
...
+    if (item.status === "fulfilled" && item.value.prebuiltProviderToAdd) {
+      return [...packages, item.value.prebuiltProviderToAdd];
+    }
+    return packages;
+  }, [] as PackageConstraint[]);
 
+  const sideEffects: Promise<unknown>[] = [];
```
ansgarm, line 460: hehe, this is kinda cool 😁

packages/cdktf-cli/src/lib/dependencies/package-manager.ts
```diff
@@ -12,6 +12,11 @@ import fetch from "node-fetch";
 import * as z from "zod";
 import { logger } from "../logging";
 
+export type PackageConstraint = {
+  name: string;
+  version?: string;
+};
```
ansgarm, line 18: love it!

packages/cdktf-cli/src/lib/dependencies/package-manager.ts
```diff
@@ -141,22 +144,25 @@ class NodePackageManager extends PackageManager {
...
 
     // Install exact version
     // Yarn: https://classic.yarnpkg.com/lang/en/docs/cli/add/#toc-yarn-add-exact-e
     // Npm: https://docs.npmjs.com/cli/v8/commands/npm-install#save-exact
     args.push("-E");
 
-    console.log(
-      `Installing package ${packageName} @ ${packageVersion} using ${command}.`
+    logger.debug(
+      `Installing package ${packages
```
ansgarm, line 158:
```suggestion
      `Installing package(s) ${packages
```

packages/cdktf-cli/src/lib/dependencies/package-manager.ts
```diff
@@ -238,30 +244,35 @@ class PythonPackageManager extends PackageManager {
...
+          },
+          stdio: ["inherit", 1, 1],
+        }
+      );
 
       console.log("Package installed.");
     } else {
       console.log(
-        `Installing package ${packageName} @ ${packageVersion} using pip.`
+        `Installing package ${packages
```
ansgarm, line 273:
```suggestion
        `Installing package(s) ${packages
```

packages/cdktf-cli/src/lib/dependencies/dependency-manager.ts
```diff
@@ -290,16 +293,21 @@ export class DependencyManager {
     return packageVersion;
   }
 
-  async addPrebuiltProvider(constraint: ProviderConstraint) {
+  async addPrebuiltProvider(
```
ansgarm, line 296: We should rename this method, as it is not adding anything anymore, but rather returns a `PackageConstraint` that can be added later on.

packages/cdktf-cli/src/lib/dependencies/dependency-manager.ts
```diff
@@ -137,12 +137,13 @@ export class DependencyManager {
     );
   }
 
-  async addProvider(
-    constraint: ProviderConstraint
-  ): Promise<{ addedLocalProvider: boolean }> {
+  async addProvider(constraint: ProviderConstraint): Promise<{
+    addedLocalProvider: boolean;
+    prebuiltProviderToAdd?: PackageConstraint;
```
ansgarm, line 142: maybe name this `prebuiltProviderToInstall` to denote that this is a dependency in the target language's depedency manager? "Add" seems a bit overused in this context.

packages/cdktf-cli/src/lib/dependencies/dependency-manager.ts
```diff
@@ -137,12 +137,13 @@ export class DependencyManager {
     );
   }
 
-  async addProvider(
-    constraint: ProviderConstraint
-  ): Promise<{ addedLocalProvider: boolean }> {
+  async addProvider(constraint: ProviderConstraint): Promise<{
```
ansgarm, line 140: nit: The method is a bit leaky now, as it either has no side-effect (when the provider ends up being available as pre-built provider) or it does have a side-effect (when it's added locally to the `cdktf.json`).
I could see us changing the method name to try to reflect the non-side-effect nature or to maybe refactor this in a way that changes to the local-adding bit to have no side-effect, too, and also return "installation instructions" aka a constraint that can be added later on.

ansgarm: While it is pretty rare – in theory (e.g. with a slow filesystem) reading the local providers, filtering them, adding another local provider, and writing them might be a different race condition ([source](https://github.com/hashicorp/terraform-cdk/blob/950c919b901157673f5afda47dc0e4b7ecd3f231/packages/cdktf-cli/src/lib/dependencies/dependency-manager.ts#L366)). Maybe refactoring that bit in a similar way to the pre-built provider packages would get rid of that race condition while solving my nit about this inconsistency / about the methods name 🤔

DanielMSchmidt:
Closing this since fixing all the small race conditions makes the code harder to read and is not worth the effort for the little performance improvement we get out of it.

Issue #2233, opened 2022-10-31
Terraform Cloud Bearer token plain text printed in log when cdktf command fails


daniel-laszlo:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### cdktf & Language Versions

Python 3.9
CDKTF 0.13.2

### Affected Resource(s)

`cdktf diff`
`cdktf deploy`

### Debug Output

```
Creating Terraform Cloud configuration version
Created Terraform Cloud configuration version cv-XXXXXXXXXXXXX
Zipping up the directory cdktf.out/stacks/STACK
Uploading the directory to Terraform Cloud
Uploaded the directory to Terraform Cloud
Waiting for configuration version to become ready...
Creating speculative Terraform Cloud run
Created speculative Terraform Cloud run: https://app.terraform.io/app/ORGANIZATION/workspaces/WORKSPACE_NAME/runs/run-XXXXXXXXXXXXXXXX
Speculative Terraform Cloud run done
Getting plan output
Deploying Terraform Cloud run


1 Stack deploying     0 Stacks done     0 Stacks waiting
[2022-10-31T14:34:34.126] [ERROR] default - Error in Deploy: {"message":"...","name":"Error","stack":"...","config":{"url":"...","method":"post","data":"{}",
"headers":{"Authorization":"Bearer TOKEN",  <--- TOKEN is in plain text!!
"Accept":"application/json","Content-Type":"application/vnd.api+json","User-Agent":"terraform-cloud/1.15.0","Content-Length":2},"baseURL":"https://app.terraform.io/api/v2","transformRequest":[null],"transformResponse":[null],"timeout":0,"xsrfCookieName":"XSRF-TOKEN","xsrfHeaderName":"X-XSRF-TOKEN","maxContentLength":-1,"maxBodyLength":-1,"transitional":{"silentJSONParsing":true,"forcedJSONParsing":true,"clarifyTimeoutError":false}}}
```

### Expected Behavior

No token is printed when there is an error.

### Actual Behavior

[...]

PR #2241, opened 2022-11-02
fix: Redact authorization header when error logging from TFC/TFE

mutahhir:
Closes: https://github.com/hashicorp/terraform-cdk/issues/2233

### DIFF
--- a/packages/cdktf-cli/src/lib/models/terraform-cloud.ts
+++ b/packages/cdktf-cli/src/lib/models/terraform-cloud.ts
@@ -112,11 +112,11 @@ export function LogBetterErrorAndThrow(name: string, e: AxiosError) {
       throw new Error(
         `${name}: Request to Terraform Cloud failed with status ${
           e.response.status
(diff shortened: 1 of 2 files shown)


packages/cdktf-cli/src/lib/models/terraform-cloud.ts
```diff
@@ -64,6 +81,45 @@ const wait = (ms = 1000) => {
...
+  }
+
+  return JSON.stringify(error, function (key: string, value: string) {
+    if (EXCLUDED_KEYS.includes(key)) return "REDACTED";
+    return value;
+  });
+}
+
+// Exporting for testing purposes only
+export function LogBetterErrorAndThrow(name: string, e: AxiosError) {
```
ansgarm, line 104: nit:
```suggestion
export function logBetterErrorAndThrow(name: string, e: AxiosError) {
```

ansgarm (approved):

Issue #2232, opened 2022-10-31
cdktf deploy auto-approve error when ran on TFC workspace with cost estimation enabled


daniel-laszlo:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### cdktf & Language Versions

Python 3.9
CDKTF 0.13.2
Terraform Cloud

### Affected Resource(s)

`cdktf deploy STACK --auto-approve`

### Debug Output

```
Creating Terraform Cloud configuration version
Created Terraform Cloud configuration version cv-XXXXXXXXXXXXX
Zipping up the directory cdktf.out/stacks/STACK
Uploading the directory to Terraform Cloud
Uploaded the directory to Terraform Cloud
Waiting for configuration version to become ready...
Creating speculative Terraform Cloud run
Created speculative Terraform Cloud run: https://app.terraform.io/app/ORGANIZATION/workspaces/WORKSPACE_NAME/runs/run-XXXXXXXXXXXXXXXX
Speculative Terraform Cloud run done
Getting plan output
Deploying Terraform Cloud run


1 Stack deploying     0 Stacks done     0 Stacks waiting
[...]

PR #2243, opened 2022-11-02
fix: Transition to planned state after cost estimation and policy checks with TFC/TFE

daniel-laszlo:
Closes https://github.com/hashicorp/terraform-cdk/issues/2232

### DIFF
--- a/packages/cdktf-cli/src/lib/models/terraform-cloud.ts
+++ b/packages/cdktf-cli/src/lib/models/terraform-cloud.ts
@@ -298,7 +298,7 @@ export class TerraformCloud implements Terraform {
     const url = `https://${this.hostname}/app/${this.organizationName}/workspaces/${this.workspaceName}/runs/${result.id}`;
     sendLog(`Created speculative Terraform Cloud run: ${url}`);
 
-    const pendingStates = ["pending", "plan_queued", "planning"];
+    const pendingStates = ["pending", "plan_queued", "planning", "cost_estimating", "policy_checking"];
 
     while (pendingStates.includes(result.attributes.status)) {
       result = await this.update(this.client, result.id, "plan");


DanielMSchmidt (approved):

packages/cdktf-cli/src/lib/models/terraform-cloud.ts
```diff
@@ -298,7 +298,7 @@ export class TerraformCloud implements Terraform {
     const url = `https://${this.hostname}/app/${this.organizationName}/workspaces/${this.workspaceName}/runs/${result.id}`;
     sendLog(`Created speculative Terraform Cloud run: ${url}`);
 
-    const pendingStates = ["pending", "plan_queued", "planning"];
+    const pendingStates = ["pending", "plan_queued", "planning", "cost_estimating", "policy_checking"];
```
ansgarm, line 301: I'm wondering whether we'd need to add other states ([docs](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/run#run-states)) as pending here, too.
For example `pre_plan_running` also exists and might happen.

Related to this: Do we also need to add `cost_estimated`? I imagine that our current implementation will only rarely catch this one, as TFC might directly jump to the next state after it. So that might be a race condition waiting to happen.. 😅

Does anyone know whether we have a flowchart somewhere that shows the order in which these states happen? (E.g. `planned` might always be the final one when planning, and we could just invert the logic here and wait for that state to be reached?)

mutahhir: I think it might be hard to get the 'final' state, since it depends on your TFC plan. Cost Estimation is not available for the cheaper tiers I think (though the [pricing page](https://www.hashicorp.com/products/terraform/pricing) doesn't elaborate).

mutahhir (approved):

packages/cdktf-cli/src/lib/models/terraform-cloud.ts
mutahhir: Though, re-reading your comment, I agree. Having `planned` be the state we care about might be good enough.

daniel-laszlo: - Adding `pre_plan_running` sounds like a good idea
- `cost_estimated` can be a final state in the sense that if there are no policy checks, an apply run can be created on TFC
- `planned` state only shows that planning has finished, after it can come the cost estimation and policy checks.

Here is an example order of events:
```
plan-queueable-at: | 2022-11-02T14:12:03+00:00,
queuing-at: | 2022-11-02T14:12:04+00:00,
plan-queued-at: | 2022-11-02T14:12:04+00:00,
planning-at: | 2022-11-02T14:12:10+00:00,
cost-estimating-at: | 2022-11-02T14:13:48+00:00
planned-at: | 2022-11-02T14:13:48+00:00,
cost-estimated-at: | 2022-11-02T14:13:57+00:00,
```

I don't know if you have access to TFE/TFC codebase, but maybe it's a good place to search for logic on when a run can be applied.

daniel-laszlo: Ok I think I got a good approach: We need to check the `data.attributes.actions.is-confirmable` flag from the [GET /runs/:run_id](https://developer.hashicorp.com/terraform/cloud-docs/api-docs/run#get-run-details) API. Is this something you could take over @mutahhir / @ansgarm ?

DanielMSchmidt:
I merged the rebased version of this: #2360

PR #2257, opened 2022-11-06
fix(cli): do not check for version when DISABLE_VERSION_CHECK

vincentbernat:
Without it, "npm view cdktf-cli version" is invoked.


ansgarm (approved):
Good catch! Thanks @vincentbernat 👍

PR #2259, opened 2022-11-08
chore: adding mgarrell777 to docs codeowners to be notified for docs changes

ansgarm:
as requested by @laurapacilio


mutahhir:
Would it be an idea to create a docs team and give them ownership? That way Github can also pick one reviewer at random.

Issue #2229, opened 2022-10-28
`cdktf provider list` command


ansgarm:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

When working with provider bindings it is easy to lose track of which are installed as pre-built packages and which are defined in the `cdktf.json` config. It gets even harder if one tries to find out the versions of the underlying Terraform providers.
To help with this, let's introduce a new command: `cdktf provider list`.

The output of that command _could_ look like this:
```
> cdktf provider list
|            provider name            | provider version |  cdktf  | constraint |               package               |
|-------------------------------------|------------------|---------|------------|-------------------------------------|
| aws                                 |            4.0.1 | ^0.13.0 |            | @cdktf/provider-aws                 |
| random                              |            3.2.1 | ^0.13.0 |            | @cdktf/provider-random              |
| kreuzwerker/docker                  |            2.3.8 | ^0.13.0 |            | @cdktf/provider-docker              |
| my.registry.co/acme/privateprovider |            0.1.0 | 0.13.2  | ~>0.1.0    | .gen/providers/acme/privateprovider |
```
(just a rough draft, feel free to adjust when implementing; colors might also be used but shouldn't solely relied on; column headers can be adjusted, reordered, etc.)

#### provider name
The name of the underlying Terraform provider, using the simplified version:
https://github.com/hashicorp/terraform-cdk/blob/39241c4640da6116485d8fb6ea243f596080dc36/packages/cdktf-cli/src/lib/dependencies/dependency-manager.ts#L95-L104

#### provider version
the exact version of the Terraform provider used to generate the bindings. Can be read from the `version.json` ([example](https://github.com/cdktf/cdktf-provider-null/blob/main/src/version.json)) or the `package.json` from the NPM Registry:
https://github.com/hashicorp/terraform-cdk/blob/39241c4640da6116485d8fb6ea243f596080dc36/packages/cdktf-cli/src/lib/dependencies/prebuilt-providers.ts#L178

#### cdktf
[...]

PR #2260, opened 2022-11-08
feat: Add `cdktf provider list` command

mutahhir:
Closes: #2229

WIP Images of how it looks like:
image


packages/cdktf-cli/src/lib/dependencies/dependency-manager.ts
```diff
@@ -384,4 +385,79 @@ export class DependencyManager {
...
+    const regexes = {
+      [Language.GO]: /github.com\/(cdktf|hashicorp)\/cdktf-provider-(.+)-go\//i,
+      [Language.TYPESCRIPT]: /(.+)/i,
+      [Language.CSHARP]: /HashiCorp\.Cdktf\.Provider\.(.+)/i,
+      [Language.JAVA]: /com\.hashicorp\.cdktf-provider-(.+)/i,
+      [Language.PYTHON]: /cdktf-cdktf-provider-(.+)/i,
+    };
+    const regex = regexes[this.targetLanguage];
+    if (!regex) {
+      throw new Error("Language not supported for pre-built providers");
```
ansgarm, line 404: I think we can use one of our [custom errors](https://github.com/hashicorp/terraform-cdk/blob/86e8afa3d35fdcfeeca73ba6a70872b1fb06a8b1/packages/cdktf-cli/src/lib/errors.ts#L40) here instead

mutahhir: Nice! Didn't know about those.

mutahhir:
Rebasing PR to be based on: https://github.com/hashicorp/terraform-cdk/pull/2265 since I need to use the versions.json file from here on.

ansgarm:
> Rebasing PR to be based on: #2265 since I need to use the versions.json file from here on.

It could also conflict with https://github.com/hashicorp/terraform-cdk/pull/2266 (haven't checked, though)

PR #2261, opened 2022-11-08
chore(lib): Add deprecation notices to backends that were deprecated in Terraform v1.2.3 and removed in TF 1.3

### LINKED ISSUE #2244: Mark backends that were removed in Terraform 1.3 as deprecated in the CDKTF docs
### Community Note

### Description

Several remote backends were removed in Terraform 1.3 but are currently still supported in CDKTF because we support Terraform 1.1+. We should at minimum mark these backends as deprecated in our own documentation and in the docblocks to discourage new use of these backends in case we require a higher version of Terraform for future releases with which these backends are no longer compatible.

The main page affected by this is: https://developer.hashicorp.com/terraform/cdktf/concepts/remote-backends#supported-backends

### References

From the Terraform core changelog for 1.3:

> The following backends, which were deprecated in v1.2.3, have now been removed: `artifactory`, `etcd`, `etcdv3`, `manta`, `swift`. The legacy backend name `azure` has also been removed, because the current Azure backend is named `azurerm`. (https://github.com/hashicorp/terraform/issues/31711)

ansgarm:
Marks our backend constructs and their options as deprecated.
Closes #2244

### DIFF
--- a/packages/cdktf/lib/backends/artifactory-backend.ts
+++ b/packages/cdktf/lib/backends/artifactory-backend.ts
@@ -9,7 +9,7 @@ import {
 } from "../terraform-remote-state";
 
 /**
- * @deprecated the artifactory backend has been deprecated in Terraform v1.2.3 and was removed in Terraform v1.3
+ * @deprecated CDK for Terraform no longer supports the artifactory backend. Terraform deprecated artifactory in v1.2.3 and removed it in v1.3.
  */
 export class ArtifactoryBackend extends TerraformBackend {
   constructor(
@@ -33,7 +33,7 @@ export class ArtifactoryBackend extends TerraformBackend {
 }
 
 /**
- * @deprecated the artifactory backend has been deprecated in Terraform v1.2.3 and was removed in Terraform v1.3
+ * @deprecated CDK for Terraform no longer supports the artifactory backend. Terraform deprecated artifactory in v1.2.3 and removed it in v1.3.
  */
 export class DataTerraformRemoteStateArtifactory extends TerraformRemoteState {
   constructor(
@@ -57,7 +57,7 @@ export class DataTerraformRemoteStateArtifactory extends TerraformRemoteState {
  * Read more about this backend in the Terraform docs:
  * https://www.terraform.io/language/settings/backends/artifactory
  *
- * @deprecated the artifactory backend has been deprecated in Terraform v1.2.3 and was removed in Terraform v1.3
+ * @deprecated CDK for Terraform no longer supports the artifactory backend. Terraform deprecated artifactory in v1.2.3 and removed it in v1.3.
  */
 export interface ArtifactoryBackendProps {
   /**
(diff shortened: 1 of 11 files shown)


DanielMSchmidt (approved):

laurapacilio (commented):
@ansgarm we're also going to want to add something on this page:
https://developer.hashicorp.com/terraform/cdktf/concepts/remote-backends

My opinion is that we remove the backends that are no longer supported from the "Supported Backends" section and instead add a note that says something like: 

"-> **Note:** CDK for Terraform v0.14 no longer supports the artifactory, etcd, etcdv3, manta, and swift backends. Terraform removed these backends in v1.3. For migration paths from these removed backends, refer to [Upgrading to Terraform v1.3](https://developer.hashicorp.com/terraform/language/v1.3.x/upgrade-guides)."

Feel free to edit this to make sure it matches what makes sense for CDK users, but I adapted it from this page in the Terraform Language docs: https://developer.hashicorp.com/terraform/language/settings/backends/configuration

packages/cdktf/lib/backends/artifactory-backend.ts
```diff
@@ -8,7 +8,9 @@ import {
   DataTerraformRemoteStateConfig,
 } from "../terraform-remote-state";
 
-// eslint-disable-next-line jsdoc/require-jsdoc
+/**
+ * @deprecated the artifactory backend has been deprecated in Terraform v1.2.3 and was removed in Terraform v1.3
```
laurapacilio, line 12: @ansgarm What if we did something like this? I was thinking maybe we should differentiate between our product and Terraform itself. But let me know if this feels like too much. If you like it, maybe we could do this for all of them?

```suggestion
 * @deprecated CDK for Terraform no longer supports the artifactory backend. Terraform deprecated artifactory in v1.2.3 and removed it in v1.3. 
```

ansgarm: Sounds good! I like it 👍

website/docs/cdktf/concepts/remote-backends.mdx
```diff
@@ -215,10 +199,8 @@ In addition to Terraform Cloud, Terraform and CDKTF support the following backen
   ```typescript
   new S3Backend(stack, {...});
   ```
-- [swift](/language/v1.2.x/settings/backends/swift)
-  ```typescript
-  new SwiftBackend(stack, {...});
-  ```
+
+->**Note**: CDK for Terraform v0.14 deprecated the artifactory, etcd, etcdv3, manta, and swift backends. Terraform removed these backends in v1.3. For migration paths from these removed backends, refer to [Upgrading to Terraform v1.3](/language/v1.3.x/upgrade-guides).
```
ansgarm, line 203: @laurapacilio took the sentence you proposed and only changed "removed" to "deprecated" as we did not yet remove them. Is this fine, or did it make things unclear?

laurapacilio: I think this is good! Folks know what deprecated means, so this works for me :)

laurapacilio (approved):

PR #2265, opened 2022-11-09
feat(provider-generator): publish versions.json file for all languages

mutahhir:
Currently we only publish the `versions.json` file for Typescript projects only. For other languages, there's currently no way to ascertain the version of the provider the bindings are built against. 

This is also important for the upcoming `cdktf provider list` feature, as we would like to present to the user the provider version for the locally generated binding.

 Caveat: This is me hacking around to find the simplest solution for getting the versions.json file published. If there's a much smarter way of doing this, please let me know. 🙏🏽


ansgarm:
Is this PR supposed to target only the case of locally build provider bindings or pre-built ones too?

mutahhir:
> Is this PR supposed to target only the case of locally built provider bindings or pre-built ones too?

From what I understood, we generate the versions.json file when using `cdktf provider add` or `get`. Can you explain a bit more about the pre-built situation?

PR #2270, opened 2022-11-11
feat(cli): Add cdktf provider list command

mutahhir:
> **Note**
> Duplicate of https://github.com/hashicorp/terraform-cdk/pull/2260. 
> My Git foo was not good enough to get out of the hole of multiple rebases 😢 

Closes: #2229

WIP Images of how it looks like:
image


packages/cdktf-cli/src/lib/local-provider-versions.ts
```diff
@@ -0,0 +1,80 @@
...
+    } catch (e) {
+      // If we cannot read the versions.json file, it may not exist
+      logger.debug("versions.json file reading error: ", e);
+      this.versions = undefined;
+      return;
+    }
+
+    let providerVersions;
+    try {
+      providerVersions = JSON.parse(versionsJson) as LocalProviderVersions;
```
DanielMSchmidt, line 34: Wouldn't a record be a better type, `LocalProviderVersions` is a class so this cast is definitely misleading.
```suggestion
      providerVersions = JSON.parse(versionsJson) as Record<string, string>;
```

packages/cdktf-cli/src/lib/local-provider-versions.ts
```diff
@@ -0,0 +1,80 @@
...
+    } catch (e) {
+      throw Errors.External(
+        "versions.json file is malformed. The root must be a JSON object."
+      );
+    }
+
+    this.versions = Object.fromEntries(
```
DanielMSchmidt, line 41: Fun fact: @ansgarm is going to love this :D I seriously don't understand how this is more readable than

```ts
Object.entries(providerVersions).reduce((obj, [providerFqn, versions]) => ({
   ...obj,
   [providerFqn
            .replace("registry.terraform.io/", "")
            .replace("hashicorp/", "")
            .toLowerCase()]: versions
}), {})
```

DanielMSchmidt (approved):
Looks good 👍 I'm ok with shipping this as is, but would love to see the performance related stuff being done either here or as a follow up

packages/cdktf-cli/src/bin/cmds/handlers.ts
```diff
@@ -14,6 +14,7 @@ import {
   TerraformModuleConstraint,
   TerraformProviderConstraint,
 } from "@cdktf/provider-generator";
+import Table from "cli-table";
```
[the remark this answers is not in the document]
ansgarm: I have no opinion on this yet – I'd like to get rid of React (as it seems to complicate things when working with hooks, etc.) but I suspect that building an interactive UI without it might be harder – so that would be my only reason to keep `ink`. Ah, well, error handling is also something were we make use of `ink` features iirc.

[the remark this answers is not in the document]
mutahhir: Ah, good one. I haven't used ink before, but I'll try to take a closer look at it.

packages/cdktf-cli/src/lib/local-provider-versions.ts
mutahhir: Haha, I've been told to stay away from `reduce` too 😅 . I haven't used fromEntries before, and I agree I don't like the wrapper, but honestly: both approaches are ugly. So, I'll do the one that at least one person prefers 😇

packages/cdktf-cli/src/lib/local-provider-versions.ts
```diff
@@ -0,0 +1,80 @@
...
+      Object.entries(providerVersions).map(([providerFqn, versions]) => {
+        return [
+          providerFqn
+            .replace("registry.terraform.io/", "")
+            .replace("hashicorp/", "")
+            .toLowerCase(),
```
ansgarm, line 47: nit: While I don't think that this changes – we do have constants for those:
https://github.com/hashicorp/terraform-cdk/blob/4f03ceb074c9710a406da5cd90d86c927ce6e5ca/packages/cdktf-cli/src/lib/dependencies/dependency-manager.ts#L20

ansgarm (approved):

packages/cdktf-cli/src/lib/local-provider-versions.ts
ansgarm: Haha, now you're stuck in some kind of "who's your favorite child" situation because Daniel and I each like the opposite one 😂

ansgarm: I could see me 🤝 on the "both are ugly" though 😁

mutahhir: oh no. Time to go back to lodash 😎

packages/cdktf-cli/src/bin/cmds/handlers.ts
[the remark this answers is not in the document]
DanielMSchmidt: Here is an ink table component: https://github.com/maticzav/ink-table

team-tf-cdk:

team-tf-cdk:

team-tf-cdk:

team-tf-cdk:

PR #2278, opened 2022-11-11
chore(docs): Add CDK for Terraform to page titles

laurapacilio:
Currently when folks google CDK for Terraform Concepts, it's impossible to tell whether some of the results are for regular Terraform or CDK for Terraform. This is because our content has a lot of similar concepts to Terraform (like providers) and every page is automatically appended with "- Terraform | HashiCorp Developer"

This makes it more difficult for users to be sure that they are getting the right content when they click on a page. This PR adds "CDK for Terraform" to every page title to make it clear which content relates to CDKTF specifically.

## Examples

A provider page - this is for CDK for Terraform, but the title makes it look like it's for regular Terraform only.

Screen Shot 2022-11-11 at 3 10 27 PM

Terraform Content uses "-Configuration Language" so we modeled the "- CDK for Terraform" after this for consistency.

Screen Shot 2022-11-11 at 3 15 20 PM

Search for **Terraform Functions** yields two pages - one for configuration language and one for cdktf. But by the titles, it's incredibly difficult to know what's different about the second page (it's the one for CDKTF).

Screen Shot 2022-11-11 at 3 17 28 PM

### DIFF
--- a/website/docs/cdktf/create-and-deploy/aws-adapter.mdx
+++ b/website/docs/cdktf/create-and-deploy/aws-adapter.mdx
@@ -1,5 +1,5 @@
 ---
-page_title: "[technical preview] AWS Adapter"
+page_title: [technical preview] AWS Adapter | CDK for Terraform
 description: "[technical preview] Use AWS CDK (v2) constructs in CDK for Terraform projects."
 ---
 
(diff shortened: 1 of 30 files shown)


ansgarm (approved):

website/docs/cdktf/create-and-deploy/aws-adapter.mdx
```diff
@@ -1,5 +1,5 @@
 ---
-page_title: "[technical preview] AWS Adapter"
+page_title: [technical preview] AWS Adapter - CDK for Terraform
```
ansgarm, line 2: Vercel fails without these (as the `[` is interpreted by yaml as opening an array)
```suggestion
page_title: "[technical preview] AWS Adapter - CDK for Terraform"
```

PR #2283, opened 2022-11-14
chore: add workflow to update our github actions

DanielMSchmidt:
We don't keep them up to date manually, leaving us open to security risks and to an unreasonable amount of deprecation warnings in CI


ansgarm (approved):
Is there a way we can detect breaking changes to workflows that don't run on PRs?

In the worst case this would introduce an update that fails our release workflow and we'd only notice it the next time we try to release 😅 (Luckily the pre-release workflow runs on `main` and is almost identical to the release workflow 👍)

PR #2284, opened 2022-11-14
feat(lib): Introduce Token.nullValue() that allows to pass null to Terraform configurations in languages other than TypeScript

### LINKED ISSUE #2221: Can't pass `null` value ("`None`") when using Python
### Community Note

There's a workaround for https://github.com/hashicorp/terraform-cdk/issues/2220 which depends on an upstream issue, unfortunately there seems to be no way to set `null` in a Python CDKTF project (not even using overrides).

### cdktf & Language Versions

* 0.13.1

### Affected Resource(s)

value `null` ([Terraform JSON config](https://developer.hashicorp.com/terraform/language/syntax/json#expression-mapping) supports it)

#### Python (invalid config, fails)
```jsonc
{
  // ...
  "locals": {
  },
  "output": {
    "out": {
    }
  },
  // ...
}
```

#### TypeScript (works)
```jsonc
{
  // ...
  "data": {
    "null_data_source": {
      "null2": {
        // ...
        "inputs": {
          "x": null
        }
      }
    }
  },
  "locals": {
    "null": null
  },
  "output": {
    "out": {
      "value": null
    }
  },
  // ...
}
```

### Expected Behavior

There should be a way to set a `null` value when using Python (and all other languages as well, they haven't been tested yet)

### Actual Behavior

Python example produces invalid config, no way to get a `null` value into the output (see invalid output above).

#### Python
```python
    out = TerraformOutput(self, "out", value="to be null");
    out.add_override("value", None); # results in invalid config
    out.add_override("value", Token.as_string(None)); # same as above
        
    TerraformLocal(self, "local", None); # produces invalid config
```

#### TypeScript

```ts
    new DataNullDataSource(this, "null2", { inputs: { x: Token.asString(null) } });
    new TerraformLocal(this, "null", null);

    const out = new TerraformOutput(this, "out", {
      value: "to be null", // setting null directly does not work
    })
    out.addOverride("value", null);
```

### References

- #2220

ansgarm:
Resolves #2221


packages/cdktf/lib/tokens/token.ts
```diff
@@ -171,6 +171,19 @@ export class Token {
   public static asAny(value: any): IResolvable {
     return Tokenization.isResolvable(value) ? value : new Intrinsic(value);
   }
+
+  /**
+   * Return a Token containing a `null` value
+   * @returns a Token resolving to `null` as understood by Terraform
+   * Note: This is different than `undefined`, `nil`, `None` or similar
```
DanielMSchmidt, line 178: The note is not included in the generated documentaton, is this by choice?

DanielMSchmidt (approved):

packages/cdktf/lib/tokens/token.ts
ansgarm: Nope, by accident 😅 I'll re-order

PR #2292, opened 2022-11-15
feat(lib): expose operators on their own API class

DanielMSchmidt:


packages/cdktf/lib/terraform-operators.ts
```diff
@@ -0,0 +1,113 @@
+import { Expression, OperatorExpression } from "./tfExpression";
+import { IResolvable } from "./tokens/resolvable";
+
+/**
+ * This class contains static functions for all arithmetical and logical operators
+ * in the Terraform configutation language.
+ */
+export class Op {
+  /**
+   * Renders !expr
```
ansgarm, line 10: nit: shall we match the name of the argument?
```suggestion
   * Renders !expression
```

packages/cdktf/lib/terraform-operators.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  /**
+   * Renders expr1 + expr2
+   */
+  public static add(left: Expression, right: Expression): IResolvable {
+    return new OperatorExpression("+", left, right);
+  }
+
+  /**
+   * Renders expr1 - expr2
```
ansgarm, line 52:
```suggestion
   * Renders left - right
```

ansgarm: same for other comments

ansgarm (approved):
Nice!
Just a small set of nits regarding the code comments 😇

mutahhir (approved):
Nice! ❤️

laurapacilio (approved):
I assume the API docs are autogenerated? So I'm assuming it doesn't make sense to really review them here :). Otherwise, hooray - nice work!

ansgarm:
> I assume the API docs are autogenerated?

@laurapacilio: They are based off of the comments in this file: https://github.com/hashicorp/terraform-cdk/blob/main/packages/cdktf/lib/terraform-operators.ts

PR #2297, opened 2022-11-16
feat(docs): document Token.nullValue()

ansgarm:


website/docs/cdktf/concepts/tokens.mdx
```diff
@@ -59,3 +60,33 @@ Later in synthesis, CDKTF will resolve the token to `${module.<module id>.public
 ```
 
 Refer to the [AWS CDK documentation](https://docs.aws.amazon.com/cdk/latest/guide/tokens.html) for more detailed information about tokens.
+
+### Using Terraform's `null` value
+Some edge cases require passing the Terraform `null` value to for example an attribute of a resource.
```
ansgarm, line 65: @laurapacilio, what do you think?

website/docs/cdktf/concepts/tokens.mdx
```diff
@@ -59,3 +60,33 @@ Later in synthesis, CDKTF will resolve the token to `${module.<module id>.public
...
+
+<CodeTabs>
```
DanielMSchmidt, line 67: We should add these to the documentation examples, but we can do it in a follow up

DanielMSchmidt (approved):

website/docs/cdktf/concepts/tokens.mdx
ansgarm: Yeah, I agree – I copied them from the edge provider tests for now 😁

ansgarm: Follow-up: https://github.com/hashicorp/terraform-cdk/issues/2298
Might be easiest to do this when we have documentation examples for all languages in place

Issue #1916, opened 2022-07-08
Add -no-color option


nekhebit:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description
Please add the option `-no-color` to cdk. 

A lot of users run cdk in a context where the output is not rendered by the system, but instead used for further computation or redirected to other sinks, that cannot interpret terminal formatting. The additional formatting strings makes log reading very difficult. 
Thank you very much for your work and for your attention. :+1: 

### References
https://www.terraform.io/cli/commands/apply#no-color

ansgarm:
Hi @nekhebit!

Sounds like a good flag to add. In the meantime you could use the environment variable `FORCE_COLOR=0` which should disable colors ([supported](https://github.com/chalk/chalk#supportscolor) by the underlying lib we use).

Independently from adding support for the `--no-color` flag, we should at least document `FORCE_COLOR` in the [environment variables docs](https://www.terraform.io/cdktf/create-and-deploy/environment-variables).

nekhebit:
Hi @ansgarm! :)
Thanks for the fast reply! 

I did not know about the flag, I will give it a try. Is there a way to diminish the verbosity of the output? For example, to get only the error back, or to get the responses as a json so that it is easier to filter?

ansgarm:
Hi @nekhebit,

for which commands do you want such output? And could you explain a bit more about your underlying use case for this?

nekhebit:
The output given by cdk is really verbose. It is good for a human to go through it and understand what is going on. It is not so good if you want to output it to a existing logging structure. 
I would like to either receive a confirmation of success  or the error that was generated, without the terraform formatted string/redundant information. 

A quick and **dirty** solution would be to parse the output and search for the `Error:` string and get the content until the next new line, but everything in me is screaming against it. Maybe there is already an option to get this information without having to resort to this terrible solution. :grin:  

About the `FORCE_COLOR=0`: Im still getting color. :sob:  Less than before, but still some color:

From pod's log:
[...]

PR #2307, opened 2022-11-16
feat(cli): add -no-color tag

Maed223:
Closes: #1916

--no-color tag for `cdktf deploy`,  `cdktf diff`, and `cdktf destroy` disables colored terminal output stemming from both CDKTF CLI and Terraform CLI

--no-color demo


packages/cdktf-cli/src/bin/cmds/deploy.ts
```diff
@@ -76,6 +76,12 @@ class Command extends BaseCommand {
         // Setting value to negative will prevent it from being forwarded to terraform as an argument
         default: -1,
       })
+      .option("no-color", {
```
ansgarm, line 79: We could add this as a note ([chalk docs](https://github.com/chalk/chalk#supportscolor)).
```suggestion
      // The presence of this flag is automatically picked up by `chalk` as well which
      // causes colors to be disabled too for output that the cdktf-cli itself colors.
      .option("no-color", {
```

packages/cdktf-cli/src/bin/cmds/deploy.ts
```diff
@@ -76,6 +76,12 @@ class Command extends BaseCommand {
...
+        type: "boolean",
+        required: false,
+        desc: "Disables terminal formatting sequences in the output.",
+        default: false,
```
ansgarm, line 83: In regards to our [docs](https://developer.hashicorp.com/terraform/cdktf/create-and-deploy/environment-variables) on the environment flag `FORCE_COLOR` we could change the default to
```suggestion
        default: process.env.FORCE_COLOR === "0",
```
which would disable colored output for Terraform, too, if the users sets the flag that we already document in our docs.

Maed223: Don't think the environment flag `FORCE_COLOR` can alter the colored output coming from Terraform. Looks like the `no-color` option needs to be passed to Terraform CLI as well.

Standard View with colored output:
Screen Shot 2022-11-30 at 9 21 36 AM

View with `FORCE_COLOR=0`:
Screen Shot 2022-11-30 at 9 22 08 AM

View with  `FORCE_COLOR=0` and `TF_CLI_ARGS === "-no-color"`:
Screen Shot 2022-11-30 at 9 27 28 AM

Main issue I see with this approach is overriding the `TF_CLI_ARGS` environment variable with a tag

ansgarm: Yeah, `FORCE_COLOR` won't alter the output, but if the user sets it and that affects the default value of our `--no-color` option, the we'd act as if the user used that flag, although they used the env var 😅
I hope I'm not too bad at explaining what I try to say 😅

Maed223: Ahhhh I see what your saying now 😁

ansgarm:
Oh, and only because I just happened to spot this: The `destroy` command might benefit from this `-no-color` option, too 😇

packages/@cdktf/cli-core/src/lib/cdktf-stack.ts
```diff
@@ -385,14 +401,15 @@ export class CdktfStack {
     });
   }
 
-  public async destroy(terraformParallelism?: number) {
+  public async destroy(terraformParallelism?: number, noColor?: boolean) {
     await this.run(async () => {
       this.updateState({ type: "planning", stackName: this.stack.name });
-      const terraform = await this.initalizeTerraform({ isSpeculative: false });
```
DanielMSchmidt, line 391: Not: Why did you decide against putting the noColor option into the terraform client. That would mean fewer places to remember to pass it down 🤔

DanielMSchmidt (commented):
Great job, I would approve it if it has had an integration test. Could be part of the hello world integration test, ideally not checking through snapshots but ensuring no line has a terminal color coding character (there is probably a npm lib to check for those that one could use in the tests)

packages/@cdktf/cli-core/src/lib/cdktf-stack.ts
Maed223: Now that you mention it, that would've been a great idea 😅. Initial thought was to follow the previous pattern, though that wasn't exactly needed since the no-color option isn't being used in various places like many of the other options.

DanielMSchmidt (approved):

PR #2312, opened 2022-11-18
chore: make feature flag tests more resilient to provider updates

DanielMSchmidt:


test/typescript/feature-flags/test.ts
```diff
@@ -56,10 +56,14 @@ describe("full integration test", () => {
...
   const loadStackJson = (workingDir, stackName) => {
-    return fs.readFileSync(
+    const stack = fs.readFileSync(
       path.join(workingDir, "cdktf.out", "stacks", stackName, "cdk.tf.json"),
       "utf-8"
     );
+
+    const json = JSON.parse(stack);
+    delete json.terraform.required_providers;
+    return JSON.stringify(json, null, 2);
```
ansgarm, line 66: nit: Using `json-stable-stringify` here as well (like the stack does) could preserve the way we print empty objects currently – but I see this as completely optional because I expect the current JSON.parse + JSON.stringify dance to be stable (as it parses a stable json string in the first place 😄) – I only suggest this in case you'd prefer to keep the rest of the snapshot as is, so feel free to resolve 😄

ansgarm (approved):

test/typescript/feature-flags/test.ts
DanielMSchmidt: Yeah, I didn't know if we had that at our disposal inside the tests and deemed it not worth it to install it in the tests as well :D

mutahhir (approved):

PR #2313, opened 2022-11-18
chore: run CI more selectively

DanielMSchmidt:
For this I had to pull out linting from the unit workflow. I would love to reduce the examples down to only the changed example, but this does not seem to be trivial so I decoupled it from this change.


mutahhir (approved):
❤️

PR #2314, opened 2022-11-18
chore(docs): prepare for code samples in multiple languages

xiehan:
I happened to look at the docs today and noticed that even in the places where we already have CodeTabs set up with code samples in multiple languages, our docs consistently have the text:

> The following TypeScript example ...

Since we are actively working on translating all the code samples to multiple languages and are hoping to start merging those changes very soon, I think it's okay to do away with the "TypeScript" qualifier now.

This PR is just a result of running the following on the command line:

```
grep -RlI 'The following TypeScript example' ./website | xargs sed -i "" 's/The following TypeScript example/The following example/g'
```

I did not do a manual check for other phrasings.

(re: #1929)


DanielMSchmidt (approved):

laurapacilio (approved):
Thank you so much! I think I had flagged this in another PR (idk somewhere, at some point when we started adding new examples) and so I'm glad to see the TypeScript specific language coming out :) Kudos to the team for adding all these nice, new examples! 🥳

PR #2318, opened 2022-11-18
chore: limit concurrency on releases

DanielMSchmidt:


mutahhir (approved):

.github/workflows/release.yml
```diff
@@ -15,6 +15,10 @@ env:
   SENTRY_ORG: hashicorp
   SENTRY_PROJECT: cdktf-cli
 
+concurrency:
+  group: release
+  cancel-in-progress: true
```
ansgarm, line 20: Can't this be dangerous 😅
If something is merged while a release is running? Is there something we can do to safeguard that case other than "don't merge while we release a new version" 😄

PR #2332, opened 2022-11-22
chore: add expression paranthesis

DanielMSchmidt:


.github/labeler.yml
```diff
@@ -0,0 +1,55 @@
+ci/run-unit/@cdktf/api-core:
```
ansgarm, line 1: ?
```suggestion
ci/run-unit/@cdktf/cli-core:
```

.github/workflows/unit-all.yml
```diff
@@ -32,3 +37,94 @@ jobs:
...
+    uses: ./.github/workflows/unit.yml
+    strategy:
+      fail-fast: false
+      matrix:
+        terraform_version: ["1.2.8"]
+    with:
+      concurrency_group_prefix: pr-cdktf-api-core
+      package: "@cdktf/api-core"
+      terraform_version: ${{ matrix.terraform_version }}
+    secrets: inherit
```
ansgarm, line 52:
```suggestion
  cdktf_cli_core:
    if: ${{ contains(github.event.pull_request.labels.*.name, 'ci/run-unit/@cdktf/cli-core') }}
    uses: ./.github/workflows/unit.yml
    strategy:
      fail-fast: false
      matrix:
        terraform_version: ["1.2.8"]
    with:
      concurrency_group_prefix: pr-cdktf-cli-core
      package: "@cdktf/cli-core"
      terraform_version: ${{ matrix.terraform_version }}
    secrets: inherit
```

ansgarm (changes requested):
I think your mind was already further with that package name than the current reality 😄

.github/workflows/unit-all.yml
DanielMSchmidt: haha good point :)

.github/labeler.yml
```diff
@@ -0,0 +1,55 @@
+ci/run-unit/@cdktf/cli-core:
+  # The workflow used
+  - ".github/workflows/unit.yml"
+  # All packages under test
+  - "packages/@cdktf/cli-core/**"
+  # All scripts we use
+  - "tools/align-version.sh"
```
ansgarm, line 7: I'm wondering whether we could simplify this config by adding `ci/run-unit` with those two repeated files (that are common for all individual unit labels currently). It would then effectively run all unit tests if we change those files while not spamming the PR with all individual unit test labels 🤔

ansgarm (approved):
I really like this improvement! Only had one small idea that could simplify it even more.

.github/labeler.yml
DanielMSchmidt: Sounds good, will do it in a follow-up though

DanielMSchmidt:
Merging this with a red check on the csharp documentation example, will need to fix this in the future, but this at least gets the general CI unstuck

PR #2366, opened 2022-12-01
feat(examples): python upcloud server

ansgarm:
- feat(examples): add upcloud-server python example
- chore(examples): run prettier

supersedes #2239 as I can't push to that branch to rebase & fix lint 😇

### DIFF
--- a/examples/python/upcloud-server/Pipfile
+++ b/examples/python/upcloud-server/Pipfile
@@ -7,4 +7,3 @@ verify_ssl = true
 python_version = "3"
 
 [packages]
-cdktf = {path = "./../../../dist/python/cdktf-0.0.0-py3-none-any.whl"}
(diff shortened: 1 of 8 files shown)


examples/python/upcloud-server/Pipfile
```diff
@@ -0,0 +1,10 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[requires]
+python_version = "3"
+
+[packages]
+cdktf = {path = "./../../../dist/python/cdktf-0.0.0-py3-none-any.whl"}
```
DanielMSchmidt, line 10: I think this changed in the mean time and we need to remove it

DanielMSchmidt (changes requested):

examples/python/upcloud-server/Pipfile
DanielMSchmidt: I mean apparently it works, but removed it from the others as well

ansgarm: Ah, right!

ansgarm: Fixed it anyway 👍

PR #2367, opened 2022-12-01
fix: patch-package the JSII bug

DanielMSchmidt:
Taking the JSII fix into the project early https://github.com/aws/jsii/pull/3866

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -40,7 +40,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-yarn-
       - name: install dependencies
-        run: yarn install --frozen-lockfile --prefer-offline
+        run: yarn install --frozen-lockfile --prefer-offline && yarn run postinstall
       - name: set version
         if: ${{ !inputs.skip_setup }}
         run: tools/align-version.sh "-dev.$GITHUB_RUN_ID"
--- a/patches/jsii-pacmak+1.71.0.patch
+++ b/patches/jsii-pacmak+1.71.0.patch
@@ -1,0 +1,13 @@
+diff --git a/node_modules/jsii-pacmak/lib/targets/python/type-name.js b/node_modules/jsii-pacmak/lib/targets/python/type-name.js
+index d2a2207..fe33ee0 100644
+--- a/node_modules/jsii-pacmak/lib/targets/python/type-name.js
++++ b/node_modules/jsii-pacmak/lib/targets/python/type-name.js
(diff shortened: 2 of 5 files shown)


yarn.lock
```diff
@@ -811,7 +811,7 @@
     chalk "^4.1.2"
     semver "^7.3.8"
 
-"@jsii/spec@1.71.0", "@jsii/spec@^1.71.0":
+"@jsii/spec@1.71.0", "@jsii/spec@^1.68.0":
```
ansgarm, line 814: Why did this spec version change?

patches/jsii-pacmak+1.71.0.patch
```diff
@@ -0,0 +1,4180 @@
+diff --git a/node_modules/jsii-pacmak/bin/jsii-pacmak.js b/node_modules/jsii-pacmak/bin/jsii-pacmak.js
+index 96120e9..cb54420 100644
```
DanielMSchmidt, line 2: weird, the diff shouldn't be this big

yarn.lock
```diff
@@ -6744,7 +6765,7 @@ jsii-docgen@^7.0.157:
     semver "^7.3.8"
     yargs "^16.2.0"
 
-jsii-pacmak@^1.71.0:
+jsii-pacmak@^1.68.0:
```
ansgarm, line 6768: Is this intended?

patches/jsii-pacmak+1.71.0.patch
ansgarm: You might've had 1.68.0 installed at the time of running the patch-package script? 🤔

DanielMSchmidt: Yeah, patch-package took the package-lock.json file instead of the yarn.lock

.github/workflows/integration.yml
```diff
@@ -40,7 +40,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-yarn-
       - name: install dependencies
-        run: yarn install --frozen-lockfile --prefer-offline
+        run: yarn install --frozen-lockfile --prefer-offline && yarn run postinstall
```
ansgarm, line 43: Just out of curiosity: Why is this required here but not locally?

ansgarm (approved):

.github/workflows/integration.yml
DanielMSchmidt: No idea, it's still not applying the patch, I'll just disable the test now to be able to debug the next steps further

PR #2369, opened 2022-12-01
chore: disable python edge test

DanielMSchmidt:
The issue causing this problem will be released in the next days, but we have
other problems in the release pipeline that should be fixed beforehand.
Therefore I'll disable these tests for now and re-add them when updating JSII.

The fix we are waiting for to be released is https://github.com/aws/jsii/pull/3866

### DIFF
--- a/test/python/edge/test.ts
+++ b/test/python/edge/test.ts
@@ -4,7 +4,7 @@ import { QueryableStack, TestDriver } from "../../test-helper";
 import * as path from "path";
 import * as fs from "fs-extra";
 
-describe("full integration test", () => {
+describe.skip("full integration test", () => {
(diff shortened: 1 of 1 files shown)


test/python/edge/test.ts
```diff
@@ -4,7 +4,7 @@ import { QueryableStack, TestDriver } from "../../test-helper";
 import * as path from "path";
 import * as fs from "fs-extra";
 
-describe("full integration test", () => {
+describe.skip("full integration test", () => {
```
ansgarm, line 7: I'd like to see a follow-up task to re-enable this after upgrading JSII to not forget to re-enable this 😇

ansgarm (approved):

PR #2379, opened 2022-12-02
chore: fix finalizing the Sentry release

ansgarm:
By piping to `/dev/null` we can get rid of this Sentry output:
`| (unreleased) | cdktf-cli-0.14.1        | 7          | 7 hours ago    |`
which was part of `status` previously (and required the newline replacement).

By grepping for `unreleased` and inverting the expression, we catch the
expected unreleased release which is the right condition to be able to
finalize the release (the previous one was wrong, as at this time, there
should always have been a release as it is created by an earlier workflow step).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -280,12 +280,12 @@ jobs:
       - name: release status
         id: get_release_status
         run: |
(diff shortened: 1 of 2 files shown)


DanielMSchmidt (approved):
LGTM 👍

PR #2390, opened 2022-12-05
chore: github action expressions can only access items within their declaration

DanielMSchmidt:


.github/workflows/release.yml
```diff
@@ -104,7 +104,7 @@ jobs:
 
   unit_test:
     uses: ./.github/workflows/unit.yml
-    if: jobs.prepare-release.outputs.release_status == 'unreleased'
```
mutahhir, line 107: Needs a

```
needs: 
  - prepare-release  
```

too?

mutahhir: Same for all the others below if true.

mutahhir (approved):

PR #2393, opened 2022-12-05
fix(ci): Improve checking for sentry release states

mutahhir:
Ref: Failing release workflow https://github.com/hashicorp/terraform-cdk/actions/runs/3619258049


mutahhir:
Hmm, this is not really useful. I misunderstood the flow between the two different sentry checks. Closing for now.

PR #2394, opened 2022-12-05
fix: Don't do additional operations on release output

mutahhir:


.github/workflows/release.yml
```diff
@@ -46,7 +46,7 @@ jobs:
         run: |
           status=$(sentry-cli releases list | grep 'cdktf-cli-${{ steps.get_version.outputs.version }} ' && echo 'released' || echo 'unreleased')
           echo "Sentry returned: ${status}"
-          echo "release=$(echo $status | tr '\n' ' ')" >> $GITHUB_OUTPUT
```
ansgarm, line 49: Well they did have a reason 🙈

PR #1955, opened 2022-07-25
chore: replace TF Cloud implementation with TF CLI

DanielMSchmidt:
This will allow us to integrate better with TFC and TFE without additional work being spend on polishing this part.

## TODO

- [x] Check if this solves https://github.com/hashicorp/terraform-cdk/issues/490
- [ ] Check if this solves https://github.com/hashicorp/terraform-cdk/issues/489

PR #2398, opened 2022-12-05
fix: add completed tfc pending states

michaellzc:
related issues & pr:

- https://github.com/hashicorp/terraform-cdk/issues/2232
- https://github.com/hashicorp/terraform-cdk/pull/1955
- https://github.com/hashicorp/terraform-cdk/pull/2243
- https://github.com/hashicorp/terraform-cdk/pull/2360

While https://github.com/hashicorp/terraform-cdk/pull/1955 is currently being worked on, this is the workaround for various causes of the issue described in https://github.com/hashicorp/terraform-cdk/issues/2232

## Context

https://github.com/hashicorp/terraform-cdk/pull/2243 & https://github.com/hashicorp/terraform-cdk/pull/2360 are missing a few more pending states from TFC API

## Testing

This is tested internally across our TFC workspaces with and without auto-apply enabled. I can confirm no more `409` error from TFC and all stacks are deployed in one go.

### DIFF
--- a/packages/@cdktf/cli-core/src/lib/models/terraform-cloud.ts
+++ b/packages/@cdktf/cli-core/src/lib/models/terraform-cloud.ts
@@ -356,12 +356,25 @@ export class TerraformCloud implements Terraform {
     const url = `https://${this.hostname}/app/${this.organizationName}/workspaces/${this.workspaceName}/runs/${result.id}`;
     sendLog(`Created speculative Terraform Cloud run: ${url}`);
 
+    // the source of truth of all pending states are from
+    // https://developer.hashicorp.com/terraform/cloud-docs/api-docs/run#run-states
+    // any state before `apply_queued` is considered pending unless it is specified as a final state
     const pendingStates = [
       "pending",
+      "fetching",
+      "fetching_completed",
+      "pre_plan_running",
+      "pre_plan_completed",
+      "queuing",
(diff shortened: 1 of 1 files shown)


packages/@cdktf/cli-core/src/lib/models/terraform-cloud.ts
```diff
@@ -356,12 +356,25 @@ export class TerraformCloud implements Terraform {
...
       "plan_queued",
       "planning",
       "cost_estimating",
+      "cost_estimated",
       "policy_checking",
+      "policy_override",
+      "policy_checked",
+      "post_plan_running",
+      "post_plan_completed",
     ];
```
ansgarm, line 378: Shouldn't `apply_queued` and `applying` be pending states, too? 🤔
I know we didn't have them previously, but from their explanation it sounds like the run is not done yet, if it has one of these two stages.

ansgarm: @DanielMSchmidt told me that this is about a speculative run, so my above comment makes no sense :)

DanielMSchmidt: This is in the context of a speculative run, so apply is not in the phases :)

ansgarm (approved):

PR #2399, opened 2022-12-06
fix: update the providers test in hcl2cdk

mutahhir:
Experiments in snapshot issues


mutahhir:
Just running some experiments, please ignore for now

PR #2400, opened 2022-12-06
chore: update hcl2cdk tests to be 🟢 

mutahhir:
This PR re-enables the hcl2cdk tests, updates the snapshots, and removes some of the improvements to minimizing the schema. The run time is still ~5 minutes, so that's still somewhat acceptable for now. We'll revisit it for more improvements.


DanielMSchmidt (approved):
Good enough for now :)

PR #2402, opened 2022-12-06
chore: use sentry releases info instead of list and grep

DanielMSchmidt:


.github/workflows/release_next.yml
```diff
@@ -43,7 +43,7 @@ jobs:
       - name: release status
         id: get_release_status
         run: |
-          status=$(sentry-cli releases list | grep 'cdktf-cli-${{ steps.get_version.outputs.version }} ' && echo 'released' || echo 'unreleased')
+          status=$(sentry-cli releases info 'cdktf-cli-${{ steps.get_version.outputs.version }}' && echo 'released' || echo 'unreleased')
```
ansgarm, line 46: Did it work? What is this fixing exactly?

When I last looked at the `info` command, it didn't contain the `unreleased` state of releases already 😅 (But we don't really need it here, but rather in the other check later in the release workflow)
image
image


Also, this will still result in a multiline string that will be echoed into `$GITHUB_OUTPUT`, so I don't think it actually fixes the error we're seeing for releases that already exist 😬

DanielMSchmidt: locally it did work, but as you said it does not tell us anything about the released status. I'm trying to fix the mulit-line issue

ansgarm:
Superseded by https://github.com/hashicorp/terraform-cdk/pull/2403

PR #2406, opened 2022-12-06
chore: clean up unreleased sentry releases

DanielMSchmidt:


ansgarm (approved):
Nice 👏

If this works, we can remove/reword this section: https://github.com/hashicorp/terraform-cdk/blob/main/CONTRIBUTING.md?plain=1#L294

mutahhir (approved):

mutahhir:
Just remembered: Will this revert the release if Homebrew fails? or was that fixed already?

ansgarm:
Ah, good catch. I'd say it shouldn't revert it in that case 🙈

DanielMSchmidt:
It was not fixed yet, I'm still working on getting the overall release out

ansgarm:
> Maybe we should just move to a different mechanic for determening if sth has been released, e.g. the npm release

One benefit to using Sentry is that it can block further release workflows from starting. Using NPM will only "activate" the block when the release is done.

So we'd need to ensure that we (or Mergify 😅) don't merge anything while a release is running.

Issue #1893, opened 2022-06-23
TFE well-known endpoints not being used


hoo29:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### cdktf & Language Versions

0.11.2

### Affected Resource(s)

cdktf cli
npm package "@skorfmann/terraform-cloud"

### Description

The API path in the underlying terraform cloud library is hardcoded to `/api/v2` 
https://github.com/skorfmann/terraform-cloud/blob/ed3f22da9ccd7ae6fdc983633173deef3a07df2c/src/api/terraformCloudApiClient.ts#L6

This means TFE installations with custom paths specified at `.well-known/terraform.json` do not work as all calls return a 404.

### Expected Behavior

The `well-known` configuration is used instead of hardcoded values.

### References

- #1854

hoo29:
An example of the well known endpoint is the main terraform cloud one at https://app.terraform.io/.well-known/terraform.json

DanielMSchmidt:
Yeah that's not ideal, we are thinking about replacing our custom TFC API calls with Terraform so we can implement all the features once instead of twice, the issue is #1913

hoo29:
@DanielMSchmidt - #1913 has been marked as resolved but I am still seeing the same error. Is this expected?

DanielMSchmidt:
Yeah unfortunately it's expected. It's a bug we have been planning to fix by moving to Terraform directly for all TFC/TFE related interaction, but we haven't finalised the fix yet

hoo29:
Ah okay, so there is more to it then just the plan and apply which was fixed in #1913. Is there an open issue for moving all operations to the terraform cli I can follow?

PR #2427, opened 2022-12-15
chore: Support custom endpoints

mutahhir:
Fixes: https://github.com/hashicorp/terraform-cdk/issues/1893

### DIFF
--- a/packages/cdktf-cli/src/bin/cmds/helper/terraform-cloud-client.ts
+++ b/packages/cdktf-cli/src/bin/cmds/helper/terraform-cloud-client.ts
@@ -39,6 +39,44 @@ export interface Workspaces {
   data: WorkspaceData[];
 }
 
+export interface ServiceDiscovery {
+  "modules.v1": string;
+  "providers.v1": string;
+  "motd.v1": string;
+  "state.v2": string;
+  "tfe.v2": string;
+  "tfe.v2.1": string;
+  "tfe.v2.2": string;
+  "versions.v1": string;
+}
+
+let cachedServiceDiscovery: ServiceDiscovery | undefined;
+async function discoverService(hostname: string): Promise<ServiceDiscovery> {
+  if (cachedServiceDiscovery) return Promise.resolve(cachedServiceDiscovery);
+
+  return new Promise<ServiceDiscovery>((resolve, reject) => {
+    const req = https.request(
+      new URL(`https://${hostname}/.well-known/terraform.json`).toString(),
+      (res) => {
+        if (res.statusCode !== 200) {
+          const error = new Error(res.statusMessage);
+          (error as any).statusCode = res.statusCode;
+          return reject(error);
+        }
+        const data = new Array<Buffer>();
+        res.on("data", (chunk) => data.push(chunk));
+        res.once("error", (err) => reject(err));
+        res.once("end", () => {
+          const response = JSON.parse(Buffer.concat(data).toString("utf-8"));
+          cachedServiceDiscovery = response as ServiceDiscovery;
+          return resolve(cachedServiceDiscovery);
+        });
+      }
+    );
+    req.end();
+  });
+}
+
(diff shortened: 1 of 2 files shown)


packages/cdktf-cli/src/bin/cmds/helper/terraform-cloud-client.ts
```diff
@@ -39,6 +39,49 @@ export interface Workspaces {
...
+  "state.v2": string;
+  "tfe.v2": string;
+  "tfe.v2.1": string;
+  "tfe.v2.2": string;
+  "versions.v1": string;
+}
+
+let cachedServiceDiscovery: ServiceDiscovery | undefined;
+async function discoverService(hostname: string): Promise<ServiceDiscovery> {
+  if (cachedServiceDiscovery) return Promise.resolve(cachedServiceDiscovery);
```
ansgarm, line 55: nit: You don't need `Promise.resolve` in an async function
```suggestion
  if (cachedServiceDiscovery) return cachedServiceDiscovery;
```

packages/cdktf-cli/src/bin/cmds/helper/terraform-cloud-client.ts
```diff
@@ -39,6 +39,49 @@ export interface Workspaces {
...
+  "providers.v1": string;
+  "motd.v1": string;
+  "state.v2": string;
+  "tfe.v2": string;
+  "tfe.v2.1": string;
+  "tfe.v2.2": string;
+  "versions.v1": string;
+}
+
+let cachedServiceDiscovery: ServiceDiscovery | undefined;
```
ansgarm, line 53: I suspect it would be safer for the cache to take the hostname into account. I'm thinking about scenarios where users would have TFE and TFC configured and we make requests to both of them in the same process (e.g. when listing orgs that they have access to?)

mutahhir: It's possible, but we'd have to refactor the init and terraform login to support scenarios like that. We only check for the supplied tfe hostname, or use app.terraform.io if not. When we start supporting multiple logins and checking the cached credentials to supply org names, we can revisit this.

ansgarm (approved):

packages/cdktf-cli/src/bin/cmds/helper/terraform-cloud-client.ts
```diff
@@ -117,9 +160,14 @@ async function post(url: string, token: string, data: string) {
   });
 }
 
+async function endpointUrl(tfeHostname: string, path: string) {
+  const serviceDiscovery = await discoverService(tfeHostname);
+  return `https://${tfeHostname}${serviceDiscovery["tfe.v2"]}${path}`;
```
JeroennC, line 165: The serviceDiscovery map values will include `https` and the hostname as part of the value.
https://developer.hashicorp.com/terraform/internals/remote-service-discovery#discovery-process
```suggestion
  return `${serviceDiscovery["tfe.v2"]}${path}`;
```

JeroennC, line 165: Also, i see inconsistencies between my experience with artifactory, and terraform docs, whether the service URLs end with a `/` or not. It would be safer to use a library to combine paths

JeroennC: On further inspection of that page it is actually either absolute or relative, and should follow at least 1 redirect:

> This URL may be either absolute or relative, and if relative it is resolved against the final discovery URL (after following redirects).

mutahhir: In my testing, they were all relative paths, but you're right, the code should be more defensive.

mutahhir:
@JeroennC Thank you for the review! ❤️

Issue #2425, opened 2022-12-15
Terraform CLI waits for variable input forever if variables are missing


ansgarm:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

<img width="959" alt="image" src="https://user-images.githubusercontent.com/1112056/207833869-3e109dae-28f2-4297-b6d7-630316a4f87b.png">

As of #1955 we now invoke Terraform interactively for applying, which causes it to ask for input (e.g. for variables) and to wait basically forever. So the UX to users that don't know Terraform and this output will be confusing.
We have two main options here:
1. Detect when TF asks for input (which we don't support) and crash with an error message that users need to supply variables via TF_VAR_ env variables.
2. Support passing various inputs to the TF CLI (basically extending the "wait for approval" logic for different kinds of questions).

Option 1 is simpler and possibly a good first stepping stone (as detecting those questions is also required for option 2). I'd propose to start with option 1 and then create a follow-up issue to implement option 2 at some point.

DanielMSchmidt:
Same problem with abort, if terraform is killed it sends

```
Do you want to cancel the remote operation?
  Only 'yes' will be accepted to cancel.

  Enter a value: yes
```

PR #2431, opened 2022-12-15
chore: support aborting node-pty commands

### LINKED ISSUE #2429: Aborting cdktf-cli does not stop the plan / apply currently running
### Community Note

### Expected Behavior

The plan / aplpy run should be cancelled when cdktf-cli is aborted

### Actual Behavior

The plan runs to the end, but the apply does not start. If the apply has already started the apply runs until the end

### Steps to Reproduce

1. Create a cdktf program using TFC
2. Add a resource that takes some time, e.g.

```ts
import { NullProvider } from "./.gen/providers/null/provider";
import { Resource } from "./.gen/providers/null/resource";

// ...

new NullProvider(this, "null", {});

new Resource(this, "test4", {
  provisioners: [
    {
      type: "local-exec",
      command: "sleep 100 && echo 'hello world'",
    },
  ],
});
```

3. Run `cdktf deploy --auto-approve` and abort after the plan / apply has started

### References

- #0000

DanielMSchmidt:
Please note:

- I had to cast some parts as any since the global node abort controller definition collides with the polyfilled one we use and I'd like to be consistent in using the polyfilled definitions, since we use the polyfilled implementation
- I formatted the log output for the commands called, mainly to remove the environment. These logs might get sent to sentry so logging the environment is dangerous
- This implementation does not work, after killing the terraform process it asks if the run should be cleaned up, so this working is blocked by https://github.com/hashicorp/terraform-cdk/issues/2425#issuecomment-1353263687 

Closes #2429

### DIFF
--- a/packages/@cdktf/cli-core/src/lib/cdktf-stack.ts
+++ b/packages/@cdktf/cli-core/src/lib/cdktf-stack.ts
@@ -10,6 +10,7 @@ import * as fs from "fs";
 import * as path from "path";
 import { ProviderConstraint } from "./dependencies/dependency-manager";
 import { terraformJsonSchema, TerraformStack } from "./terraform-json";
+import { AbortSignal } from "node-abort-controller";
 
 export type StackUpdate =
   | {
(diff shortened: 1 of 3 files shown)


ansgarm (changes requested):
With this change we have two listeners on the `abortSignal`, doing different things that both race to stop the `pty` 😅

ansgarm:
> This implementation does not work, after killing the terraform process it asks if the run should be cleaned up, so this working is blocked by https://github.com/hashicorp/terraform-cdk/issues/2425#issuecomment-1353263687

Yeah.. It seems like we should add a proper nested state within the `running` state in the state machine that is triggered by the `STOP` event and handles the "Do you want to clean up?"-question by passing "yes" automatically if the machine was stopped from the outside and afterwards leaving the `running` state and transitioning to the `stopped` state.

DanielMSchmidt:
Closing this, it seems to work already

PR #2439, opened 2022-12-21
feat(docs): Convert some code snippets in docs to Golang

ansgarm:
Note: This PR depends on #2440 (so it can start from a clean state)

- chore: add go.work file to allow VSCode Go tools to work in unrelated subdirectories
- feat: run prettier after docs snippets have been inserted into docs
- chore: add go.work.sum
- feat(docs): document first snippet in data-sources.mdx, split assets and constructs into their own files


examples/go/documentation/cdktf.json
```diff
@@ -1,6 +1,6 @@
 {
     "language": "go",
-    "app": "go run main.go",
+    "app": "go run .",
```
DanielMSchmidt, line 3: That's nice in go

packages/@cdktf/hcl2json/build-go.sh
```diff
@@ -1,6 +1,12 @@
 #!/bin/bash
 set -euo pipefail
 
+# We use go.work for examples (so we can have editor support in those)
+# We have to disable it here however, because else this script would try to
+# build all go workspaces (which includes the examples which rely on the cdktf
+# go package which might not exist yet)
+export GOWORK=off
```
DanielMSchmidt, line 8: This is a bit weird, but I'm ok with it for now

DanielMSchmidt (approved):

packages/@cdktf/hcl2json/build-go.sh
ansgarm: It's because we generate Go code (`cdktf`, through JSII) that we use elsewhere in our project 😅
Using a `go.work` file however, gives us autocomplete in all the Go examples, which is reallyyyy helpful when writing them.

PR #2445, opened 2022-12-22
feat(docs): Translate Aspect example to Go

ansgarm:
Note: This PR depends on #2439 (It is a separate PR because the example ~currently still fails due to some JSII error~).

Running this example fails with:
```
panic: @jsii/kernel.SerializationError: Passed to parameter aspect of method cdktf.Aspects#add: Unable to deserialize value as cdktf.IAspect
├── 🛑 Failing value is an object
│      { Tags: [Object] }
╰── 🔍 Failure reason(s):
    ╰─ Value does not have the "$jsii.byref" key
```

which seems to be because we create a struct for our Aspect in Go (to store the tags to append) that is not available in the Nodejs side of things and hence causes an error.

Unfortunately there is no existing example from the AWS CDK that uses an Aspect in Golang. So I don't know whether this is even possible currently.

Related examples / issues I found:
* A (closed) bug in JSII that doesn't store data in the struct of the Aspect though: https://github.com/aws/jsii/issues/2793
* The AWS CDK Go docs Aspects example also does not store data (and seems to be invalid Go code): https://pkg.go.dev/github.com/aws/aws-cdk-go/awscdk/v2#Aspects


ansgarm:
Update: With the help of this [Go testcase](https://github.com/aws/jsii/blob/a11a1c5065118b46c80ca8f4994663e7f5262966/packages/%40jsii/go-runtime-test/project/nativeobjectwithinterdfaces.go#L33) within JSII, I was able to stumble upon my error of passing the struct and not **a pointer to the struct**. Using that solved it ([commit](https://github.com/hashicorp/terraform-cdk/pull/2445/commits/f14bcb1b2afaef337e55b0f9adefde15bb88e072)).

DanielMSchmidt (approved):

PR #2446, opened 2022-12-22
fix: Adds `environment` to keys to preserve case

Jiafi:
Adds `environment` to keys to preserve case.  When using `local-exec`, the environment would get lower cased.  This is unintuitive with how environment variables generally work.

### DIFF
--- a/packages/cdktf/lib/util.ts
+++ b/packages/cdktf/lib/util.ts
@@ -90,7 +90,7 @@ export function keysToSnakeCase(object: any): any {
   }
   const keys = Object.keys(object);
   return keys.reduce((newObject: any, key: string) => {
-    if (key === "tags") {
+    if (key === "tags" || key === "environment") {
       newObject[key] = object[key];
       return newObject;
     }


packages/cdktf/lib/util.ts
```diff
@@ -90,7 +90,7 @@ export function keysToSnakeCase(object: any): any {
   }
   const keys = Object.keys(object);
   return keys.reduce((newObject: any, key: string) => {
-    if (key === "tags") {
+    if (key === "tags" || key === "environment") {
```
jrouly, line 93: Arguably it might be nice to continue applying snake casing to environment variables, since they're typically UPPER_SNAKE_CASE. Although I agree case should be preserved.

Could definitely see the argument, though, that environment variables should simply be left literally as specified (like tags).

ansgarm (approved):
Thank you so much for taking a stab at fixing this, @Jiafi 👏 Appreciate it!

I'm a bit torn about this, as the current special case for `tags` is already something that is probably not used anymore (because bindings generated for Terraform providers don't use `keysToSnakeCase` anymore since #395, and this was probably added originally for the `tags` attribute in various AWS resources).

Upon researching the use of the `keysToSnakeCase` method, I noticed that it probably also wrongly converts keys in [`assumeRoleTags`](https://github.com/hashicorp/terraform-cdk/blob/1908d763a5a676651579983100f92cb08c74777c/packages/cdktf/lib/backends/s3-backend.ts#L146) in the `S3Backend` and possibly other backends that have attributes with the type `{ [key: string]: string }`.

However, I'm overall still in favor of this particular PR because it is simple and we currently already have a special case for `tags` in here and there's no other `environment` in e.g. backends that could be affected by this 😄

PR #2497, opened 2023-01-16
feat(docs): Document `node-gyp` errors if no precompiled binaries are available and the target system misses a native toolchain

ansgarm:


website/docs/cdktf/cli-reference/cli-configuration.mdx
```diff
@@ -61,3 +61,50 @@ The CDKTF CLI ([cdktf-cli](https://github.com/hashicorp/terraform-cdk/tree/main/
...
 
 The use of Checkpoint is completely optional. Refer to the [telemetry documentation](/cdktf/telemetry) for more information about Checkpoint and you can disable it if desired.
+
+## `node-gyp` errors
+
+As of version 0.15 the CDKTF CLI invokes the Terraform CLI interactively in a pseudo-terminal (PTY) for the deploy and destroy commands. This requires native code that needs to be compiled for the architecture of the system the CDKTF CLI is installed on. To make things easier, we maintain [a fork](https://github.com/cdktf/node-pty-prebuilt-multiarch) that includes pre-compiled native bindings for multiple target architectures and Node.js versions. However, there is still the rare possibility, that no precompiled binary is available. In such a case `node-gyp` will attempt to compile the native libraries upon installation of the `cdktf-cli` package and will require additional software to compile the native code that might not be installed yet.
```
DanielMSchmidt, line 67: These docs are only shown from 0.15 onward so we don't need the pre-cursor here

DanielMSchmidt (approved):
One nit, can live with or without it being fixed :) 

error

website/docs/cdktf/cli-reference/cli-configuration.mdx
ansgarm: My reason in favor of the precursor was that by default users would arrive at the docs of the latest version, but I agree that this is not necessary to include.

PR #2504, opened 2023-01-17
fix(provider-generator): fix resources named 'provider' breaking the code generation

ansgarm:


DanielMSchmidt (changes requested):
This change also needs to be reflected in convert

ansgarm:
Is this the right place to implement that?
https://github.com/hashicorp/terraform-cdk/blob/faea9e503af934c8c56b634e598c6573ab23a050/packages/@cdktf/hcl2cdk/lib/generation.ts#L282-L295

And if yes, it seems to be that there are lots of resource name overrides not handled yet, correct? I suspect there is an open issue that is related to that?
I can add a special case to `convert` as part of this PR, but maybe it's better to add a note to an issue tracking this and solve it once and for all of these case? What do you think?

DanielMSchmidt:
> Is this the right place to implement that?

Yes I think so 👍 

I don't think there is an open issue for this, it was fixed in the past already. there shouldn't be any unhandled renamings. It can be that we did not catch these in the coresponding code reviews :(

DanielMSchmidt:
The problem is that these two places are dependant but people are not super aware and it's hard to catch. I try my best to look out for these in code reviews, but I don't catch them all the time

ansgarm:
Yeah, that may be true 😅
I'm asking because this whole bit doesn't appear to be supported either:
https://github.com/hashicorp/terraform-cdk/blob/faea9e503af934c8c56b634e598c6573ab23a050/packages/@cdktf/provider-generator/lib/get/generator/resource-parser.ts#L23-L45

It probably slipped when I implemented that. I'm fine with either a follow-up PR or fixing this as part of this PR, as long as the fix for the generation makes its way into 0.15.1 (so we can release a pre-built provider for datablocks for 0.15).

ansgarm:
To clarify (for future me): In the generator code linked above `name` needs some special handling for supporting this PR and the other linked code, as name will contain e.g. `provider` whereas the resource is named e.g. `databricks.ProviderResource`.

ansgarm:
Okay, so I did a bit of research and it seems that `convert` is already using the code that I adjusted through this:
https://github.com/hashicorp/terraform-cdk/blob/faea9e503af934c8c56b634e598c6573ab23a050/packages/@cdktf/hcl2cdk/lib/expressions.ts#L231-L238

So, it should already work. I can add a test if you want, @DanielMSchmidt?
That said, it'll only work if we have a provider schema, is that right?
So, this "sensible default" is something we should address – I'd vote for a separate issue and separate PR for this though, as this is not a small addition and would fit the theme for 0.16 either way. What do you think?

ansgarm:
fyi: I created a follow-up issue for improving the current case where `convert` does not try to do overridden name changes: https://github.com/hashicorp/terraform-cdk/issues/2508

DanielMSchmidt (approved):

Issue #2116, opened 2022-09-14
Allow skipping stacks with no changes


jld-adriano:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

We use multi stack deployments quite frequently during development, using "*", but do not want to auto-approve since seeing the changes is part of the development process. 
However, we have to manually skip all the stacks with no changes before getting to the ones that may have changed on each loop.

Something like `--skip-no-changes` would speed up this use case significantly

Another useful flag would be to enable supression of "refresh" and "init" logs

ansgarm:
Hi 👋

I just tested this and this issue is now obsolete since we refactored how we invoke the Terraform CLI.

In a short example with a chain of three stacks with each depending on the previous one, it didn't wait for approval on the first stack and last stack when there were only changes on the second one.

<img width="770" alt="Screenshot 2023-07-27 at 15 08 54" src="https://github.com/hashicorp/terraform-cdk/assets/1112056/b07fb1c0-1e47-4089-b365-e4958196c075">

PR #2533, opened 2023-01-24
chore(lib): Skipping deploy on stacks with no changes

Maed223:
Closes: #2116

Currently a WIP

The basic idea here is to only deploy stacks that have no changes. This could potentially speed up multi-stack deployments using `*` as well as removing clutter from the terminal.

My current approach is to create a plan file, read the plan file using `terraform show`, check if there are any changes with in that plan file, and then selectively deploy based on that condition.  Open to feedback on whether this is the simplest approach 😁.

### DIFF
--- a/packages/@cdktf/cli-core/src/lib/cdktf-project.ts
+++ b/packages/@cdktf/cli-core/src/lib/cdktf-project.ts
@@ -610,7 +610,6 @@ export class CdktfProject {
           method === "deploy"
             ? nextRunningExecutor.deploy(opts)
             : nextRunningExecutor.destroy(opts);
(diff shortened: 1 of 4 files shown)


jld-adriano:
Some feedback here from someone who would use this:

Would be awesome to have this, but a simpler version that would be very useful for our use case would just be something that would auto "Dismiss" when there are no changes, even if we get all the console output

ansgarm:
I'd say we can close this one, since it is [obsolete](https://github.com/hashicorp/terraform-cdk/issues/2116#issuecomment-1653601003) by now 😇

PR #2539, opened 2023-01-27
feat: Add a wrap in stack optional param to cdktf convert

mutahhir:
I wanted to have generated code that could be imported and run without any additional steps, so I created an extra flag for wrapping convert in a stack.

### DIFF
--- a/packages/cdktf-cli/src/bin/cmds/convert.ts
+++ b/packages/cdktf-cli/src/bin/cmds/convert.ts
@@ -45,6 +45,11 @@ class Command extends BaseCommand {
         type: "array",
         default: [],
       })
+      .option("stack", {
+        describe: "Wrap the generated code within a stack class",
+        type: "boolean",
+        default: false,
+      })
       .showHelpOnFail(true);
 
   public async handleCommand(argv: any) {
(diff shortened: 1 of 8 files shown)


ansgarm:
Sounds a bit related to our `--from-terraform` flag for `cdktf init` 😄 But I still like this idea (as the other requires a flag).

To be able to run this, however, there still needs to be something that adds the provider block. Or is that part of the convert result already?

mutahhir:
> Sounds a bit related to our `--from-terraform` flag for `cdktf init` 😄 But I still like this idea (as the other requires a flag).
In fairness, this is also a `--stack` flag, so that's similar, but I guess the `--from-terraform` creates a full project.

> To be able to run this, however, there still needs to be something that adds the provider block. Or is that part of the convert result already?

If the terraform contains a provider block, then the output will have it. Was that what you were referring to? If you're asking about which providers to process, you pass along the `--provider="..."` flags to the convert command and it does the right thing. Convert doesn't seem to (nor does it need to, I suppose) import providers based on the required providers block in the source TF file

ansgarm:
> If the terraform contains a provider block, then the output will have it. Was that what you were referring to?

ansgarm (approved):

PR #2540, opened 2023-01-27
feat: add short link to PRs

ansgarm:


cdk.tf/vercel.json
```diff
@@ -33,6 +33,11 @@
       "destination": "https://github.com/hashicorp/terraform-cdk/issues/:id",
       "permanent": false
     },
+    {
+      "source": "/prs/:id",
```
mutahhir, line 37: Since this is a team-only thing, let's put this under an `internal` or such path.

ansgarm: But that makes the url longer 😄
(And we already have /issues; see directly above)

DanielMSchmidt (approved):

cdk.tf/vercel.json
xiehan: Couldn't you technically reuse that one? All PRs are issues (but not all issues are PRs).
https://cdk.tf/issues/2540

ansgarm: Right.. Yeah, I could do that 💡

PR #2543, opened 2023-01-27
fix: cdktf login issues

mutahhir:
fixes: https://github.com/hashicorp/terraform-cdk/issues/2538

### DIFF
--- a/packages/cdktf-cli/src/bin/cmds/__tests__/handlers-login.test.ts
+++ b/packages/cdktf-cli/src/bin/cmds/__tests__/handlers-login.test.ts
@@ -1,0 +1,82 @@
+import { login } from "../handlers";
+import { TerraformLogin } from "../helper/terraform-login";
+
+jest.mock("../helper/terraform-check", () => {
+  return {
+    __esmodule: true,
+    terraformCheck: jest.fn().mockResolvedValue(true),
+  };
+});
+jest.mock("../helper/version-check", () => {
+  return {
+    __esmodule: true,
+    displayVersionMessage: jest.fn().mockResolvedValue(true),
+  };
+});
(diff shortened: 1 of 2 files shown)


packages/cdktf-cli/src/bin/cmds/__tests__/handlers-login.test.ts
```diff
@@ -0,0 +1,82 @@
+import { login } from "../handlers";
```
mutahhir, line 1: I wanted to add a regression test for this issue, but as you can see it's a lot of work, and is a bit 'icky', to use a professional term. Ideally, if there was more time, I'd refactor this a bit more, but I doubt we can do away with mocking a _lot_ of things. 

What do you think? We should try to aim for having more of these kinds of tests for this level of our CLI, since we don't seem to have many any right now.

packages/cdktf-cli/src/bin/cmds/handlers.ts
```diff
@@ -342,7 +342,7 @@ export async function login(argv: { tfeHostname: string }) {
     }
   }
 
-  await showUserDetails(sanitizedToken);
+  await showUserDetails(sanitizedToken || token);
```
DanielMSchmidt, line 345: Maybe we should just sanitise the token later in the process

packages/cdktf-cli/src/bin/cmds/__tests__/handlers-login.test.ts
DanielMSchmidt: I'm unsure, I sort of like them, but these are also the kind of tests that might break easily when we do internal refactorings. Maybe a CLI e2e test would be an alternative 🤔

DanielMSchmidt (approved):

PR #2556, opened 2023-01-31
feat: add cache flag for prettier to speed up subsequent runs

ansgarm:
Uses `--cache` options to speed up prettier runs through `yarn run format`


DanielMSchmidt (approved):

DanielMSchmidt:
Nice  👌

PR #2558, opened 2023-01-31
chore: translate rest of the docs to C#

DanielMSchmidt:
The Operators example does not work in C# I added a bug into our issues and linked it, so we are transparent to users. Right below is a workaround using plain strings so I don't feel like this is a big problem in this case

### DIFF
--- a/examples/csharp/documentation/OperatorStack.cs
+++ b/examples/csharp/documentation/OperatorStack.cs
@@ -1,0 +1,36 @@
+
+using System;
+using System.IO;
+using System.Collections.Generic;
+using System.Linq;
+using Constructs;
+using HashiCorp.Cdktf;
+using aws.Provider;
+using aws.DataAwsAvailabilityZones;
+
+namespace Examples
+{
+    class OperatorsStack : TerraformStack
+    {
+        public OperatorsStack(Construct scope, string name) : base(scope, name)
+        {
+            new AwsProvider(this, "aws", new AwsProviderConfig {
+                Region = "eu-central-1"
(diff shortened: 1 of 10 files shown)


examples/csharp/documentation/OperatorStack.cs
```diff
@@ -0,0 +1,36 @@
...
+                Region = "eu-central-1"
+            });
+
+            /*
+            // DOCS_BLOCK_START:operators
+            DataAwsAvailabilityZones zones = new DataAwsAvailabilityZones(this, "zones", new DataAwsAvailabilityZonesConfig {
+                State = "available"
+            });
+
+            // This does not work in CDKTF as of now, see
```
ansgarm, line 27:
```suggestion
            // This does not work in CDKTF as of now, refer to
```

examples/csharp/documentation/VariablesStack.cs
```diff
@@ -0,0 +1,59 @@
...
+        // const commonTags = new TerraformLocal(this, "common_tags", {
+        // Service: "service_name",
+        // Owner: "owner",
+        // });
+
+        // new Instance(this, "example", {
+        // ami: "ami-abcde123",
+        // instanceType: "t2.micro",
+        // tags: commonTags.expression,
+        // });
```
ansgarm, line 45: leftover? If not: What purpose does this serve?

ansgarm (commented):
Looking good 😎
Only two small things

ansgarm (approved):

PR #2562, opened 2023-01-31
fix: use correctly named syntax in comments

DanielMSchmidt:


ansgarm (approved):
Nice!

Not blocking on this, but is there a way to reuse that same initialization that is now copied into multiple places?

DanielMSchmidt:
No idea, I'm not a C# expert. Maybe a method?

PR #2570, opened 2023-02-01
chore: refactor 'needs upgrade'

mutahhir:
Step 1 in trying to improve our `cdktf plan/apply/destroy` running `terraform init`. Will base other PRs on top of this one.

### DIFF
--- a/packages/@cdktf/cli-core/src/lib/terraform-provider-lock.ts
+++ b/packages/@cdktf/cli-core/src/lib/terraform-provider-lock.ts
@@ -1,6 +1,7 @@
 import path from "path";
 import fs from "fs/promises";
 import { ProviderConstraint } from "./dependencies/dependency-manager";
+import { logger } from "@cdktf/commons";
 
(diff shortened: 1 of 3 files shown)


packages/@cdktf/cli-core/src/lib/terraform-provider-lock.ts
```diff
@@ -0,0 +1,103 @@
...
+  public async providers(forceReread = false) {
+    if (!this._providerLockData || forceReread) {
+      const contents = await this.readProviderLockFile();
+      this._providerLockData = this.parseProviderLockFile(contents);
+    }
+
+    return this._providerLockData.providers;
+  }
+
+  public async hasProvider(constraint: ProviderConstraint) {
```
ansgarm, line 92: Would this more specific name be fitting?
```suggestion
  public async hasMatchingProvider(constraint: ProviderConstraint) {
```

ansgarm (approved):

PR #2572, opened 2023-02-01
chore: Don't run `terraform init` when not required

mutahhir:
This PR will prevent us from running `terraform init` when there isn't a change between what's present within provider lock, and the generated required providers. This makes cdktf plan/apply/destroy during development much faster.

depends on #2570

### DIFF
--- a/packages/@cdktf/cli-core/src/test/lib/terraform-provider-lock.test.ts
+++ b/packages/@cdktf/cli-core/src/test/lib/terraform-provider-lock.test.ts
@@ -1,10 +1,11 @@
 import { TerraformProviderLock } from "../../lib/terraform-provider-lock";
-import { readFile } from "fs/promises";
+import { readFile, stat } from "fs/promises";
 import { ProviderConstraint } from "../../lib/dependencies/dependency-manager";
 
 jest.mock("fs/promises", () => {
   return {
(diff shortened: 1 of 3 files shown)


packages/@cdktf/cli-core/src/test/lib/terraform-provider-lock.test.ts
```diff
@@ -1,10 +1,11 @@
 import { TerraformProviderLock } from "../../lib/terraform-provider-lock";
-import { readFile } from "fs/promises";
+import { readFile, stat } from "fs/promises";
 import { ProviderConstraint } from "../../lib/dependencies/dependency-manager";
 
 jest.mock("fs/promises", () => {
   return {
     readFile: jest.fn().mockResolvedValue("contents"),
+    stat: jest.fn().mockResolvedValue({}),
```
ansgarm, line 8: Since watching The Good Doctor (recommended!) "stat" has a completely new meaning to me 😄

ansgarm (approved):

villelahdenvuo:
I see the latest version on npm is 0.16.0-pre.27, how often do you usually publish to npm?

ansgarm:
@villelahdenvuo `0.15.3` (which includes this fix) just got released 👍

PR #2589, opened 2023-02-02
feat(provider-generator): Add docstring to main class generated for modules

ansgarm:
Includes link to the Terraform Registry if the module is sourced from there.

This is something that I could have used a few times while writing the examples, so here it is.


packages/@cdktf/provider-generator/lib/get/__tests__/generator/__snapshots__/module-generator.test.ts.snap
```diff
@@ -8354,6 +8490,11 @@ export interface ModuleConfig extends TerraformModuleUserConfig {
    */
   readonly fetchCallerIdentity?: boolean;
 }
+/**
+ * Defines an Module based on a Terraform module
+ *
+ * Source at ./module
```
DanielMSchmidt, line 8496: Isn't this file in e.g. `./.gen/modules/Module.ts` or the like and the module would be in `./module`, so the path is incorrect, isn't it?

ansgarm: Unfortunately, I don't have much information about where the module code is really coming from. That said, I'm rendering a docstring for the `Module.ts` file here, so the code should actually be located next to the generated file as it is also passed like that to the modules `source` attribute a few lines further down in the module generator.

DanielMSchmidt (approved):

DanielMSchmidt (approved):

PR #2592, opened 2023-02-02
feat(lib): Add validation against accidental backend reuse

ansgarm:
depends on #2591

- fix(lib): fix cause for circular dependency in tests when importing app
- feat(lib): Add validation for accidental backend reuse

### DIFF
--- a/packages/cdktf/lib/validations/validate-backends.ts
+++ b/packages/cdktf/lib/validations/validate-backends.ts
@@ -1,0 +1,70 @@
+// Copyright (c) HashiCorp, Inc
+// SPDX-License-Identifier: MPL-2.0
+import { IConstruct, IValidation } from "constructs";
+import isEqual = require("lodash.isequal");
+import { TerraformBackend } from "..";
+
+/**
+ * A validation that is added by default, ensuring that the same backend config
+ * is not accidentally used for different stacks
+ */
+export class ValidateBackends implements IValidation {
+  /**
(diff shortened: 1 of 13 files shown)


packages/cdktf/lib/validations/validate-backends.ts
```diff
@@ -0,0 +1,70 @@
...
+): boolean {
+  // TODO: we could add `LocalBackend.isLocalBackend()` static helpers (breaking change!)
+  // and a isConflicting(otherBackend) method to our backends to move this logic into
+  // the individual backends instead of having it here
+
+  const ta = a.toTerraform(); // e.g. { terraform: { backend: { local: { path: 'terraform.tfstate' } } } }
+  const tb = b.toTerraform();
+
+  // do a deep compare for now
+  // TODO: in theory backends could be partly configured via environment variables which could make this a false positive
```
ansgarm, line 68: I'm unsure whether we need to add a way to skip this validation (other than skipping all validations) for this case

DanielMSchmidt: We could think about adding this in the user-code and in our templates to make skipping this easier and to educate folks about validations

DanielMSchmidt (dismissed):
Let's make this a configuration option for the app so folks can disable this if they don't want it

mutahhir (approved):

team-tf-cdk:

PR #2605, opened 2023-02-07
feat(cli): support migrating state between backends

### LINKED ISSUE #489: Handle Switching Backend from Remote to Something Else
### Community Note

### Description

When the Remote Backend is dropped in favour of something else (e.g. Local), the Terraform CLI detects this change and asks if the existing state should be copied, see:

Screenshot 2021-01-13 at 14 53 04

The `init` command has a `force-copy` flag, which would essentially confirm this question with a yes. What do you think, is it safe to assume it's ok confirming this all the time and make it a default flag on the `init` step? 

### References

- #449

### LINKED ISSUE #490: Auto Migrate State into Terraform Cloud
### Community Note

### Description

We're aiming to improve the usability for Terraform Cloud scenarios in cdktf. One part of this is the state migration into the Terraform Cloud. There are two scenarios:

1. Fresh / empty remote workspace
2. Remote workspace with pre-existing state

While it seems safe to assume that it's ok to auto-migrate state in the first case, we need to confirm with the user in the second case. 

The Terraform CLI is confirming even in he first case, and erroring out in the 2nd case.

We'll need to think about the full process and potential edge cases, if we really want to do this automatically - even when confirming with the user.

##### No State

Screenshot 2021-01-13 at 15 34 29

#####  Existing State

Screenshot 2021-01-13 at 15 25 41

### References

- #449

DanielMSchmidt:
After this change a --migrate-state flag can be passed to diff, deploy, destroy
that will migrate the state from the previous backend to the current one.

Please not that Terraform CLI does not support migrating from the cloud configuration
to other backends, as a result this is not supported in cdktf either.

Closes #490
Closes #489

### DIFF
--- a/packages/@cdktf/cli-core/src/lib/models/terraform-cli.ts
+++ b/packages/@cdktf/cli-core/src/lib/models/terraform-cli.ts
@@ -122,8 +122,13 @@ export class TerraformCli implements Terraform {
       args.push("-no-color");
     }
 
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    let initCanNotContinue = (_err: any) => {};
+    const rejectsIfInitCanNotContinue = new Promise((_resolve, reject) => {
+      initCanNotContinue = reject;
+    });
+
     const stdout = this.onStdout("init");
-    const stderr = this.onStderr("init");
     const { actions, progress } = spawnPty(
       {
         file: terraformBinaryName,
@@ -139,8 +144,8 @@ export class TerraformCli implements Terraform {
           if (migrateState) {
             actions.writeLine("yes");
           } else {
-            actions.writeLine("no");
-            stderr(
+            actions.stop();
+            initCanNotContinue(
               "Please pass the --migrate-state flag to migrate your state"
             );
           }
@@ -151,7 +156,7 @@ export class TerraformCli implements Terraform {
       actions.stop();
     });
 
-    await progress;
+    await Promise.race([progress, rejectsIfInitCanNotContinue]);
 
     // TODO: this might have performance implications because we don't know if we're
     // running a remote plan or a local one (so we run it always for all platforms)
--- a/packages/@cdktf/cli-core/src/lib/models/terraform-cli.ts
+++ b/packages/@cdktf/cli-core/src/lib/models/terraform-cli.ts
@@ -129,7 +129,7 @@ export class TerraformCli implements Terraform {
     });
 
     const stdout = this.onStdout("init");
-    const { actions, progress } = spawnPty(
+    const { actions, exitCode } = spawnPty(
       {
         file: terraformBinaryName,
         args,
@@ -156,6 +156,11 @@ export class TerraformCli implements Terraform {
       actions.stop();
     });
 
+    const progress = exitCode.then((code) => {
+      if (code !== 0) {
+        throw new Error(`terraform init failed with exit code ${code}`);
+      }
+    });
     await Promise.race([progress, rejectsIfInitCanNotContinue]);
 
     // TODO: this might have performance implications because we don't know if we're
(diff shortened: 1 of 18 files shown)


packages/@cdktf/cli-core/src/lib/models/terraform-cli.ts
```diff
@@ -106,28 +107,51 @@ export class TerraformCli implements Terraform {
         createTerraformLogHandler(phase, filter)(stderr.toString(), true);
   }
 
-  public async init(needsUpgrade: boolean, noColor?: boolean): Promise<void> {
+  public async init(
+    needsUpgrade: boolean,
+    noColor: boolean,
+    migrateState: boolean
```
ansgarm, line 113: nit: We're slowly approaching a count of args here, where I'd prefer a config object over a list of args 😅

packages/@cdktf/cli-core/src/lib/models/terraform-cli.ts
```diff
@@ -106,28 +107,51 @@ export class TerraformCli implements Terraform {
...
+      (data) => {
+        stdout(data);
+        if (data.includes("Should Terraform migrate your existing state?")) {
+          if (migrateState) {
+            actions.writeLine("yes");
+          } else {
+            actions.writeLine("no");
+            stderr(
+              "Please pass the --migrate-state flag to migrate your state"
+            );
```
ansgarm, line 145: Nice!

Will `terraform init` return with a non-zero exit code if we send "no"? If I remember correctly `terraform apply` does so when we reject the plan.

packages/cdktf-cli/src/bin/cmds/deploy.ts
```diff
@@ -84,6 +84,12 @@ class Command extends BaseCommand {
         required: false,
         desc: "Disables terminal formatting sequences in the output.",
       })
+      .option("migrate-state", {
+        type: "boolean",
+        default: false,
+        required: false,
+        desc: "Reconfigure a backend, and attempt to migrate any existing state for all selected stacks.",
```
ansgarm, line 91: Maybe reword this to
```suggestion
        desc: "Pass this flag after switching state backends to approve a state migration for all targeted stacks",
```

packages/cdktf-cli/src/bin/cmds/destroy.ts
```diff
@@ -73,6 +73,12 @@ class Command extends BaseCommand {
         required: false,
         desc: "Disables terminal formatting sequences in the output.",
       })
+      .option("migrate-state", {
+        type: "boolean",
+        default: false,
+        required: false,
+        desc: "Reconfigure a backend, and attempt to migrate any existing state for all selected stacks.",
```
ansgarm, line 80: Maybe reword as proposed above?
```suggestion
        desc: "Pass this flag after switching state backends to approve a state migration for all targeted stacks",
```

packages/cdktf-cli/src/bin/cmds/diff.ts
```diff
@@ -65,6 +65,12 @@ class Command extends BaseCommand {
         required: false,
         desc: "Disables terminal formatting sequences in the output.",
       })
+      .option("migrate-state", {
+        type: "boolean",
+        default: false,
+        required: false,
+        desc: "Reconfigure a backend, and attempt to migrate any existing state for the selected stack.",
```
ansgarm, line 72: Reeeeword, too?
```suggestion
        desc: "Pass this flag after switching state backends to approve a state migration for the targeted stack",
```

ansgarm (approved):
Looking good!
I have a proposed rewording for the flag description, but no strong feelings there

packages/@cdktf/cli-core/src/lib/models/terraform-cli.ts
DanielMSchmidt: Hmm, I'm unsure. I think I'll just cancel the init and throw an error, thats safer

PR #2621, opened 2023-02-09
fix(provider-generator): pin @types/node version (Ansgars attempt)

ansgarm:
- fix(provider-generator): pin @types/node version
- fix: update yarn.lock
- chore: resolve transitive dependencies to this version
- chore: remove node-fetch from hcl2json
- chore: add type depdendency on node-fetch
- chore: don't bundle @types/node
- Revert "chore: remove node-fetch from hcl2json"

#2623 would be preferred, if that one works

⚠️ if we end up merging this, we should be careful with the implications of the `chore: don't bundle @types/node` commit


ansgarm:
Closing this in favor of #2623 which seems to becoming green as well

PR #2628, opened 2023-02-10
chore: add copywrite tool

DanielMSchmidt:
We are now using: https://github.com/hashicorp/copywrite


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -1,3 +1,6 @@
+# Copyright (c) HashiCorp, Inc.
```
xiehan, line 1: FYI: I asked whether issue templates should be included or not, and we're still waiting for a response from Legal but I'm guessing the answer will be these don't need the header: https://hashicorp.slack.com/archives/C03CR2GBBB2/p1675361917510739

ansgarm (changes requested):
It seems that the pre-commit hook still points to the script that you removed?
https://github.com/hashicorp/terraform-cdk/blob/59404b741908488e855ee007d8917435f90f1fa4/package.json#L93

ansgarm (approved):

PR #2632, opened 2023-02-13
fix(hcl2cdk): fix type coercion

### LINKED ISSUE #1891: Convert: type coercion does not work
### Community Note

### Description

A converted variable can be of any type possibly lack type information at all. Each variable is used in the converted code through `myVar.value` (I think value is even dropped if a property access happens). They should be used according to the type they have and either access through `stringValue`, `numberValue` or `Token.asStringMap(myVar.value)` etc.

### Example

Convert this:

```tf
variable "default_tags" {
  type        = map(string)
  description = "Map of default tags to apply to resources"
  default = {
    project = "Learning Live with AWS & HashiCorp"
  }
}

resource "aws_eip" "nat" {
  vpc = true
  tags = {
    "Name" = "${var.default_tags.project}-nat-eip"
  }
}

```

### Things to cover

- [ ] Locals and Variables should be handled via `numberValue`, etc.
- [ ] References should be handled via `Token.asNumber` (also for Module outputs)
- [ ] Expressions: How to deal with types in expressions and functions

### References

- #0000

DanielMSchmidt:
Closes #1891


team-tf-cdk:

packages/@cdktf/hcl2cdk/test/__snapshots__/iteration.test.ts.snap
```diff
@@ -56,17 +57,19 @@ awsRoute53RecordExample.addOverride(
...
 awsAcmCertificateValidationExample.overrideLogicalId(\\"example\\");
 const awsLbListenerExample = new aws.lbListener.LbListener(this, \\"example_5\\", {
-  certificateArn: awsAcmCertificateValidationExample.certificateArn,
+  certificateArn: cdktf.Token.asString(
+    awsAcmCertificateValidationExample.certificateArn
+  ),
   defaultAction: [
     {
-      targetGroupArn: \\"best-target\\",
+      target_group_arn: \\"best-target\\",
```
DanielMSchmidt, line 72: This is a bug I introduced that needs fixing

packages/@cdktf/hcl2cdk/test/__snapshots__/iteration.test.ts.snap
```diff
@@ -34,7 +35,7 @@ const awsRoute53RecordExample = new aws.route53Record.Route53Record(
     records: [\\"\${each.value.record}\\"],
     ttl: 60,
     type: \\"\${each.value.type}\\",
-    zoneId: dataAwsRoute53ZoneExample.zoneId,
+    zoneId: cdktf.Token.asString(dataAwsRoute53ZoneExample.zoneId),
```
ansgarm, line 38: Why is this wrapped? `zoneId` [already returns a string token](https://github.com/cdktf/cdktf-provider-aws/blob/eba8748e62f84d95bbc639703e6aabd6277cf7d3/src/data-aws-route53-zone/index.ts#L226) 👀

packages/@cdktf/hcl2cdk/test/__snapshots__/resources.test.ts.snap
```diff
@@ -67,49 +67,59 @@ new aws.cloudfrontDistribution.CloudfrontDistribution(this, \\"s3_distribution\\
...
   origin: [
     {
-      domainName: \\"aws_s3_bucket.b.bucket_regional_domain_name\\",
-      originId: \\"local_s3_origin_id\\",
-      s3OriginConfig: {
-        originAccessIdentity:
-          \\"origin-access-identity/cloudfront/ABCDEFG1234567\\",
-      },
+      domain_name: \\"aws_s3_bucket.b.bucket_regional_domain_name\\",
+      origin_id: \\"local_s3_origin_id\\",
```
ansgarm, line 116: This has been there before, but it seems like the value is an unreplaced reference

packages/@cdktf/hcl2cdk/lib/types.ts
```diff
@@ -10,5 +10,12 @@ export type Scope = {
...
   constructs: Set<string>;
-  variables: Record<string, { resource: string; variableName: string }>;
+  variables: Record<
+    string,
+    {
+      resource: string;
+      variableName: string;
+    }
+  >;
+  hasTokenBasedTypeCoercion: boolean;
```
ansgarm, line 20: maybe add a comment that this is just for adding the imports and can be removed as soon as we derive that information from the AST?

ansgarm: Else it might be around for longer 😄

packages/@cdktf/hcl2cdk/test/coerceType.test.ts
```diff
@@ -0,0 +1,191 @@
...
+      )
+    ).toEqual(code);
+  });
+
+  it.each([
+    {
+      code: `"300"`,
+      type: "string",
+      targetType: "number",
+      expectedCode: `cdktf.Token.asNumber("300")`,
```
ansgarm, line 44: I like how these are written 💯

packages/@cdktf/hcl2cdk/lib/coerceType.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  const isTerraformVariableOrLocal =
+    ast.type === "MemberExpression" &&
+    ast.property.type === "Identifier" &&
+    ast.property.name === "value" &&
+    ast.object.type === "Identifier" &&
+    Object.values(scope.variables).some(
+      (knownVars) =>
+        knownVars.variableName === (ast.object as t.Identifier).name
+    );
```
ansgarm, line 46: Would this match e.g. `someResource.value`? There are [multiple](https://github.com/search?q=repo%3Acdktf%2Fcdktf-provider-aws%20value&type=code) of those in e.g. the AWS TF provider bindings 😬

packages/@cdktf/hcl2cdk/lib/coerceType.ts
```diff
@@ -0,0 +1,128 @@
...
+        case "number":
+          scope.hasTokenBasedTypeCoercion = true;
+          return template.expression(`cdktf.Token.asNumberList(%%ast%%)`)({
+            ast: ast,
+          });
+        case "bool":
+          scope.hasTokenBasedTypeCoercion = true;
+          return template.expression(`cdktf.Token.asAny(%%ast%%)`)({
+            ast: ast,
+          });
```
ansgarm, line 69: There's no `default` case for lists here – is that intended? (Map has one)

packages/@cdktf/hcl2cdk/lib/generation.ts
```diff
@@ -59,6 +62,75 @@ function getReference(graph: DirectedGraph, id: string) {
   }
 }
 
+function destructureAst(ast: t.Expression): string[] | undefined {
```
ansgarm, line 65: What does this function do? Create a path of an expression? A small comment with an example would really help.

packages/@cdktf/hcl2cdk/lib/generation.ts
```diff
@@ -59,6 +62,75 @@ function getReference(graph: DirectedGraph, id: string) {
...
+    }
+
+    const [astVariableName, ...attributes] = destructuredAst;
+    const variable = Object.values(scope.variables).find(
+      (x) => x.variableName === astVariableName
+    );
+
+    if (!variable) {
+      // We don't know, this should not happen, but if it does we assume the worst case and make it dynamic
+      return "dynamic";
```
ansgarm, line 112: I'd prefer failing if this should not happen – or at least logging an error.

packages/@cdktf/hcl2cdk/lib/generation.ts
```diff
@@ -97,7 +174,7 @@ export const valueToTs = async (
         return t.arrayExpression(
           await Promise.all(
             unwrappedItem.map((i) =>
-              valueToTs(scope, i, path, nodeIds, scopedIds)
+              valueToTs(scope, i, `${path}[]`, nodeIds, scopedIds)
```
ansgarm, line 177: Why do we need this? is this faking an array index access? Or is path a type of some sort?

packages/@cdktf/hcl2cdk/lib/provider.ts
```diff
@@ -255,3 +269,33 @@ export function getProviderRequirements(plan: Plan) {
...
+    return attributeType;
+  }
+
+  // Schema
+  if ("version" in attributeType) {
+    return "dynamic";
+  }
+
+  // Block type
+  console.log(
```
ansgarm, line 297: Should we change this into a debug statement instead?

packages/@cdktf/hcl2cdk/lib/provider.ts
```diff
@@ -255,3 +269,33 @@ export function getProviderRequirements(plan: Plan) {
...
+
+  // Complex attribute type
+  if (Array.isArray(attributeType)) {
+    return attributeType;
+  }
+
+  // Schema
+  if ("version" in attributeType) {
+    return "dynamic";
+  }
```
ansgarm, line 294: When does this happen?

ansgarm (commented):
Nice work – looks good mostly 👍

packages/@cdktf/hcl2cdk/lib/types.ts
DanielMSchmidt: It might be around for longer, if we decide against doing the refactoring :D But yeah I can add a comment

team-tf-cdk:

team-tf-cdk:

team-tf-cdk:

team-tf-cdk:

team-tf-cdk:

team-tf-cdk:

packages/@cdktf/hcl2cdk/lib/coerceType.ts
DanielMSchmidt: Yes good point, scope.variables holds all TS-variable names, therefore also resources that are referenced. I added a new property to the map now indicating if the variable is varOrLocal, so that this can not lead to errors

team-tf-cdk:

packages/@cdktf/hcl2cdk/lib/coerceType.ts
DanielMSchmidt: I can't recall but I think making it explicit is best, the code is already hard enough without trying to be clever :D

team-tf-cdk:

team-tf-cdk:

team-tf-cdk:

team-tf-cdk:

team-tf-cdk:

team-tf-cdk:

packages/@cdktf/hcl2cdk/lib/coerceType.ts
DanielMSchmidt: I added a check for the resource type to find variables and locals specifically

packages/@cdktf/hcl2cdk/lib/coerceType.ts
DanielMSchmidt: I added one now :)

packages/@cdktf/hcl2cdk/lib/generation.ts
DanielMSchmidt: We don't usually fail with ambigous content, so I would not like to add this here. I can add a debug log though

packages/@cdktf/hcl2cdk/lib/generation.ts
DanielMSchmidt: I would not call it faking. The path in general denotes where we are when we convert a value so we can lookup auxiliar type information. The extra `[]` denote that we have entered an array. This is important to know, so that if e.g. the parent type is `["array", "string"]` we now know once we went into another level of recursion that the new type we are concerned with is `"string"`

packages/@cdktf/hcl2cdk/test/__snapshots__/iteration.test.ts.snap
DanielMSchmidt: I looked into it and it's because of the order the parser returns on the resources, the resource does not exist yet at the point in time when the valueToTs function is called apparently. I think this is an more invasive change than I'd like to do in this PR so I'd say we scope this out into a follow up and discuss the prioritisation then. @mutahhir is that ok with you?

packages/@cdktf/hcl2cdk/lib/generation.ts
DanielMSchmidt: Added a comment

team-tf-cdk:

packages/@cdktf/hcl2cdk/test/__snapshots__/iteration.test.ts.snap
mutahhir: Sounds good to me. Let's create an issue for this.

mutahhir (approved):
Very nice! 👏🏽

PR #2633, opened 2023-02-14
fix(lib): zipmap arg validators

nikowatari:
when trying to use `Fn.zipmap` getting `Error: zipmap takes 1 arguments, but 2 were provided`
after this fix it works

### DIFF
--- a/packages/cdktf/lib/terraform-functions.ts
+++ b/packages/cdktf/lib/terraform-functions.ts
@@ -496,7 +496,12 @@ export class Fn {
     keyslist: any[] | IResolvable,
     valueslist: any[] | IResolvable
   ) {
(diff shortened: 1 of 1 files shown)


ansgarm (approved):
Thank you @nikowatari 👍

The error message you posted sounds like the opposite, though 😅 But I checked the [docs](https://developer.hashicorp.com/terraform/language/functions/zipmap) and they confirm that two arguments are expected 👍

ansgarm:
(fyi: force-pushed to fix the lint issue)

mutahhir (approved):

ansgarm:
Hi @nikowatari 👋
This has recently been released as part of `v0.15.5` – thanks again for the fix!

PR #2666, opened 2023-02-17
chore: Automate go package copyright/license inculsion

Maed223:
Added script that appends copyright headings to files in `cdktf` go package for compliance.

Unsure if there's a better way of including the script in the packaging process.


ansgarm (approved):
Looks good! Tried it locally and all headers were added as expected 💯

Issue #2649, opened 2023-02-14
Incorrect generation of Terraform expression with fqn to access property


mutahhir:
### Source

```hcl
# Documentation: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/availability_zones
data "aws_availability_zones" "changeme_az_list_ebs_snapshot" {
  state = "available"
}


# Documentation: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ebs_volume
# Explanation: The AWS 'aws_ebs_volume' resource is responsible for creating the EBS volume. The only required field is Availability Zone
# where the EBS vol is going to be created. Some optional arguments are Size in GBi, encryption(TRUE/FALSE) and he type of EBS volume. Can be "standard", "gp2", "io1", "sc1" or "st1" (Default: "standard").
resource "aws_ebs_volume" "changeme_ebs_volume_snapshot" {
  availability_zone = data.aws_availability_zones.changeme_az_list_ebs_snapshot.names[0]
  size              = 10
  type              = "standard"
  encrypted         = false
  tags = {
    Name = "changeme_ebs_volume_tag"
  }
}
```

### Synthed

```ts
    const awsEbsVolumeChangemeEbsVolumeSnapshot = new aws.ebsVolume.EbsVolume(
      this,
      "changeme_ebs_volume_snapshot",
      {
        availabilityZone: `\${${dataAwsAvailabilityZonesChangemeAzListEbsSnapshot.names.fqn}[0]}`,
        encrypted: false,
        size: 10,
        tags: {
          Name: "changeme_ebs_volume_tag",
        },
        type: "standard",
      }
    );
```

```py
        aws_ebs_volume_changeme_ebs_volume_snapshot = aws.ebs_volume.EbsVolume(self, "changeme_ebs_volume_snapshot",
            availability_zone=f"\\${{dataAwsAvailabilityZonesChangemeAzListEbsSnapshot.names.fqn}[0]}",
            encrypted=False,
            size=10,
            tags={
                "Name": "changeme_ebs_volume_tag"
            },
            type="standard"
        )
```

Note that `dataAwsAvailabilityZonesChangemeAzListEbsSnapshot.names.fqn` should not have been generated

PR #2671, opened 2023-02-23
chore: Fix generation of fqn

mutahhir:
Fixes: #2649


team-tf-cdk:

team-tf-cdk:

team-tf-cdk:

team-tf-cdk:

team-tf-cdk:

team-tf-cdk:

packages/@cdktf/hcl2cdk/test/astGeneration.test.ts
```diff
@@ -0,0 +1,140 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import { convertToTypescript } from "../lib";
+import schema from "./__fixtures__/schemas/astGeneration";
+
+describe("generate typescript", () => {
```
DanielMSchmidt, line 9: I understand where you are coming from here, but wouldn't it be easier to limit the test to just the expression parsing that you changed or if you want to check the generated code rely on the default testing mechanism we use throughout the hcl2cdk codebase?

packages/@cdktf/hcl2cdk/test/references.test.ts
```diff
@@ -242,3 +242,34 @@ describe("references", () => {
     { resources: ["aws_eip"] }
   );
 });
+
+describe("better fqn targeting", () => {
```
DanielMSchmidt, line 246: Can you try to describe what is better?

packages/@cdktf/hcl2cdk/lib/expressions.ts
```diff
@@ -82,46 +82,48 @@ export async function extractReferencesFromExpression(
...
-      // Can not use FQN on vars
-      !isVariable &&
-      // Can not use FQN on locals
-      !value.startsWith("local.") &&
-      // If the following character is
-      (input.substr(endPosition + 1, 1) === "*" || // a * (splat) we need to use the FQN
-        input.substr(endPosition, 1) === "[" || // a property access
-        isThereANumericAccessor || // a numeric access
-        fullReference.split(".").length < 3);
+    const canUseFqn =
```
DanielMSchmidt, line 94: I'm okay with this change, but I'm not sure if it's the right way forward. There could still be brittle / broken use-cases we just haven't uncovered. 

I would suggest we at least entertain the idea of using the parseExpression function in hcl2cdkto parse these possible references and using the resulting HCL-AST to answer the questions we have about the reference here :)

DanielMSchmidt (changes requested):
I'm open for push back on each of my comments :) Sorry for bringing them so late in the process

packages/@cdktf/hcl2cdk/lib/expressions.ts
mutahhir: Ah, that's an interesting idea. I'll give it a shot.

packages/@cdktf/hcl2cdk/test/astGeneration.test.ts
mutahhir: So, there are two problems with the current system, the biggest one by far being too slow. The reason why this test is in the middle of the two approaches is because we need a bit more context than what I have in `references.test.ts` below. I can still run this in a watch, while ensuring that the generated code is changing the way I expect it to.

The reason this test exists here is two-fold. Firstly because I wanted to have a way to check the generation when the expression being assigned (in the test below, `subnet_ids = ...`) is not within a Terraform templated string, does the code do the right thing? Secondly, because I wanted to have a template for this kind of test. 

I'm ok with having a longer discussion about this kind of test, if you still disagree with it and have a better alternative.

packages/@cdktf/hcl2cdk/test/references.test.ts
mutahhir: :) good call. Will do.

packages/@cdktf/hcl2cdk/test/astGeneration.test.ts
DanielMSchmidt: xkcd about competing standards

I'd like to have a longer conversation about his. I get where you are coming from, I think we can solve this for all the tests (eg by setting an environment variable and not synthesising for watched tests). This would help the entire codebase and not just that part.

ansgarm: My only concern about this topic would be to introduce environment variables into the codebase that would change behavior while testing, but I'm sure that's not what you plan to do, @DanielMSchmidt 😄

Issue #2670, opened 2023-02-22
Convert: variables / locals / referenced maps need to use accessor syntax


DanielMSchmidt:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Summary

Converting this example leads to an invalid access to the project key in the map, where a `Fn.lookup` call would have been necessary

```hcl
variable "default_tags" {
  type        = map(string)
  description = "Map of default tags to apply to resources"
  default = {
    project = "Learning Live with AWS & HashiCorp"
  }
}

resource "aws_eip" "nat" {
  vpc = true
  tags = {
    "Name" = "${var.default_tags.project}-nat-eip"
  }
}
```

ansgarm:
While this might not be that much better than `Fn.lookup` – we might think about exposing the [`PropertyAccess` expression](https://github.com/hashicorp/terraform-cdk/blob/857882f07071c471071dab6c414b6944a5d9188b/packages/cdktf/lib/tfExpression.ts#L193). The good thing about that way, would be that it supports a path, which might be more readable for nested access:
```ts
Fn.Lookup(Fn.Lookup(Fn.Lookup(mapRef, "c"), "b"), "a")
// vs.
Todo.propertyAccess(mapRef, ["a", "b", "c"])
```

(Sidenote: I left out the `defaultValue` param for lookup for simplicity, as that one is actually still optional but discouraged since TF 0.7; [docs](https://developer.hashicorp.com/terraform/language/functions/lookup))

ansgarm:
We might even expose this via TF functions 🤔
Maybe `Fn.property(mapRef, ["a", "b", "c"])` (like lodash) or some better wording. Another idea would be to do the following:
```ts
Fn.lookup(ref, property, defaultValue) // => lookup() TF call
Fn.lookup(ref, property) // => ref[property] TF expression
Fn.lookupNested(ref, [propertyA, propertyB]) // => ref[propertyA][propertyB] TF expression
```
What do you think? 🤔

DanielMSchmidt:
I think that's a good idea

PR #2672, opened 2023-02-23
feat(lib): Make default property in `Fn.lookup` optional and introduce `Fn.lookupNested`

ansgarm:
This PR implements the idea I had and posted in https://github.com/hashicorp/terraform-cdk/issues/2670#issuecomment-1440347032

### Todos
- [x] Add docs


mutahhir:
Github's acting weird on this PR (showing commits that are from the parent), so I'll wait to review once the parent PR is merged (or this PR is fixed).

ansgarm:
@mutahhir just rebased this PR – has a much cleaner diff now 👍

mutahhir (approved):
Approving, nice change!

One thing I'd like to see would be an update to the docs to introduce this, as I'm sure a lot of people would want to use it. Secondly, can I do array accesses for a specific property in here?

DanielMSchmidt (approved):

PR #2673, opened 2023-02-23
feat(cli): Upgrade init template to .NET 6.0

ansgarm:


mutahhir (approved):

ansgarm:
This still needs some fix (printing less warnings) to the documentation example

ansgarm:
Okay, so we still need JSII to release a new version with this [PR](https://github.com/aws/jsii/pull/3957) included before we can merge this change.
The underlying reason is that we get a lot of warnings in the form of
```
/Users/ansgar/.nuget/packages/microsoft.extensions.dependencyinjection.abstractions/7.0.0/buildTransitive/netcoreapp2.0/Microsoft.Extensions.DependencyInjection.Abstractions.targets(4,5): warning : Microsoft.Extensions.DependencyInjection.Abstractions 7.0.0 doesn't support netcoreapp3.1 and has not been tested with it. Consider upgrading your TargetFramework to net6.0 or later. You may also set <SuppressTfmSupportBuildWarnings>true</SuppressTfmSupportBuildWarnings> in the project file to ignore this warning and attempt to run in this unsupported configuration at your own risk. [/Users/ansgar/projects/hashicorp/terraform-cdk/examples/csharp/documentation/.gen/kubernetes/kubernetes.csproj]
```
which stem from the generated bindings still targetting the older .NET version – something that is fixed by the upstream JSII PR.

To get our documentation example to work, we'd need to adjust that generated code which would be very hacky (and is the reason why we can't update our C# template just yet).

I'll find a different way to fix this in the meantime and keep this PR waiting until we have a new JSII release.

PR #2679, opened 2023-02-24
feat(lib): Add TerraformCachedAsset – initial commit

ansgarm:


mutahhir (approved):
Good to go. Before we merge this, can you either update the Assets document or create an issue that we should add documentation about this cached assets feature?

packages/cdktf/lib/terraform-cached-asset.ts
```diff
@@ -0,0 +1,101 @@
...
+   *
+   * @param x
+   * @returns {boolean} whether the given object is a TerraformCachedAsset
+   */
+  static isTerraformCachedAsset(x: any): x is TerraformCachedAsset {
+    return x !== null && typeof x === "object" && CACHED_ASSET_SYMBOL in x;
+  }
+
+  /**
+   * @returns {Record<string, string>} The hash entries for this asset
```
ansgarm, line 78: The correct JSDoc syntax ([docs](https://jsdoc.app/tags-type.html)) would be this:
```suggestion
   * @returns {Object.<string, string>} The hash entries for this asset
```

packages/cdktf/lib/terraform-cached-asset.ts
```diff
@@ -0,0 +1,101 @@
...
+   */
+  public getHashEntries(): Record<string, string> {
```
ansgarm, line 80: Given that this is a public method exposed through the `cdktf` package, we should fix the name.

```suggestion
  public getHashEntry(): Record<string, string> {
```

ansgarm: hm, not sure about this, technically it could have multiple entries 😅 Feel free to ignore me here – I'm just unsure how to handle that this is publicly exposed while we only need it internally..

packages/cdktf/test/cached-assets.test.ts
```diff
@@ -0,0 +1,125 @@
...
+import { TerraformStack, Testing } from "../lib";
+import * as path from "path";
+import { TerraformCachedAsset } from "../lib/terraform-cached-asset";
+import * as fs from "fs";
+
+describe("cached asset", () => {
+  afterEach(() => {
+    try {
+      fs.rmSync(path.join(__dirname, "fixtures", "cached-asset", "output"), {
+        recursive: true,
```
ansgarm, line 12: By passing `force: true` ([docs](https://nodejs.org/api/fs.html#fsrmsyncpath-options)) we can get rid of the try-catch here – this would help us to debug this test case if something is weird with the filesystem (else the error will just get swallowed).

packages/cdktf/test/cached-assets.test.ts
```diff
@@ -0,0 +1,125 @@
...
+    expect(wasCalled).toHaveBeenCalled();
+    expect(oldHash).toEqual(cachedAsset.hash);
+
+    fs.writeFileSync(
+      path.join(__dirname, "fixtures", "cached-asset", "output", "foo.md"),
+      "hello world"
+    );
+
+    expect(oldHash).not.toEqual(cachedAsset.hash);
+  });
```
ansgarm, line 124: I think it would be valuable to add a test case that has an output directory that is nested within the input directory (which will be common for e.g. JS code that is compiled).

ansgarm: ... whoops, we already have such a case 🙈 Sorry for the noise and confusion!

packages/cdktf/test/cached-assets.test.ts
Maed223: 👍

packages/cdktf/lib/terraform-cached-asset.ts
Maed223: 👍

PR #2706, opened 2023-03-09
fix(hcl2cdk): handle JSII Rosetta translation properly

### LINKED ISSUE #2646: Use correct types instead of POD classes in strongly typed languages
As the following examples (for C#) shows:
```cs
            new aws.Provider.AwsProvider(this, "aws", new Struct
            {
                DefaultTags = new[] { new Struct {
                    Tags = new Struct {
                        CsTerraformExamples = "aws_db_instance/db_snapshot"
                    }
                } },
                Region = "us-east-1"
            });
```

The properties for the provider should've been of type `AwsProviderConfig`. Since in Typescript types can be enforced on plain objects, this approach works well without too overhead for the developer. However, for strongly typed languages, they expect parameters of a specific type. When JSII tries to convert the same snippet from Typescript:

```ts
    new aws.provider.AwsProvider(this, "aws", {
      defaultTags: [
        {
          tags: {
            csTerraformExamples: "aws_db_instance/db_snapshot",
          },
        },
      ],
      region: "us-east-1",
    });
```

For all practical purposes it looks like it _should_ become a `Struct`, but that's not accurate.

### LINKED ISSUE #2647: Use specific imports for used resources when converting
Currently convert generates a single import line per provider for typescript, something like so:

```ts
import * as aws from "../../.gen/providers/aws";
```

and then uses the following syntax for referencing resources / data sources:

```ts
    new aws.provider.AwsProvider(this, "aws", {
```

While this works fine for Typescript, it doesn't work for languages that don't allow * imports, like C#.

That import becomes:

```cs
using Gen.Providers.Aws;
```

and declaring the resource becomes:

```cs
            new aws.Provider.AwsProvider(this, "aws", new Struct
```

which does not work.

Apart from it being broken on C#, there are performance implications for doing a * import for large providers. Thus, this is a performance improvement as well.

```cs
            new aws.Provider.AwsProvider(this, "aws", new Struct
            {
// ...
            });
```

DanielMSchmidt:
### Related issue

Fixes #2646 

### Description

JSII Rosetta needs to have compilable code in order to determine the correct classes to use in Non-TS languages.
This PRs aims to do this for cases where we have node.js bindings in the same folder this runs in. This is the step on the way of making it work in all cases

### Follow up work

I decided against doing a full test setup here that synthesises across all languages to cut scope from this PR, I'd suggest we do this in separate follow up PRs.

### Checklist


team-tf-cdk:

DanielMSchmidt:
This should also have fixed #2647 but the snapshots beg to differ for some languages, I'll double check this with the test repository to see if the output is the same. For now we will consider #2647 a separate issue

team-tf-cdk:

mutahhir (approved):
Nice change 👏🏽

PR #2714, opened 2023-03-10
fix(cli): Give hint when cdktf synth failed to create config

ansgarm:
There was a recent case where someone on the CDK Dev Slack got this error and it took them a while to figure out that they were missing the `app.synth()` call which is probably the most common mistake leading to this error.

The current error message is not really helpful in that regard:
image

So this PR adds a hint.

### DIFF
--- a/packages/@cdktf/cli-core/src/lib/synth-stack.ts
+++ b/packages/@cdktf/cli-core/src/lib/synth-stack.ts
@@ -154,7 +154,7 @@ Command output on stdout:
     }
 
     if (!(await fs.pathExists(path.join(outdir, Manifest.fileName)))) {
-      const errorMessage = `ERROR: synthesis failed, app expected to create "${outdir}/${Manifest.fileName}". Make sure your CDKTF App is synthesized.`;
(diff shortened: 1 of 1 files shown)


packages/@cdktf/cli-core/src/lib/synth-stack.ts
```diff
@@ -154,7 +154,7 @@ Command output on stdout:
     }
 
     if (!(await fs.pathExists(path.join(outdir, Manifest.fileName)))) {
-      const errorMessage = `ERROR: synthesis failed, app expected to create "${outdir}/${Manifest.fileName}"`;
+      const errorMessage = `ERROR: synthesis failed, app expected to create "${outdir}/${Manifest.fileName}". Make sure your CDKTF App is synthesized.`;
```
mutahhir, line 157:
```suggestion
      const errorMessage = `ERROR: synthesis failed, because app was expected to call 'synth()`, but didn't. Thus "${outdir}/${Manifest.fileName}"  was not created.`;
```

DanielMSchmidt (approved):

PR #2723, opened 2023-03-14
chore: improve development performance by excluding synth

DanielMSchmidt:
### Related issue

None

### Description

This removes the synths from the local test invocations by default, but not during CI

### Checklist

### DIFF
--- a/packages/@cdktf/hcl2cdk/README.md
+++ b/packages/@cdktf/hcl2cdk/README.md
@@ -77,21 +77,6 @@ This transforms your Terraform project into a CDK for Terraform project, besides
 
 ## Known Limitations
 
-### Terraform Expressions are of the wrong TS/Java/C# type
-
-When working with typed languages the converter can run into problems where the Terraform Expression evaluates to a certain type but it's encoded in a string. Therefore the type checker of the language detects a type mismatch, resulting in an compilation error. These problems need to be manually solved by adding a typecast. One example would be:
-
-```ts
-{
-  // Typescript expects a boolean here, but a string is passed
-  booleanProperty:
-    ( // This Terraform expression evaluates to a boolean as required by the property
-      `\${${shouldBeTrue.value} ? true : false}`
-      as boolean // We need to cast the type so that Typescript understand the right type is being passed
-    );
-}
-```
-
 ### Providers guessed can be not functional
 
 If your HCL includes providers that are not mentioned under `required_providers` we infer the name, e.g. if you use the `datadog_dashboard` resource we infer the provider `datadog` which is right, but the namespace is missing, for DataDog it would be `datadog/datadog`. Instead we will try to use `hashicorp/datadog` and fail because this provider is not known to the registry.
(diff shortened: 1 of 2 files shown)


packages/@cdktf/hcl2cdk/README.md
```diff
@@ -100,3 +85,11 @@ Please see the [required providers docs](https://www.terraform.io/docs/language/
...
 
 We don't move modules or files for you, if you reference local modules you have to move them so that the relative paths are correct. If you want to make use of files you need to wrap them in a [`TerraformAsset`](../../docs/working-with-cdk-for-terraform/terraform-assets.md) before using them.
+
+### Development
+
+We have two types of test cases, one within `lib` that are on the unit level and one within `test` that are testing the entire package at once by converting and then synthesizing the resulting code.
+
+In general, both test types can be run by `npx jest <pathToTestCase>`. You can add `-u` to update the snapshots and `--watch` to run the tests in watch mode.
+
+To make the tests inside `test` faster we disable synthesizing and multi-language snapshots by default. You can enable them by setting the envinronment variable `CI=true`. Another way of improving the performance significantly is setting the `TF_PLUGIN_CACHE_DIR` to a valid directory in order to cache the provider binaries used within the tests. E.g. by running `TF_PLUGIN_CACHE_DIR=(mktemp -d) npx jest <pathToTestCase> --watch`.
```
ansgarm, line 95: Nice tip with the plugin cache!

packages/@cdktf/hcl2cdk/test/helpers/convert.ts
```diff
@@ -296,42 +302,48 @@ const createTestCase =
         }, 500_000);
 
         if (
-          shouldSynth === Synth.yes ||
-          shouldSynth === Synth.yes_all_languages
+          includeSynthTests &&
```
ansgarm, line 305: One alternative that could make this flag a bit more visible would be to render `it.skip` if the flag is set. That way it might be easier to discover this behavior (even if documented) as a contributor if tests fail in CI while running locally seems fine. That way Jest would at least print the skipped tests.

packages/@cdktf/hcl2cdk/test/helpers/convert.ts
```diff
@@ -15,9 +15,10 @@ import {
   ConstructsMakerProviderTarget,
   readSchema,
 } from "@cdktf/provider-generator";
-
 import deepmerge from "deepmerge";
 
+const includeSynthTests = Boolean(process.env.CI);
```
ansgarm, line 20: My personal opinion would be to invert how this works and to make tests that run synth opt-out instead. We could e.g. add this opposite flag to the NPM `watch` script to make it easier to use. To me the default behavior should be the one that runs more tests and is as closely resembling what happens in CI as possible. Doing different things locally than in CI by default is a recipe for struggle (especially for contributors trying to run tests locally).

ansgarm (commented):
I personally would rather see this behaviour as opt-in, but I don't have a strong opinion here and am curious why you and @mutahhir prefer this one.

packages/@cdktf/hcl2cdk/test/helpers/convert.ts
DanielMSchmidt: I have no strong opinion here, but I struggled with folks not using my carefully crafted tools and flags if they were not the default for local development, that's why I biased in the favour of the fast solution. 

Summoning @mutahhir as the tie breaker.

mutahhir: ### Ahem

I hereby declare the tie broken! 

I would also prefer the local development thing be default, and using CI as a special case. Primarily because I'll forget to set the flag and will also not remember what that flag was named, whereas CI has a pretty good memory.

ansgarm: I'm fine with going that path if the two of you prefer a different default :)

ansgarm (approved):

PR #2724, opened 2023-03-15
refactor(hcl2json): rename wasm method to match its implementation and expose it more direct aswell

ansgarm:
This PR prepares the library for the convert work that's going to happen to support e.g. replacing functions and operations.

- fix(hcl2json): correctly check for empty array
- refactor(hcl2json): rename wasm method to match its implementation and expose it more direct aswell

### DIFF
--- a/packages/@cdktf/hcl2json/lib/index.ts
+++ b/packages/@cdktf/hcl2json/lib/index.ts
@@ -98,7 +98,7 @@ export async function parse(
 
 export async function convertFiles(
   workingDirectory: string
-): Promise<Record<string, any> | void> {
+): Promise<Record<string, any>> {
   let tfFileContents = "";
   const tfJSONFileContents: Record<string, any>[] = [];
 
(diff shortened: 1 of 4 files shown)


packages/@cdktf/hcl2json/lib/index.ts
```diff
@@ -114,7 +114,7 @@ export async function convertFiles(
     }
   }
 
-  if (tfFileContents === "" && tfJSONFileContents === []) {
+  if (tfFileContents === "" && tfJSONFileContents.length === 0) {
```
ansgarm, line 117: whoops – this appears to have broken tests 😅
We can also remove this entire thing 😄 What do you think?

ansgarm: Fixed it by returning an empty object as required by the test. Also adjusted the testcase to check that the error is actually printed

DanielMSchmidt (approved):

Issue #1852, opened 2022-06-07
Convert: use native dynamic expressions instead of overrides


DanielMSchmidt:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Currently we synthesize an escape hatch override if we detect count or for each being used. Instead we should use functionality created in #1797 and #1798

DanielMSchmidt:
Closed in #2732

PR #2732, opened 2023-03-17
feat(hcl2cdk): use TerraformIterator API instead of overrides

DanielMSchmidt:
### Related issue

Preparation for #1852 

### Description

Use `TerraformIterator` in cdktf converted code to prepare using the iterators for dynamic blocks.

### Checklist

- [X] I have updated the PR title to match [CDKTF's style guide](https://github.com/hashicorp/terraform-cdk/blob/main/CONTRIBUTING.md#pull-requests-1)
- [X] I have run the linter on my code locally
- [X] I have performed a self-review of my code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation if applicable
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works if applicable
- [X] New and existing unit tests pass locally with my changes


packages/@cdktf/hcl2cdk/test/__snapshots__/iteration.test.ts.snap
```diff
@@ -27,40 +27,54 @@ class MyConvertedCode extends cdktf.TerraformStack {
       });
     /*This allows the Terraform resource name to match the original name. You can remove the call if you don't need them to match.*/
     dataAwsRoute53ZoneExample.overrideLogicalId(\\"example\\");
+    /*In most cases loops should be handled in the programming language context and 
+    not inside of the Terraform context. If you are looping over something external, e.g. a variable or a file input
+    you should consider using a for loop. If you are looping over something only known to Terraform, e.g. a result of a data source
+    you need to keep this like it is.*/
+    const awsRoute53RecordExampleForEachIterator =
+      cdktf.TerraformIterator.fromList(
+        cdktf.Token.asAny(\`\\\\\${{
```
DanielMSchmidt, line 36: We could think about using cdktf.Token.asList here

packages/@cdktf/hcl2cdk/test/__snapshots__/iteration.test.ts.snap
```diff
@@ -27,40 +27,54 @@ class MyConvertedCode extends cdktf.TerraformStack {
       });
     /*This allows the Terraform resource name to match the original name. You can remove the call if you don't need them to match.*/
     dataAwsRoute53ZoneExample.overrideLogicalId(\\"example\\");
+    /*In most cases loops should be handled in the programming language context and 
+    not inside of the Terraform context. If you are looping over something external, e.g. a variable or a file input
+    you should consider using a for loop. If you are looping over something only known to Terraform, e.g. a result of a data source
+    you need to keep this like it is.*/
+    const awsRoute53RecordExampleForEachIterator =
+      cdktf.TerraformIterator.fromList(
```
DanielMSchmidt, line 35: In general we should be using e.g. `fromMap` here as well, but we have little chance to understand if the input value is closer to a list or a map. I was thinking about adding this to the comment to let the user change the code slightly if needed

packages/@cdktf/hcl2cdk/lib/generation.ts
```diff
@@ -416,8 +438,12 @@ function mapConfigPerResourceType(resource: string, item: Resource[0]) {
   return item;
 }
 
+const loopComment = `In most cases loops should be handled in the programming language context and 
+not inside of the Terraform context. If you are looping over something external, e.g. a variable or a file input
```
ansgarm, line 442: The word "variable" is a bit too ambiguous 😅 I'd expect most practitioners using convert to think if "Terraform Variables" which would require iterators – Can we make this more clear somehow?

packages/@cdktf/hcl2cdk/test/__snapshots__/iteration.test.ts.snap
```diff
@@ -27,40 +27,54 @@ class MyConvertedCode extends cdktf.TerraformStack {
...
-          }
-        }}\`
-    );
     const awsAcmCertificateValidationExample =
       new aws.acmCertificateValidation.AcmCertificateValidation(
         this,
         \\"example_4\\",
         {
-          certificateArn: awsAcmCertificateExample.arn,
+          certificateArn: cdktf.Token.asString(awsAcmCertificateExample.arn),
```
ansgarm, line 77: What caused this to appear? 🤔 It should not need to exist, does it?

packages/@cdktf/hcl2cdk/test/__snapshots__/iteration.test.ts.snap
```diff
@@ -189,37 +203,36 @@ class MyConvertedCode extends cdktf.TerraformStack {
...
-    );
+    const awsS3BucketObjectExamplebucketObjectForEachIterator =
+      cdktf.TerraformIterator.fromList(
+        cdktf.Token.asAny(\`\\\\\${toset(\${awsS3BucketExamplebucket.fqn}.*)}\`)
+      );
+    new aws.s3BucketObject.S3BucketObject(this, \\"examplebucket_object\\", {
+      bucket: awsS3BucketObjectExamplebucketObjectForEachIterator.key,
+      key: \\"someobject\\",
+      kmsKeyId: cdktf.Token.asString(awsKmsKeyExamplekms.arn),
+      source: \\"index.html\\",
```
ansgarm, line 234: (This is true for all examples) For this to work, the iterator has to be passed to the `forEach` property of the resource as well ([related docs with example](https://developer.hashicorp.com/terraform/cdktf/concepts/iterators#when-to-use-iterators)).

packages/@cdktf/hcl2cdk/test/__snapshots__/jsiiLanguage.test.ts.snap
```diff
@@ -285,7 +285,7 @@ class MyConvertedCode extends cdktf.TerraformStack {
     new aws.s3BucketObject.S3BucketObject(this, \\"examplebucket_object\\", {
       bucket: \`\\\\\${element(\${awsS3BucketExamplebucket.fqn}, 0).id}\`,
       key: \\"someobject\\",
-      kmsKeyId: awsKmsKeyExamplekms.arn,
+      kmsKeyId: cdktf.Token.asString(awsKmsKeyExamplekms.arn),
```
ansgarm, line 288: same as above, this seems like an unwanted side effect 🤔

ansgarm (changes requested):
Looks good so far, I noticed one thing missing from the converted code to actually work, which is passing the iterator itself to resources via `forEach` 😅

packages/@cdktf/hcl2cdk/lib/expressions.ts
```diff
@@ -129,7 +129,32 @@ export async function extractReferencesFromExpression(
   }, [] as Reference[]);
 }
 
-export function referenceToVariableName(scope: Scope, ref: Reference): string {
+export type IteratorVariable = {
```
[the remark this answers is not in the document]
DanielMSchmidt: Yes it is

packages/@cdktf/hcl2cdk/lib/generation.ts
DanielMSchmidt: Let's change the text in a different PR, I just moved it but is has been like this before. We can definitely change it but it's not in the scope of this PR.

packages/@cdktf/hcl2cdk/test/__snapshots__/iteration.test.ts.snap
DanielMSchmidt: I think I somehow caused a new data race condition. We need to do some extra work beforehand to get this sorted out in general. Not sure if it's a priority for 0.16 though

packages/@cdktf/hcl2cdk/test/__snapshots__/jsiiLanguage.test.ts.snap
DanielMSchmidt: Yes, but it's okay, it's likely related to the order of execution slightly changing

packages/@cdktf/hcl2cdk/test/__snapshots__/iteration.test.ts.snap
DanielMSchmidt: Yes, did that now.

Issue #2024, opened 2022-08-12
Support accessing resources created by an iterator


jsteinich:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

Iterators allow you to create a dynamic amount of resources based on data only know at Terraform execution time; however, if you then need to reference those resources for some other purpose, there isn't a supported way (besides escape hatches). This is still dynamic information so not everything will be possible, but some operations should still work. We could support creating another iterator from the resource which could then be feed to another resource. We may also be able to expose list/map operations for direct access (via Terraform lookups).

### References

- https://discuss.hashicorp.com/t/how-to-iterate-over-objects-made-with-an-iterator/43115
- https://github.com/hashicorp/terraform-cdk/issues/1094

PR #2739, opened 2023-03-22
feat(lib): TerraformIterators allow chaining

DanielMSchmidt:
This won't solve the mismatch between programming language types and internal type of the reference (e.g. accessing a string property on an object using for_each will be a string in from the type systems of the programming languages, but a list of strings for terraform). This is not worse than what we currently have (an invalid Terraform reference), therefore I think we can go forward with this.

Relates to #2024

### DIFF
--- a/packages/cdktf/lib/terraform-data-source.ts
+++ b/packages/cdktf/lib/terraform-data-source.ts
@@ -155,7 +155,9 @@ export class TerraformDataSource
 
   public interpolationForAttribute(terraformAttribute: string) {
     return ref(
-      `data.${this.terraformResourceType}.${this.friendlyUniqueId}.${terraformAttribute}`,
+      `data.${this.terraformResourceType}.${this.friendlyUniqueId}${
+        this.forEach ? ".*" : ""
+      }.${terraformAttribute}`,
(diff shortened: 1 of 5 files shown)


ansgarm (commented):
Looks good! Could be merged from my POV (but does not solve everything).

The only thing that is still off, is accessing e.g. `.stringAttribute` will now actually return a reference to a list of strings (via `.*.string_attribute`) if an iterator is used. But that is not worse than the current implementation, which will just produce invalid Terraform (by trying to access `.string_attribute` which does not work if iteration is used 😅).

So, if someone e.g. does something like this:
```
const s3buckets = new S3Bucket(this, "bucket", {forEach: someIterator, ... } );

new TerraformOutput(this, "bucketNames", { value: s3Buckets.name })
```

They'll need to wrap `s3Buckets.name` in `Token.asList()` whenever they would want to pass that somewhere were there's a list required. And similarly, passing `s3Buckets.name` to something that takes a string, will appear to work but eventually fail in the Terraform context.

ansgarm (approved):
Won't resolve #2024 but the improvement is worthwile 👍

PR #2772, opened 2023-04-05
fix(tests): set JSII env var to not print Node version warning

ansgarm:
as that one caused our tests to fail


DanielMSchmidt (approved):

ansgarm:
merging w/o waiting for CI as the tests in question did succeed 👍

PR #2801, opened 2023-04-14
fix: Use property access when refrencing numeric and map accessors

mutahhir:
### Related issue

Fixes https://github.com/orgs/hashicorp/projects/207/views/12?pane=issue&itemId=25543682

### Description

Using String templates instead of propertyAccess causes problems with nesting template strings (which would need more context to solve properly), and also when used within CDKTF functions. This PR moves most template string usage that required use of FQN over to using the CDKTF propertyAccess function.

### Checklist


packages/@cdktf/hcl2cdk/test/expressionToTs.test.ts
```diff
@@ -417,7 +417,7 @@ describe("expressionToTs", () => {
       getType
     );
     expect(code(result)).toMatchInlineSnapshot(
-      `""\${" + awsS3BucketExamplebucket.fqn + "}.network_interface[0].access_config[0].assigned_nat_ip""`
+      `"cdktf.Token.asString(cdktf.propertyAccess(awsS3BucketExamplebucket, ["network_interface", "0", "access_config", "0", "assigned_nat_ip"]))"`
```
DanielMSchmidt, line 420: I think this would be a nicer goal to target, making everything possible access via TS and only then fall back to property access. What do you think, is this out of scope?

```suggestion
    `"cdktf.Token.asString(cdktf.propertyAccess(awsS3BucketExamplebucket.networkInterface, ["0", "access_config", "0", "assigned_nat_ip"]))"`
```

mutahhir: Fixed now.

ansgarm (approved):

PR #2804, opened 2023-04-14
fix(hcl2cdk): handle object expressions

DanielMSchmidt:


packages/@cdktf/hcl2cdk/lib/expressions.ts
```diff
@@ -451,11 +453,24 @@ function convertTFExpressionAstToTs(
     return expressionForSerialStringConcatenation(expressions);
   }
 
+  if (tex.isObjectExpression(node)) {
+    return t.objectExpression(
+      await Promise.all(
```
mutahhir, line 458: I'm wondering if using this kind of construct to find the child for the item value as we do here: https://github.com/hashicorp/terraform-cdk/blob/main/packages/%40cdktf/hcl2cdk/lib/expressions.ts#L621-L624 might also be a solution? I'm not sure if the children are there or if the value is only present in `items`, so ignore if you tried this and it doesn't work

DanielMSchmidt: Did not work right of the bat, let's revisit this later

mutahhir (approved):

PR #2812, opened 2023-04-18
fix(hcl2cdk): Handle multi-line strings better

mutahhir:
### Related issue

Fixes https://github.com/orgs/hashicorp/projects/207/views/12?pane=issue&itemId=25824022

### Description

This is a pretty hacky PR, and I'm not really comfortable with the solution. That being said, I think this does resolve our immediate woes around heredocs. 

The problem is as such: By the time the heredoc string comes in for parsing by the expression parser, (I suspect) the hcl2json parser has already converted it into a quote containing newlines. Quote with newlines are not accepted within HCL and thus breaks our expression. 

This PR tries to detect newlines within string literals and wraps them back into a heredoc literal, making the expression parseable. As you can see from the PRs, this requires a lot of tweaks to existing logic to ensure it works ok, and even then, I'm not 100% sure this will not be caught off guard on other situations.

### Checklist

### DIFF
--- a/packages/@cdktf/hcl2json/lib/util.ts
+++ b/packages/@cdktf/hcl2json/lib/util.ts
@@ -1,3 +1,8 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
 export function wrapTerraformExpression(input: string): {
   wrap: string;
   wrapOffset: number;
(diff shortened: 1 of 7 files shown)


packages/@cdktf/hcl2json/lib/util.ts
```diff
@@ -0,0 +1,43 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+export function wrapTerraformExpression(input: string): {
```
ansgarm, line 6: (optional nit) a small doc string with an example would help a lot

ansgarm (approved):

packages/@cdktf/hcl2json/lib/util.ts
mutahhir: Good point! I'll do it in the next PR i'm working on. Will continue like this for now.

Issue #2841, opened 2023-04-26
Property 'StringListList' does not exist on type .../node_modules/cdktf/lib/index


abaerptc:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### cdktf & Language Versions

  version : latest
  language: python
  cdktf   : 0.16.0

### Affected Resource(s)

cdktf provider add eabrouwer3/airbyte

### Debug Output

https://gist.github.com/abaerptc/9d49d823781afef266df8629b65026f0

### Expected Behavior

Provider should be added.

### Actual Behavior

Error:
```
providers/airbyte/data-airbyte-source-schema-catalog/index.ts:150:48 - error TS2339: Property 'StringListList' does not exist on type 'typeof import("/private/var/folders/yr/rt84nf717mzdwj9rxfm5vtrr0000gp/T/temp-r5aVSo/node_modules/cdktf/lib/index")'.
```

### Steps to Reproduce

```
cdktf init --template=python
# Answer prompts to create template

cdktf provider add eabrouwer3/airbyte
```

### References

- #0000

ansgarm:
Hi @abaerptc 👋

Thank you for reporting this, I was able to reproduce it locally.

The provider seems to define the failing attribute as
```json
"type": [
  "list",
  [
    "list",
    "string"
  ]
],
```

and apparently our code generation process fails on this – we'll need to look into why that is as it should be supported.

PR #2849, opened 2023-04-28
test(provider-generator): Add failing test case for #2841

ansgarm:


ansgarm:
This is obsolete as it was cherry-picked into https://github.com/hashicorp/terraform-cdk/pull/2850 and merged through that PR

PR #2857, opened 2023-05-04
chore: Use tsx instead of ts-node

mutahhir:
### Description

During the hackathon, I wanted to see if we can improve performance for synth. On one of the larger stack projects that I had, I was seeing about a 57 second synth time. However, when I went to profile the synth function, it was only taking up a couple of seconds. Turns out, most of the time is taken up by `ts-node`. As an alternative, there's [tsx](https://github.com/esbuild-kit/tsx) which is based on [esbuild](https://esbuild.github.io/). With `tsx`, the synth time went down to 5 seconds. 

### Checklist


ansgarm (commented):
In general I'm in favor of switching to `tsx`, however `esbuild` is not a drop in replacement for TypeScript (e.g. it doesn't implement its type system which means it does not type check the code, [docs](https://esbuild.github.io/content-types/#typescript)) which might have some implications / drawbacks that we didn't think about yet.

DanielMSchmidt (changes requested):
Let's not do the template yet, maybe we can add the examples as well

DanielMSchmidt (approved):

ansgarm:
I did look into this some more and it seems that 60s synth times happen when TypeScript projects use star imports (could you confirm whether this is the case for your large project, @mutahhir?).

The same 60s synth project went down to a few seconds after changing the imports to specific ones. I'll create a PR to do this in our examples (e.g. the `aws-prebuilt` example uses a star import).

That said, I would say improving this would be an documentation exercise, possibly paired with some help our tooling:
One quick idea would be to set `global.CDKTF_STAR_IMPORT_USED = true` in the `index.ts` of provider bindings. We could then check whether this global is `true` during synth and print a warning (e.g. if synth takes longer than e.g. 15s) which could link to some new docs detailing how to improve TypeScript performance.

Furthermore, we could also opt-out of type checking by passing the `-T` flag or the `--swc` flag to `ts-node` which both will speed up the synth time (With `swc`, which is similar to `esbuild`, my 60s example took 10s [while keeping the slow star import]). Both these options are something we could document.

If we want type checking and faster builds on subsequent runs, we'd probably have to switch back to `tsc` as `ts-node` doesn't support any caching.

One other alternative to mention: @DanielMSchmidt has done some discovery on using an `index-lazy.ts` file, I think it was specifically for improving the performance of JSII enabled languages by loading less code into Node.js, and trying this locally in TypeScript also gave me drastic improvements in performance which came close to the non-star-imports. Type information get's lost when doing these `Object.defineProperty(exports, 'provider', { get: function () { return require('./provider'); } });` definitions, so we'd have to generate some `.d.ts` file probably.
This might be the best way forward, but might carry more work and problems with possible incompatibilities (i.e. some users use Webpack or other tools to compile their CDKTF projects).

PR #2887, opened 2023-05-24
feat(cdktf): add support for testing with Vitest

### LINKED ISSUE #2886: Support more testing frameworks than Jest for TypeScript
### Community Note

### Description

`cdktf` currently [only supports testing with Jest](https://developer.hashicorp.com/terraform/cdktf/test/unit-tests#add-testing-to-your-application) when using TypeScript.

[Vitest](https://vitest.dev/) is a popular testing framework that is intended as a drop-in replacement for Jest, so the `expect` function can be extended in basically the exact same way. However, since `cdktf` currently only provides the `setupJest()` function which only works with the global expect function from Jest it needs to be done manually via the exported `testingMatchers` at the moment.

I think it could be a good idea to formally document the exported `testingMatchers` to make it easier for others to create custom adapters for other testing frameworks 👍 

### References

- [Extending Matchers | Vitest](https://vitest.dev/guide/extending-matchers.html)

duniul:
### Related issue

Fixes #2886

### Description

[Vitest](https://vitest.dev/) is a popular testing framework that is intended as a drop-in replacement for Jest, so the `expect` function can be extended in basically the exact same way. I've tried to match the current APIs and ensure backward compatibility.

This PR:
- Adds support for Vitest in a similar way to how Jest is currently supported
  - Adds a `Testing.setupVitest()` function, which internally uses the same function as `Testing.setupJest()` to extend the `expect` function. Supports both global and imported `expect` from Vitest.
  - Does not change the current Jest behavior, simply extracts and reuses the current logic from `setupJest()`
- Moves `expect` type extensions to `types/jest.d.ts` and `types/vitest.d.ts` respectively, instead of declaring them directly inside the adapter source file.
  - This way, the types can included in `types` array in `tsconfig.json` to avoid having import them directly in every test file.
  - `lib/testing/adapters/jest.ts` still imports and declares the files directly, to stay backwards compatible with the current approach.
- Updates the website documentation to reflect the changes listed above.

### Not included

I have not included any tests for the Vitest expect similar to the ones that exists for Jest, since that would require adding Vitest as a dev dependency just for that test. However, they should be identical in behavior to the one for Jest.

If you'd like me to include this anyway then let me know and I'll add it 👍

### Checklist

### DIFF
--- a/packages/cdktf/lib/testing/adapters/vitest.ts
+++ b/packages/cdktf/lib/testing/adapters/vitest.ts
@@ -1,3 +1,8 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
 import { ExpectLike, extendExpect } from "../expect";
 
 /**


ansgarm:
Hi @duniul 👋

Thank you for your nice work on this and the stellar description 👏 Much appreciated!

While Vitest seems to be a seamless drop-in replacement for Jest, we'd still need to ensure that it continues to work over time and therefore maintain that added complexity. Given our five different languages we already support, it doesn't make a lot of sense for us to maintain two testing frameworks that only work for TypeScript projects.

However, we'd be open to a PR that makes the changes required to enable the Vitest specific bits that could then e.g. live in a plugin that could be maintained by the community. So basically the `expect.ts` refactoring that would then be available to an external package that sets up Vitest.

duniul:
@ansgarm Yeah, that makes sense 👍 As I wrote under the "Alternative solution" in the [issue](https://github.com/hashicorp/terraform-cdk/issues/2886) I think it's doable already, it's just not clear if that's the intention of the exported `testingMatchers`.

I'll have a look! Closing this for now, I'll open another PR if needed.

duniul:
Got it working alright with the exported `testingMatchers`, so I published [cdktf-vitest](https://www.npmjs.com/package/cdktf-vitest) 👍

PR #2891, opened 2023-05-27
fix(lib): Assign `TextEncoder` and `TextDecoder` to `globalThis` conditionally

zharinov:
### Description

Hello amazing Terraform folks and the mantainers of this repo in particular.

I've noticed a failure with WASM exec bridge code that works on Node 18, but fails on Node 20. Seems like `TextEncoder` and `TextDecoder` already exist in new versions of Node, and setting these for the second time will fail because the properties are read-only.

### Checklist

Sorry for not checking all the boxes, I hope my humble changes don't require a full local setup on my side.


DanielMSchmidt (approved):

mutahhir:
I'm trying to replicate the failure but can't. We're using [this file](https://github.com/golang/go/blob/master/misc/wasm/wasm_exec_node.js) as a reference, and that too seems to not have this change.

zharinov:
Sorry, this bug has gone in newer versions of `node@20` 🤷

PR #2909, opened 2023-06-01
chore: add merge queue and remove timechart

DanielMSchmidt:


.github/workflows/pr-unit.yml
mutahhir, line 1: Why don't we do it on others, like `pr-integration`, and `pr-provider-integration` etc?

mutahhir (approved):
Approving, but had one question about coverage of tests

.github/workflows/pr-unit.yml
DanielMSchmidt: Because they run are merge anyways. We decided a while back to not run them on PRs. We can change that though, no strong feelings

PR #2914, opened 2023-06-01
feat(cli): add flag to display memory / time used for get and synth

DanielMSchmidt:
These commands take often some time and space, the flag will help us measure improvements

### DIFF
--- a/packages/cdktf-cli/src/bin/cmds/handlers.ts
+++ b/packages/cdktf-cli/src/bin/cmds/handlers.ts
@@ -61,6 +61,7 @@ import {
 } from "./helper/check-environment";
 import { sanitizeVarFiles } from "./helper/var-files";
 import { askForCrashReportingConsent } from "./helper/error-reporting";
+import { startPerformanceMonitoring } from "./helper/performance";
 
 const chalkColour = new chalk.Instance();
 const config = readConfigSync();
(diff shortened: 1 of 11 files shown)


packages/cdktf-cli/src/bin/cmds/handlers.ts
```diff
@@ -236,38 +237,47 @@ export async function get(argv: {
...
-
-  const constraints: TerraformDependencyConstraint[] = [
-    ...providers.map((c) => new TerraformProviderConstraint(c)),
-    ...modules.map((c) => new TerraformModuleConstraint(c)),
-  ];
-
-  if (constraints.length === 0) {
-    logger.warn(
-      `WARNING: No providers or modules found in "cdktf.json" config file, therefore cdktf get does nothing.`
+  const printPerformanceInfo = argv.showPerformanceInfo
```
mutahhir, line 242: This `printPerformanceInfo = startPerformanceMonitoring()` makes me feel odd, but I also don't have a much nicer solution here.

packages/cdktf-cli/src/bin/cmds/helper/performance.ts
```diff
@@ -0,0 +1,45 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import pidtree from "pidtree";
+import pidusage from "pidusage";
+
+async function getCombinedMemoryUsageInBytes(
```
mutahhir, line 9: A comment here for the future would be nice. I assume we're doing this because JSII pacmak does this multi-process invocation for doing the conversion, right?

packages/cdktf-cli/src/bin/cmds/helper/performance.ts
```diff
@@ -0,0 +1,45 @@
...
+    (carry, item) => (item && "memory" in item ? carry + item.memory : carry),
+    0
+  );
+}
+
+export function startPerformanceMonitoring() {
+  const start = Date.now();
+  const memoryMeasurementsInGb: number[] = [];
+
+  const interval = setInterval(async () => {
```
mutahhir, line 24: nit: Overall, not a big fan of `setInterval`, but I can see how it's convenient right now.

packages/cdktf-cli/src/bin/cmds/helper/performance.ts
```diff
@@ -0,0 +1,45 @@
...
+  const allPids = await pidtree(pid, { root: true });
+  const usages = await pidusage(allPids);
+  return Object.values(usages).reduce(
+    (carry, item) => (item && "memory" in item ? carry + item.memory : carry),
+    0
+  );
+}
+
+export function startPerformanceMonitoring() {
+  const start = Date.now();
```
mutahhir, line 21: Nit: Should we continue to use `performance.now()` as we were doing [here](https://github.com/hashicorp/terraform-cdk/pull/2914/files#diff-711f3bf442c54e74867a9fa55969ab555b04f1bd03bd21b511bcddae7f805136L15). I do understand that the timelines we're interested in are much longer than performance.now() was designed for, so feel free to ignore my comment.

mutahhir (approved):
Nice! Minor nits, but can be merged as-is

packages/cdktf-cli/src/bin/cmds/helper/performance.ts
DanielMSchmidt: Not sure what you would replace it with here, I kinda need an interval :D

packages/cdktf-cli/src/bin/cmds/helper/performance.ts
DanielMSchmidt: Yeah I was thinking about that as well and chose Date.now since we don't need the fine grained resolution and it's in ms so a bit easier to calculate to seconds there :)

PR #2924, opened 2023-06-09
chore: remove update snapshots CI job

DanielMSchmidt:
Marie Kondo - does it spark joy meme

I'd like to propose to remove this CI job, we very rarely use it and when we do it's way too slow and fragile to be practical.


mutahhir (approved):
🗑️

PR #2925, opened 2023-06-09
chore: remove PR sizing

DanielMSchmidt:
Marie Kondo cleans up

It did not help us in the way we thought it would, the comments are annoying, let's get rid of it.


mutahhir (approved):
🗑️

PR #2926, opened 2023-06-09
chore: pin all workflows after update

xiehan:
This adds a step to the automated GitHub Action update job to pin all the workflows. If a new version hasn't been added to TSCCR yet, this effectively undoes the upgrade, the idea being the auto-update script will run again in a week, and hopefully in the meantime the new version will have been added to TSCCR.

Note: hasn't been tested, as with all my GitHub Actions work, this will probably fail on first try. 😅


DanielMSchmidt (approved):
Let’s give it a spin

Issue #2892, opened 2023-05-28
Allow skipping bugged resources, or whitelist used resources for the imported provider.


yuriy-yarosh:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

It would save too much time and effort if CDKTF adopters could skip unused resources imports for the existing providers, with a respective resource whitelist in the `cdktf.json` config file.

```json
{
  "language": "golang",
  "app": "echo CDKTF breaks too often and eats too much RAM, JSII should've been written in Rust",
  "terraformProviders": [{
    "name": "aws" ,
    "version": "~> 5.0.1",
    "include": [],
    "exclude": ["aws_quicksight_template", "aws_appsync_domain_name"]
  }]
}
```

```ts
export interface TerraformDependencyConstraint {
  readonly name: string; // name of the module / provider
  readonly source?: string; // path / url / registry identifier for the module / provider
  readonly version?: string; // version constraint
  readonly include?: string[]; // get and import only the specified resources, import everything if empty or unspecified
  readonly exclude?: string[]; // ignore any buggy or too complex resources, that are breaking builds, or take too much time
type RequirementDefinition = string | TerraformDependencyConstraint;
```

### References

It's pointless and cumbersome to regenerate and re-import everything from scratch, over and over, without any incremental builds (as per [#2116](https://github.com/hashicorp/terraform-cdk/issues/2116) ).

Existing generated modules are bloated with JSII bundles, about 1.5Gb (as per [#2829](https://github.com/hashicorp/terraform-cdk/issues/2829) ).

And existing providers are conflicting from time to time with JSII naming conventions (as per AWS Provider [#31615](https://github.com/hashicorp/terraform-provider-aws/issues/31615).

yuriy-yarosh:
![right now](https://github.com/hashicorp/terraform-cdk/assets/6811933/96d8c3d5-9dfe-49c0-ba44-96619f43f084)

It's literally taking around 90 min right now, OOM's on 4 threads Archlinux box with 32Gb of RAM.

PR #2932, opened 2023-06-12
feat(provider-generator): allow removing big, cost intensive structures

DanielMSchmidt:
## Summary

This change allows us to keep a disallow list of provider attributes that generate an insane amount of classes (in this case over 25.000) in our bindings. I decided to not automatically detect these instances to not prematurely optimise this. For a similar reasoning and since it's not automatically detected I also skipped the implementation of attribute skipping (since it's also very uncertain that attributes will ever be the problem).

### DIFF
--- a/packages/@cdktf/provider-generator/lib/get/generator/models/scope.ts
+++ b/packages/@cdktf/provider-generator/lib/get/generator/models/scope.ts
@@ -65,8 +65,8 @@ export class Scope {
       if (i === 0) {
         return p;
       }
-      const last = path[i - 1];
-      return p.replace(`${last}_`, "");
+      const prev = path[i - 1];
+      return p.replace(`${prev}_`, "");
     });
   }
 }
--- a/packages/@cdktf/provider-generator/lib/get/generator/models/scope.ts
+++ b/packages/@cdktf/provider-generator/lib/get/generator/models/scope.ts
@@ -13,6 +13,7 @@ export interface ScopeProps {
 
 export class Scope {
   public readonly name: string;
+  public readonly provider?: string;
   public readonly parent?: Scope;
   public readonly isProvider: boolean;
   public isComputed: boolean;
@@ -33,13 +34,39 @@ export class Scope {
   }
 
   public fullName(attributeName: string) {
+    return `${this.baseName}.${attributeName}`;
+  }
+
+  public get baseName() {
     // eslint-disable-next-line @typescript-eslint/no-this-alias
     let scope: Scope | undefined = this;
-    let name = attributeName;
+    let name = "";
     while (scope) {
-      name = `${scope.name}.${name}`;
+      if (name === "") {
+        name = scope.name;
+      } else {
+        name = `${scope.name}.${name}`;
+      }
       scope = scope.parent;
     }
     return name;
+  }
+
+  public get attributePath() {
+    // eslint-disable-next-line @typescript-eslint/no-this-alias
+    let scope: Scope | undefined = this;
+    const path: string[] = [];
+    while (scope) {
+      path.unshift(scope.name);
+      scope = scope.parent;
+    }
+
+    return path.map((p, i) => {
+      if (i === 0) {
+        return p;
+      }
+      const last = path[i - 1];
+      return p.replace(`${last}_`, "");
+    });
   }
 }
--- a/packages/@cdktf/provider-generator/lib/get/generator/models/scope.ts
+++ b/packages/@cdktf/provider-generator/lib/get/generator/models/scope.ts
@@ -51,22 +51,4 @@ export class Scope {
     }
     return name;
   }
-
-  public get attributePath() {
-    // eslint-disable-next-line @typescript-eslint/no-this-alias
-    let scope: Scope | undefined = this;
-    const path: string[] = [];
-    while (scope) {
-      path.unshift(scope.name);
-      scope = scope.parent;
-    }
-
-    return path.map((p, i) => {
-      if (i === 0) {
-        return p;
-      }
-      const prev = path[i - 1];
-      return p.replace(`${prev}_`, "");
-    });
-  }
 }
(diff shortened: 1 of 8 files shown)


team-tf-cdk:

DanielMSchmidt:
At least for go and `cdktf get` the performance didn't improve locally

mutahhir:
> At least for go and `cdktf get` the performance didn't improve locally

The go conversion is interesting. When I was running it, it seemed to take a lot longer, but it wasn't taxing my system where I could tell that it's really computationally expensive. I think we should take a look at the JSII aspects of the Go conversion too when we get to performance.

packages/@cdktf/provider-generator/lib/get/__tests__/generator/fixtures/aws_quicksight_template.test.fixture.json
ansgarm, line 1: That is a laaaarge file 😂
(Didn't read it 😄)

packages/@cdktf/provider-generator/lib/get/generator/models/scope.ts
```diff
@@ -33,13 +34,39 @@ export class Scope {
...
+      path.unshift(scope.name);
+      scope = scope.parent;
+    }
+
+    return path.map((p, i) => {
+      if (i === 0) {
+        return p;
+      }
+      const last = path[i - 1];
+      return p.replace(`${last}_`, "");
```
ansgarm, line 69: I feel like previous fits better 😄

```suggestion
      const prev = path[i - 1];
      return p.replace(`${prev}_`, "");
```

ansgarm (approved):

packages/@cdktf/provider-generator/lib/get/generator/models/scope.ts
```diff
@@ -33,13 +34,39 @@ export class Scope {
...
+        name = scope.name;
+      } else {
+        name = `${scope.name}.${name}`;
+      }
       scope = scope.parent;
     }
     return name;
   }
+
+  public get attributePath() {
```
jsteinich, line 55: This seems to be used in place of `fullName`; is there a particular reason why that wasn't used?

packages/@cdktf/provider-generator/lib/get/generator/skipped-attributes.ts
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) HashiCorp, Inc
+// SPDX-License-Identifier: MPL-2.0
+
+/**
+ * There are large attributes on little used resources that blow up the
+ * size and compile time of the generated providers by a lot.
+ * To mitigate this issue we replace their entire tree of attributes with
+ * a single any attribute. This can still be used, although less conveniently.
+ */
+const SKIPPED_ATTRIBUTES: string[] = [
```
jsteinich, line 10: Baking these in is a slippery slope. It sends a message of not supporting arbitrary resources. It also introduces breaking changes anytime the list is changed.

website/docs/cdktf/concepts/resources.mdx
```diff
@@ -917,3 +917,13 @@ new SecurityGroup(this, "security2", new SecurityGroupConfig
 ```
 
 </CodeTabs>
+
+## Special Cases
+
+### Large Resource Configurations
+
+A few individual Terraform Resources have very deeply nested schemas with a lot of attributes. This blows up the config classes and slows down the code generation for languages besides Typescript. To work around this we sometimes limit the depth of the config classes and use `any` on deeper level, some attributes we directly expose as `any` on the top level config class.
```
jsteinich, line 925: Might want an example of how to use them. Particularly in non-Typescript as working with `any` is quite tricky.

jsteinich (commented):
This could relatively easily be extended to support some of the ideas in https://github.com/hashicorp/terraform-cdk/issues/2892. Allowing users to generate their own trimmed down providers or having a temporary escape hatch for unsupported updates is definitely appealing.
However, forcing pieces to be omitted within the generator itself seems too heavy handed, especially since the benefit doesn't appear to have been proven out.

packages/@cdktf/provider-generator/lib/get/generator/skipped-attributes.ts
mutahhir: Agreed on the breaking changes on each change. That's definitely inconvenient.

However, I think it's not really a slippery slope. The singular purpose of this list is to unblock provider generation for providers that have a schema design that won't be usable in any one of our supported languages. In this case, the attributes in this list prevent us from publishing any new versions of the provider. 

I agree that we may be able to come up with a more sophisticated solution to this problem, and we should. However, we don't know what that solution looks like yet and we want to allow users to be able to use newer versions of this provider through CDKTF now. To that end, I think this PR makes a pragmatic tradeoff for this situation.

mutahhir (approved):

packages/@cdktf/provider-generator/lib/get/generator/models/scope.ts
```diff
@@ -33,13 +34,39 @@ export class Scope {
...
+  }
+
+  public get baseName() {
     // eslint-disable-next-line @typescript-eslint/no-this-alias
     let scope: Scope | undefined = this;
-    let name = attributeName;
+    let name = "";
     while (scope) {
-      name = `${scope.name}.${name}`;
+      if (name === "") {
```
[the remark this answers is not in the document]
DanielMSchmidt: To not have the baseName start with a dot

mutahhir:
> This could relatively easily be extended to support some of the ideas in #2892. Allowing users to generate their own trimmed down providers or having a temporary escape hatch for unsupported updates is definitely appealing. However, forcing pieces to be omitted within the generator itself seems too heavy handed, especially since the benefit doesn't appear to have been proven out.

I think there's a bit of miscommunication on our part. The benefit that we were talking about in the earlier comments was regarding overall improvement in provider generation times for languages like Go, which didn't really improve (unfortunately). However, the main objective for this PR is to allow the newer releases of the AWS provider to be generated. 

I think while #2892 is an interesting idea, I think it also is limited to people who are building providers locally. Prebuilt providers have a lot of benefit for users in terms of not requiring as much compute resources for locally building providers. Thus, unless certain situations require it, prebuilt providers would be the more popular choice.

website/docs/cdktf/concepts/resources.mdx
DanielMSchmidt: Yeah good point, I'll add some in a separate PR 👍

PR #2942, opened 2023-06-14
feat(lib): Add validation against accidental backend reuse

DanielMSchmidt:
Reverts the revert of https://github.com/hashicorp/terraform-cdk/pull/2592


ansgarm:
Seems like this revert is not reverting everything that will be reverted in #2944, or am I missing something? 😄
Is there maybe a way to re-open my original PR instead?

DanielMSchmidt:
Closed at @ansgarm request

PR #2946, opened 2023-06-14
feat: Specific imports for convert

### LINKED ISSUE #2647: Use specific imports for used resources when converting

mutahhir:
### Related issue

Fixes #2647 

### Description

This PR adds support for specific imports when using `cdktf convert`. This has the following benefits. Firstly the code is more idomatic, and what we recommend people to write, rather than doing a default import of entire providers. Secondly, this unlocks code generation that works well with C# which does not have star imports.

This also couples with https://github.com/hashicorp/terraform-cdk/pull/2935 that updates the imports to the correct destination path per language

### Checklist


packages/@cdktf/hcl2cdk/lib/generation.ts
```diff
@@ -904,3 +947,84 @@ export function wrapCodeInConstructor(
...
+      ", "
+    )} } from "./.gen/providers/${provider}/${namespace}"`
+  )() as t.Statement;
+};
+
+export function buildImports(importables: ImportableConstruct[]) {
+  const groupedImportables = importables.reduce((acc, importable) => {
+    const ns = importable.namespace || "";
+    // Doing some hacky ordering of the imports to make them look a bit nicer
+    const prefix =
```
DanielMSchmidt, line 982: Why not keep sorting separately? and meta information what type it is could be stored/augmented into the list. Might make this less entangled

DanielMSchmidt (approved):

packages/@cdktf/hcl2cdk/lib/generation.ts
mutahhir: Can you explain a bit more? The original reason for grouping is to get things like:

```ts
import { TerraformVariable, TerraformStack } from 'cdktf';
```

to work, since each importable is added to the array separately. The thing I don't know is whether this is _only_ required for `cdktf` or do we have common namespaces for multiple resources too. If we know for sure we don't need to group anything other than `cdktf`, then we can split those apart.

DanielMSchmidt: I’m not against combining them, I’m just saying the way you are achieving this goal is a bit convoluted and could be cleaned up by some map, reduce, sort calls. No hard feelings though

mutahhir: gotcha. Yeah, I'm sure there's definitely a cleaner way for doing this. That being said, I'd like to get this one done and try to get your imports path fixes rebased on top and see if we're all good there. I'll keep it in mind to think about a revision later

Issue #1618, opened 2022-03-10
Unknown type tuple


benfortuna:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### cdktf & Language Versions

Typescript

![image](https://user-images.githubusercontent.com/984966/157566434-2a404a35-c752-4437-9d97-2a31e6fa7615.png)

### Affected Resource(s)

* cdktf.json

![image](https://user-images.githubusercontent.com/984966/157566027-2a826c59-7b5b-4e1a-869e-bfb50e793b68.png)

* module "environment" variables.tf

![image](https://user-images.githubusercontent.com/984966/157566139-0ac7ef9f-5614-4eff-a304-aa6eea4486df.png)

### Debug Output

![image](https://user-images.githubusercontent.com/984966/157566231-efbce84f-15ae-4131-9ea9-a74be5cbcf16.png)

### Expected Behavior

When using modules we should be able to use "typle" type for input variables.

### Actual Behavior

Module generation fails when tuple type is used.

### References

- #0000

jsteinich:
Need to update https://github.com/hashicorp/terraform-cdk/blob/8263964d3535c913ddab2770d3a7ecc61f53fa1e/packages/%40cdktf/provider-generator/lib/get/generator/module-generator.ts#L133 to have support for tuples.

aayushharwani-aidash:
@jsteinich I want to solve this issue. I am new to community. Can you please help me ?

jsteinich:
@aayushharwani-aidash Happy to have your help.
My previous comment links to the section of code that needs to be modified. Need to make sure `tuple` is included in the check and that a case is created to handle it.

https://github.com/hashicorp/terraform-cdk/blob/main/packages/@cdktf/provider-generator/lib/get/__tests__/generator/module-generator.test.ts would be a good place to add some test cases for the new type. You can basically just make a regular hcl file that contains variables of different tuple types and confirm that they work with test cases.

kination:
[...]

PR #2964, opened 2023-06-24
feat(provider-generator): handle 'tuple' type for variable

kination:
### Related issue
Fixes https://github.com/hashicorp/terraform-cdk/issues/1618

### Description
Allow `tuple` for variable type

### Checklist

### DIFF
--- a/packages/@cdktf/provider-generator/lib/get/generator/module-generator.ts
+++ b/packages/@cdktf/provider-generator/lib/get/generator/module-generator.ts
@@ -172,6 +172,9 @@ function parseType(type: string) {
   if (type === "map") {
     return "{ [key: string]: string }";
   }
+  if (type === "tuple") {
+    return "tuple";
+  }
   if (type === "any") {
     return "any";
   }
@@ -185,7 +188,7 @@ function parseType(type: string) {
 }
 
 function parseComplexType(type: string): string | undefined {
-  const complex = /^(object|list|map|set)\(([\s\S]+)\)/;
+  const complex = /^(object|list|map|set|tuple)\(([\s\S]+)\)/;
   const match = complex.exec(type);
   if (!match) {
     return undefined;
@@ -199,6 +202,16 @@ function parseComplexType(type: string): string | undefined {
 
   if (kind === "list" || kind === "set") {
     return `${parseType(innerType)}[]`;
+  }
+
+  if (kind === "tuple") {
+    const innerTuple: string[] = innerType.slice(1, -1).split(",");
+    let tupleTypes: string[] = [];
+    innerTuple.forEach((el) => {
+      tupleTypes.push(parseType(el.trim()));
+    });
+
+    return `tuple[${tupleTypes}]`;
   }
 
   if (kind === "map") {
--- a/packages/@cdktf/provider-generator/lib/get/generator/module-generator.ts
+++ b/packages/@cdktf/provider-generator/lib/get/generator/module-generator.ts
@@ -173,7 +173,7 @@ function parseType(type: string) {
     return "{ [key: string]: string }";
   }
   if (type === "tuple") {
-    return "tuple";
+    return "string[]";
   }
   if (type === "any") {
     return "any";
@@ -205,13 +205,7 @@ function parseComplexType(type: string): string | undefined {
   }
 
   if (kind === "tuple") {
-    const innerTuple: string[] = innerType.slice(1, -1).split(",");
-    let tupleTypes: string[] = [];
-    innerTuple.forEach((el) => {
-      tupleTypes.push(parseType(el.trim()));
-    });
-
-    return `tuple[${tupleTypes}]`;
+    return `any[]`;
   }
 
   if (kind === "map") {
--- a/packages/@cdktf/provider-generator/lib/get/__tests__/generator/__snapshots__/module-generator.test.ts.snap
+++ b/packages/@cdktf/provider-generator/lib/get/__tests__/generator/__snapshots__/module-generator.test.ts.snap
@@ -8489,6 +8489,11 @@ export interface ModuleConfig extends TerraformModuleUserConfig {
   * true
   */
   readonly fetchCallerIdentity?: boolean;
+  /**
+   * RDS DB subnet group resource
+   * @default [["172.0.0.1", "172.0.0.2", 1],["172.0.10.1", "172.0.10.2", 2]]
+   */
+  readonly fetchSubnetIds?: any[][];
 }
 /**
 * Defines an Module based on a Terraform module
@@ -8503,12 +8508,19 @@ export class Module extends TerraformModule {
       source: './module',
     });
     this.fetchCallerIdentity = config.fetchCallerIdentity;
+    this.fetchSubnetIds = config.fetchSubnetIds;
   }
   public get fetchCallerIdentity(): boolean | undefined {
     return this.inputs['get_caller_identity'] as boolean | undefined;
   }
   public set fetchCallerIdentity(value: boolean | undefined) {
     this.inputs['get_caller_identity'] = value;
+  }
+  public get fetchSubnetIds(): any[][] | undefined {
+    return this.inputs['get_subnet_ids'] as any[][] | undefined;
+  }
+  public set fetchSubnetIds(value: any[][] | undefined) {
+    this.inputs['get_subnet_ids'] = value;
   }
   protected synthesizeAttributes() {
     return this.inputs;
--- a/packages/@cdktf/provider-generator/lib/get/__tests__/generator/__snapshots__/module-generator.test.ts.snap
+++ b/packages/@cdktf/provider-generator/lib/get/__tests__/generator/__snapshots__/module-generator.test.ts.snap
@@ -8491,7 +8491,7 @@ export interface ModuleConfig extends TerraformModuleUserConfig {
   readonly fetchCallerIdentity?: boolean;
   /**
    * RDS DB subnet group resource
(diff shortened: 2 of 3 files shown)


packages/@cdktf/provider-generator/lib/get/generator/module-generator.ts
```diff
@@ -166,6 +166,9 @@ function parseType(type: string) {
   if (type === "map") {
     return "{ [key: string]: string }";
   }
+  if (type === "tuple") {
+    return "tuple";
```
jsteinich, line 170: `string[]` would be the best match this case. Using a raw `tuple` for a type is mostly from older Terraform versions which didn't have the same typing information available.

packages/@cdktf/provider-generator/lib/get/generator/module-generator.ts
```diff
@@ -195,6 +198,16 @@ function parseComplexType(type: string): string | undefined {
...
   }
 
+  if (kind === "tuple") {
+    const innerTuple: string[] = innerType.slice(1, -1).split(",");
+    let tupleTypes: string[] = [];
+    innerTuple.forEach((el) => {
+      tupleTypes.push(parseType(el.trim()));
+    });
+
+    return `tuple[${tupleTypes}]`;
```
jsteinich, line 208: This needs to be a valid Typescript type and more specifically a type supported by JSII.
Tuples aren't currently supported in JSII. The simple option would be to type as `any[]`. A more involved option would be to generate code for the specific types of the tuple.

jsteinich (changes requested):
Thanks for taking a shot at this. Need to make a few adjustments and then you should be good to  add some test cases.

packages/@cdktf/provider-generator/lib/get/generator/module-generator.ts
kination: @jsteinich thanks for suggestion.
Than, for type `list(tuple([string, string, number]))` defined in tf config, should be changed as
```
any[string,string,number][]
```

If you can guide me the output form which should be returned from this function, it would be pretty helpful 🙏

jsteinich: That example would probably need to be `any[][]` since tuples aren't fully supported.
The output needs to be a JSII compatible Typescript type (see https://aws.github.io/jsii/user-guides/lib-author/typescript-restrictions/ for more info).

kination:
@jsteinich I've make few change, and trying to make test case in https://github.com/hashicorp/terraform-cdk/blob/main/packages/@cdktf/provider-generator/lib/get/__tests__/generator/module-generator.test.ts 

(I've put test tf file inside `fixtures/local-variable/variable.tf`)
```
test("check valid data types", () => {
  const code = new CodeMaker();
  const module = new TerraformModuleConstraint({
    name: "local_module",
    fqn: "local_module",
    source: path.resolve(__dirname, "fixtures", "local-variable"),
  });
  const target = new ConstructsMakerModuleTarget(module, Language.TYPESCRIPT);
  new ModuleGenerator(code, [target])
...
})
```

I'm trying with following https://github.com/hashicorp/terraform-cdk/blob/main/packages/%40cdktf/provider-generator/package.json#L15
but test does not goes on...
```
      at Resolver._throwModNotFoundError (node_modules/jest-resolve/build/resolver.js:427:11)
      at Object.<anonymous> (../hcl2json/wasm/bridge_wasm_exec.js:14:28)
      at ../hcl2json/lib/bridge.ts:39:5
      at init (../hcl2json/lib/bridge.ts:39:5)


 RUNS  lib/get/__tests__/generator/module-generator.test.ts
 RUNS  lib/get/__tests__/read-schema.test.ts

Test Suites: 14 failed, 14 of 16 total
Tests:       14 failed, 40 passed, 54 total
Snapshots:   61 passed, 61 total
Time:        114 s
```

Could you give suggestion how to make testing for this?

jsteinich:
> Could you give suggestion how to make testing for this?

Might just need to reinstall dependencies (`yarn install`) or deleting a package lock and node_modules directory and trying that.

You can also likely simplify your test to just use `expectModuleToMatchSnapshot`.

kination:
@jsteinich thanks.
Also, could you give advise how to initialize module generator?
I'm trying as following:
(keeping test tf file inside fixtures/local-variable/variable.tf)
```
test("check valid data types", () => {
  const code = new CodeMaker();
  const module = new TerraformModuleConstraint({
    name: "local_module",
    fqn: "local_module",
    source: path.resolve(__dirname, "fixtures", "local-variable"),
  });
  const target = new ConstructsMakerModuleTarget(module, Language.TYPESCRIPT);
  new ModuleGenerator(code, [target])
...
})
```
but it failed as:
```
...
    missing spec for local_module

      24 |
      25 |     if (!spec) {
    > 26 |       throw new Error(`missing spec for ${target.fqn}`);
         |             ^
      27 |     }
      28 |
      29 |     this.code.openFile(target.fileName);

      at ModuleGenerator.emitSubmodule (lib/get/generator/module-generator.ts:26:13)
      at new ModuleGenerator (lib/get/generator/module-generator.ts:17:12)
      at Object.<anonymous> (lib/get/__tests__/generator/module-generator.test.ts:140:3)
```

Thanks.

jsteinich:
> Also, could you give advise how to initialize module generator?

`ModuleGenerator` is not something you're likely to ever directly create. The tests have some useful functions to simply module generation. For example:
```typescript
expectModuleToMatchSnapshot("tuple type", "generator", [
  "local-variable/variable.tf",
]);
```

kination:
@jsteinich I've just added variable inside existing template.
But it shows following error:
```
...
   Snapshot name: `getX variables 1`

    - Snapshot  - 1
    + Received  + 1

    @@ -8,11 +8,11 @@
         * @default true
         */
        readonly fetchCallerIdentity?: boolean;
        /**
         * RDS DB subnet group resource
    -    * @default [["172.0.0.1", "172.0.0.2", 1],["172.0.10.1", "172.0.10.2", 2]]
    +    * @default 172.0.0.1,172.0.0.2,1,172.0.10.1,172.0.10.2,2
         */
        readonly fetchSubnetIds?: any[][];
...
```

What I've expect in comment part should be `[["172.0.0.1", "172.0.0.2", 1],["172.0.10.1", "172.0.10.2", 2]]`, but result shows `172.0.0.1,172.0.0.2,1,172.0.10.1,172.0.10.2,2`. Could you confirm this part?

jsteinich:
> What I've expect in comment part should be `[["172.0.0.1", "172.0.0.2", 1],["172.0.10.1", "172.0.10.2", 2]]`, but result shows `172.0.0.1,172.0.0.2,1,172.0.10.1,172.0.10.2,2`. Could you confirm this part?

That's certainly not what it should be, but I don't believe anything in the PR introduced that behavior. Can make another issue to keep track of that.

jsteinich:
> @jsteinich thanks for notice.
> 
> Anyway, this part makes test failure, so it couldn't pass the checks. So options I can think is...
> 
> * Fix value in snap file(https://github.com/hashicorp/terraform-cdk/pull/2964/files#diff-0dd3563bcd9ba2e65e9a8adc75c1133318018e4c80ba1a942def57aa26f113bdR8494) same as generated format, so pass the test.
> * Fix the generator logic to make result as `["172.0.0.1", "172.0.0.2", 1],["172.0.10.1", "172.0.10.2", 2]]`. But this can be out of scope...
> 
> How do you think?

I think it makes sense to just update the snapshot so that it passes. Can leave a comment on the test case about it.
@DanielMSchmidt does that sound good to you?

ansgarm:
Updating the snapshot would be fine with me if we also create an issue to improve the generator to fix how these defaults are rendered 👍
@kination @jsteinich

kination:
@ansgarm @jsteinich thanks for suggestion. Than I'll go on with this.

https://github.com/hashicorp/terraform-cdk/actions/runs/5628083685/job/15251540738?pr=2964
=> I'm not sure why this cause failure. It works well when I run it locally. Could you check?

jsteinich:
> https://github.com/hashicorp/terraform-cdk/actions/runs/5628083685/job/15251540738?pr=2964
> => I'm not sure why this cause failure. It works well when I run it locally. Could you check?

Looks like this got broken with https://github.com/hashicorp/terraform-cdk/pull/2990. You might need to pull in latest main to see the issue locally. Can probably lump it into a follow-up issue to fix.

kination:
@jsteinich could you confirm following once more?
https://github.com/hashicorp/terraform-cdk/actions/runs/5640398393/job/15276937617?pr=2964
Seems template shows same result, but checker indicates as diff ... (some encoding issue?)

jsteinich:
> @jsteinich could you confirm following once more? https://github.com/hashicorp/terraform-cdk/actions/runs/5640398393/job/15276937617?pr=2964 Seems template shows same result, but checker indicates as diff ... (some encoding issue?)

I would guess a line ending issue.

kination:
@jsteinich I've tried to change, but git cannot find diff of this.
Could you give me suggestion how to go on with this?

kination:
@jsteinich thanks for update. But seems still there are failures on check. Could you confirm whether these are related with this update?

ansgarm:
Hi @kination 👋

fyi, I just rebased your branch onto `main` as that should hopefully fix the CI errors you are seeing ✌️

kination:
@ansgarm thanks for your help!
Seems now 1 has left (`Vercel` authorization). Please let me know if I need to follow-up something for this ... 🤔

ansgarm:
Oh, yeah, missed that. That's just something we have to click for third party PRs 😅

jsteinich (approved):

PR #2973, opened 2023-06-28
chore: trigger project board update when issues modified

xiehan:
While we do have an hourly cron in place, this workflow will be even better if issue creation and updates triggers an update on the project board. We can trigger the script in the other repo with this workflow. (The `workflow_dispatch` is just a nice-to-have for testing.) I've confirmed it works at https://github.com/hashicorp/terraform-cdk-team/actions/runs/5402845986

Should be merged alongside #2970

### DIFF
--- a/.github/workflows/project-board.yml
+++ b/.github/workflows/project-board.yml
@@ -1,26 +1,32 @@
-name: Project Board
+name: Update project board on issue changes
 
 on:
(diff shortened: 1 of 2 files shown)


DanielMSchmidt (approved):
Nice, love the push model 👍

PR #2990, opened 2023-07-06
fix(provider-generator): sanitize all comments

DanielMSchmidt:
This should fix the issues we see in Java for the google beta provider


packages/@cdktf/provider-generator/lib/get/generator/sanitized-comments.ts
mutahhir, line 1: ❤️

mutahhir (approved):
Really nice abstraction. I think we could've gone a little farther and also just prefixed the `* ` parts to the `line` function, but that's just a cosmetic preference.

PR #2993, opened 2023-07-06
feat(cli): allow skipping synth

DanielMSchmidt:


packages/@cdktf/cli-core/src/lib/cdktf-project.ts
```diff
@@ -597,8 +624,10 @@ export class CdktfProject {
     });
   }
 
-  public async fetchOutputs(opts: MultipleStackOptions) {
-    const stacks = await this.synth();
+  public async fetchOutputs(opts: FetchOutputOptions) {
+    const stacks = opts.skipSynth
+      ? await this.readSynthedStacks()
```
mutahhir, line 629: Nit: I think we should be consistent between `synthed` and `synthesized`.

packages/@cdktf/cli-core/src/lib/synth-stack.ts
```diff
@@ -153,21 +153,50 @@ Command output on stdout:
...
+
+    return stacks;
+  }
+
+  public static async readSynthesizedStacks(
+    outdir: string
+  ): Promise<SynthesizedStack[]> {
+    const manifestPath = path.join(outdir, Manifest.fileName);
+    if (!(await fs.pathExists(manifestPath))) {
+      throw new Error(`Could not find manifest file at ${manifestPath}.`);
```
mutahhir, line 194:
```suggestion
      throw new Error(`Could not find manifest file at ${manifestPath}. In case --skip-synth was passed, please try again without the flag.`);
```

mutahhir: Unsure if this is the right place, but overall I wanted to see if there are ways to catch the fact that there is no synthed JSON available on disk, and highlight that the flag may have been passed.

mutahhir (approved):
Overall, good to go. Minor tweaks suggested.

PR #3000, opened 2023-07-11
fix: use codemaker output in java

DanielMSchmidt:
This allows java users to set the right package name by allowing them to change the org


packages/@cdktf/provider-generator/lib/get/constructs-maker.ts
```diff
@@ -527,7 +527,7 @@ export class ConstructsMaker {
     if (this.isJavaTarget) {
       opts.java = {
         outdir: ".", // generated java files aren't packaged, so just include directly in app
-        package: `imports.${target.srcMakName}`,
+        package: `${this.options.codeMakerOutput}.${target.srcMakName}`,
```
ansgarm, line 530: Maybe I'm missing something, but this seems to change the prefix from `imports` to `src/main/java/imports`
(as specified in [the template](https://github.com/hashicorp/terraform-cdk/blob/304e2507209c9657135dc7b4b7ee68030327468f/packages/cdktf-cli/templates/java/cdktf.json#L6)).

I suspect that this might not be the right value for the package 🤔

DanielMSchmidt: If I set it it worked, but I didn't test the default 🤔

ansgarm (approved):

PR #3002, opened 2023-07-12
fix(docs): python docs for aspect

gertzakis:
### Related issue

Fixes # 

### Description

I noticed that the documentation for Aspects in Python is not so easily understood. The reason is that it should call the Aspect class with the stack instance. 

### Checklist

### DIFF
--- a/examples/python/documentation/aspects.py
+++ b/examples/python/documentation/aspects.py
@@ -10,8 +10,10 @@ from imports.random.provider import RandomProvider
 from imports.random.pet import Pet
 
 # DOCS_BLOCK_START:define-aspects
-from constructs import IConstruct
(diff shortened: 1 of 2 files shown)


ansgarm:
Hi @gertzakis 👋

Thank you for the PR 👍 Unfortunately, the source of truth for these snippets is in [this file](https://github.com/hashicorp/terraform-cdk/blob/main/examples/python/documentation/aspects.py) and your changes in this PR would be overridden the next time we run `yarn generate-docs:translations`. We run these examples in CI to ensure the example code works.

You can add your changes over there and use the `DOCS_BLOCK_START` and `_END` markers to include the existing stack where required 😇

ansgarm (approved):
Thank you @gertzakis 💯

PR #3003, opened 2023-07-12
fix(examples): Properly set the Aspect on the stack instead of on the S3Bucket only

ansgarm:
This also removes the need for another Aspect instance to test that it does work with the pet resource, as adding it to the stack will also invoke it for the pet resource

This change helps users that read the source of this example as adding an Aspect to a single resource serves not much purpose


DanielMSchmidt (approved):

ansgarm:
(merged even though the Java Google example failed – but it fails everywhere currently)

PR #3004, opened 2023-07-12
fix: remove old lerna option causing it to fail when building a matrix in CI runs

ansgarm:
### Related issue

Fixes integration examples not running in CI currently, example: https://github.com/hashicorp/terraform-cdk/actions/runs/5529364286/jobs/10087325820?pr=3003#step:3:9

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -13,6 +13,5 @@
     "tools/*"
   ],
   "npmClient": "yarn",
-  "useWorkspaces": true,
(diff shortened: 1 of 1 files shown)


DanielMSchmidt (approved):

ansgarm:
Worked 🎉 https://github.com/hashicorp/terraform-cdk/actions/runs/5530092433/jobs/10089035541?pr=3004#step:3:13

PR #3029, opened 2023-07-24
fix: don't cross package boundaries for templates

DanielMSchmidt:
and copy the templates into the expected place in the bundle


packages/@cdktf/cli-core/src/lib/init.ts
```diff
@@ -46,6 +46,16 @@ export type InitArgs = {
...
 
+export const templatesDir = path.join(__dirname, "..", "..", "templates");
+
+const availableTemplates = fs
+  .readdirSync(templatesDir)
+  .filter((x) => !x.startsWith("."));
+export const templates: string[] = [];
+for (const template of availableTemplates) {
+  templates.push(template);
+}
```
ansgarm, line 57: Is there a specific reason to writing it like this? instead of assigning `availableTemplates`?

ansgarm (approved):

packages/@cdktf/cli-core/src/lib/init.ts
DanielMSchmidt: No idea, I just move the code. Don’t experiment in this pr 😅

ansgarm: It was my code originally 😂😂😂 Case closed, I'm the idiot 🤝

PR #3030, opened 2023-07-24
feat: remove root-level import of all modules in python

DanielMSchmidt:
In python importing a submodule leads to all `__init__.py` files below the import path being executed. We have an `index.ts` with a namespaced import for all submodules as an entry point for JSII (It needs to be there so all the submodules are being registered). This leads to a from a provider point of view top-level `__init__.py` importing all the namespaces, so even if users use specific submodule imports they import the entire provider.

This change cuts the synth time in half for the AWS example (from 30s to 15s).

This is a breaking change since top-level imports (`from imports.aws import s3Bucket`) don't work anymore, they need to be replaced with namespace imports (`from imports.aws.s3_bucket import S3Bucket`).

Additionally this PR upgrades JSII and JSII-Srcmak, allowing us to use the new lazy loading mechanism within JSII bringing the initial synth time down from 15s to 5s and every iterative synth (after the providers and modules are in the JSII cache) down to 2s.

**TL;DR: Breaking change bringing Pythons synth time down from 30s to 2s**


packages/@cdktf/provider-generator/lib/get/constructs-maker.ts
```diff
@@ -20,17 +20,29 @@ import {
...
 import { ModuleGenerator } from "./generator/module-generator";
 import { ModuleSchema } from "./generator/module-schema";
+import { glob } from "glob";
 
 export async function generateJsiiLanguage(
   code: CodeMaker,
   opts: srcmak.Options,
-  outputPath: string
+  outputPath: string,
+  disalllowedFileGlobs: string[] = []
```
ansgarm, line 29:
```suggestion
  disallowedFileGlobs: string[] = []
```

packages/@cdktf/provider-generator/lib/get/constructs-maker.ts
```diff
@@ -571,7 +591,9 @@ a NODE_OPTIONS variable, we won't override it. Hence, the provider generation mi
     }
 
     const jsiiTimer = logTimespan("JSII");
-    await generateJsiiLanguage(this.code, opts, this.codeMakerOutdir);
+    await generateJsiiLanguage(this.code, opts, this.codeMakerOutdir, [
+      target.isModule ? "providers/**" : "modules/**",
```
ansgarm, line 595: Is this another optimization not mentioned in the PR description? Or am I missing something?

ansgarm (commented):
Looking good! Only a small typo and a question.

As this is a breaking change, I think we should add an unreleased section to the changelog and give some before & after examples for this change. Upon the next release, we should also copy that to an upgrade guide (not sure if adding an upgrade guide already would release it too soon 😄)

packages/@cdktf/provider-generator/lib/get/constructs-maker.ts
DanielMSchmidt: Yeah, we have been shipping the JS code for all providers and modules in the tar.gz files for each module and provider. After this change the size of the vpc modules tar.gz file went from 12MB to 100kb and this affects the load time as well. 

This is a very simple way to achieve some benefit, unfortunately there is some hard coupling in place to get to know the path where the provider ends up and I didn’t want to start a bigger refactoring here

ansgarm (approved):

PR #3031, opened 2023-07-25
fix: use local NPM registry for running integration tests on unix systems

ansgarm:
This ensures we install the cdktf-cli in the same way it is installed on user machines

### DIFF
--- a/test/local-registry.sh
+++ b/test/local-registry.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+
+# Source: https://github.com/facebook/create-react-app/blob/0a827f69ab0d2ee3871ba9b71350031d8a81b7ae/tasks/local-registry.sh
+
+custom_registry_url=localhost:4873
+original_npm_registry_url=`npm get registry`
+original_yarn_registry_url=`yarn config get registry`
+default_verdaccio_package=verdaccio@^5.26.1
+
+function startLocalRegistry {
+  # Start local registry
+  tmp_registry_log=`mktemp`
+  echo "Verdaccio Registry log file: $tmp_registry_log"
+  
+  storage_dir="$(dirname $1)/storage"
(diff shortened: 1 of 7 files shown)


test/test-helper.ts
```diff
@@ -71,7 +71,21 @@ export class TestDriver {
   public workingDirectory: string;
 
   constructor(public rootDir: string, addToEnv: Record<string, string> = {}) {
-    this.env = Object.assign({ CI: 1 }, process.env, addToEnv);
+    // Node.js prepends all parent node_modules/.bin directories to the PATH env var
+    // which shadows the cdktf CLI we want to test against and set in run-against-dist.sh
+    // While the code is the same, we want the CLI to be installed in a temporary directory
```
DanielMSchmidt, line 76: Isn't this outdated by having a local registry?

DanielMSchmidt (approved):

test/test-helper.ts
ansgarm: Nope, we still install in a local directory (but with `--global-style`, which doesn't work from a tar.gz hence the local registry) and need that directory to be used.

test/local-registry.sh
```diff
@@ -0,0 +1,39 @@
...
+
+  nohup npx ${VERDACCIO_PACKAGE:-$default_verdaccio_package} -c $1 &>$tmp_registry_log &
+  verdaccio_pid="$!"
+  
+  # Wait for Verdaccio to boot
+  echo "Waiting for local Registry to start"
+  grep -q 'http address' <(tail -f $tmp_registry_log)
+
+  # Set registry to local registry
+  npm set registry "http://$custom_registry_url"
```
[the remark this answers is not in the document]
ansgarm: Yes, it's global – but there's an error trap that will trigger when anything is cancelled or fails, so it's as safe as we can make it.

PR #3033, opened 2023-07-25
fix: trying to fix tsc error by updating how we invoke ts projects

ansgarm:


DanielMSchmidt (approved):

ansgarm:
Merged with the `init-from-tf` test still failing as that is caused by something else.

PR #3039, opened 2023-07-27
chore: add workflow to rollout provider documentation

DanielMSchmidt:
Latest run: https://github.com/hashicorp/terraform-cdk/actions/runs/5688936110

### DIFF
--- a/.github/workflows/cdktf-provider-docs-rollout.yml
+++ b/.github/workflows/cdktf-provider-docs-rollout.yml
@@ -10,6 +10,7 @@ jobs:
     strategy:
       matrix:
         provider:
+          # To onboard new providers, add a new entry to this matrix and make sure the team-tf-cdk github user has write permissions to the repository.
           - repo: hashicorp/terraform-provider-archive
             fqn: "hashicorp/archive"
           - repo: hashicorp/terraform-provider-cloudinit
--- a/.github/workflows/cdktf-provider-docs-rollout.yml
+++ b/.github/workflows/cdktf-provider-docs-rollout.yml
@@ -3,7 +3,7 @@ name: CDKTF Provider Docs Rollout
 on:
   workflow_dispatch: {}
   schedule:
-    - cron: 0 0 * * 1
+    - cron: 0 0 * * 1 # Mondays at midnight
 
 jobs:
   cdktfDocs:
--- a/.github/workflows/cdktf-provider-docs-rollout.yml
+++ b/.github/workflows/cdktf-provider-docs-rollout.yml
@@ -32,8 +32,7 @@ jobs:
             fqn: "hashicorp/time"
           - repo: hashicorp/terraform-provider-tls
             fqn: "hashicorp/tls"
-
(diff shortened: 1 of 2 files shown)


.github/workflows/cdktf-provider-docs-rollout.yml
```diff
@@ -0,0 +1,61 @@
...
+  schedule:
+    - cron: 0 0 * * 1
+
+jobs:
+  cdktfDocs:
+    strategy:
+      fail-fast: false
+      matrix:
+        provider:
+          # To onboard new providers, add a new entry to this matrix and make sure the team-tf-cdk github user has write permissions to the repository.
```
ansgarm, line 14:
```suggestion
          # NOTE: This only works for Terraform providers in the HashiCorp Github org
          # To onboard new providers, add a new entry to this matrix and make sure the team-tf-cdk github user has write permissions to the repository.
```

ansgarm: If this is true, I'd like to add this note (which will also serve future me 😄)

.github/workflows/cdktf-provider-docs-rollout.yml
```diff
@@ -0,0 +1,61 @@
+name: CDKTF Provider Docs Rollout
+
+on:
+  workflow_dispatch: {}
+  schedule:
+    - cron: 0 0 * * 1
```
ansgarm, line 6: I'm bad at CRON and had to Google

```suggestion
    - cron: 0 0 * * 1 # Mondays at midnight
```

.github/workflows/cdktf-provider-docs-rollout.yml
```diff
@@ -0,0 +1,61 @@
...
+          - repo: hashicorp/terraform-provider-null
+            fqn: "hashicorp/null"
+          - repo: hashicorp/terraform-provider-random
+            fqn: "hashicorp/random"
+          - repo: hashicorp/terraform-provider-time
+            fqn: "hashicorp/time"
+          - repo: hashicorp/terraform-provider-tls
+            fqn: "hashicorp/tls"
+
+    uses: hashicorp/terraform-cdk/.github/workflows/registry-docs-pr-based.yml@1b1be8ee471bb2845356e064e9f10b484f636cd9
```
ansgarm, line 36: (optional) Do we need to pin if they point to the same repo? I could see future me trying to wonder why changes in the other workflow are not reflected 😅

ansgarm (approved):

.github/workflows/cdktf-provider-docs-rollout.yml
DanielMSchmidt: It’s not, if some other Provider repo gives push rights to our users it works for them as well

.github/workflows/cdktf-provider-docs-rollout.yml
DanielMSchmidt: No idea, I can try the local syntax and see if it works for security. But probably should

PR #3047, opened 2023-08-04
feat(cli): support gradle projects in provider add and list

DanielMSchmidt:


packages/@cdktf/cli-core/src/lib/dependencies/package-manager.ts
```diff
@@ -629,6 +634,88 @@ class MavenPackageManager extends PackageManager {
...
+      buildGradle.slice(0, dependenciesStartIndex + 1) +
+      `\n    implementation '${packageName}:${packageVersion}'` +
+      buildGradle.slice(dependenciesStartIndex + 1);
+    await fs.writeFile(buildGradlePath, newBuildGradle);
+  }
+
+  public async listProviderPackages(): Promise<
+    { name: string; version: string }[]
+  > {
+    const dependencyRegex =
```
ansgarm, line 678: There's also a [Gradle depedencies task](https://docs.gradle.org/current/userguide/viewing_debugging_dependencies.html) that might be more robust to use. But I don't have a strong opinion, so we might as well go with the regex until we hit problems.

ansgarm (approved):
Nice!

packages/@cdktf/cli-core/src/lib/dependencies/package-manager.ts
DanielMSchmidt: That one has no json output, so it felt as bad 😅

PR #3048, opened 2023-08-04
feat(cli): move java template to gradle

DanielMSchmidt:
- feat(cli): move java template to gradle
- feat: support gradle in cdktf debug


packages/@cdktf/commons/src/debug.ts
```diff
@@ -50,6 +50,10 @@ export function getMavenVersion() {
   return getBinaryVersion("mvn", "--version");
 }
 
+export function getGradleVersion() {
+  return getBinaryVersion("gradle", "--version");
```
ansgarm, line 54: This might not be the version invoked via the Gradle wrapper

ansgarm (approved):

PR #3096, opened 2023-08-18
chore: update dependencies for @cdktf/commons

DanielMSchmidt:


ansgarm (approved):

packages/@cdktf/cli-core/src/lib/error-reporting.ts
```diff
@@ -57,7 +57,7 @@ export async function initializErrorReporting(
   runConsentPrompt?: () => Promise<boolean>
 ) {
   let shouldReport = shouldReportCrash();
-  const ci: string | false = ciDetect();
+  const ci: string | false = ciInfo.isCI ? ciInfo.name! : false;
```
ansgarm, line 60: from the docs:
> Some CI servers not listed here might still trigger the ci.isCI boolean to be set to true if they use certain vendor neutral environment variables. In those cases ci.name will be null and no vendor specific boolean will be set to true.

ansgarm: Otherwise we'd have to know that in our telemetry data `ci == null` means some generic CI 🤔😄

PR #3124, opened 2023-09-07
refactor: move schema reading logic into package

DanielMSchmidt:
The goal of this PR is to

- Move the schema reading logic into a separate package (in order to)
- Cache the provider schemas with an experimental environment variable (in order to)
- Make our tests run more stable and faster since they won't always need download and run terraform providers

In order to bring the last bit of goodness to the PR I decided to split our cache into multiple smaller ones so they are less often invalidated and more focussed on one topic.


packages/@cdktf/provider-generator/lib/get/constructs-maker.ts
```diff
@@ -6,21 +6,23 @@ import { CodeMaker } from "codemaker";
...
-  readProviderSchema,
-  readModuleSchema,
-} from "./generator/provider-schema";
 import { TerraformProviderGenerator } from "./generator/provider-generator";
 import { ModuleGenerator } from "./generator/module-generator";
-import { ModuleSchema } from "./generator/module-schema";
 import { glob } from "glob";
+import { readSchema } from "@cdktf/provider-schema";
+
+type Awaited<T> = T extends PromiseLike<infer U> ? U : T;
```
ansgarm, line 25: nit: I'm sure you found the same StackOverflow post, but doesn't this shadow the built-in `Awaited`?
https://stackoverflow.com/a/49889856

packages/@cdktf/provider-generator/lib/get/constructs-maker.ts
```diff
@@ -353,25 +202,41 @@ export class ConstructsMaker {
...
     new ModuleGenerator(this.code, [target]);
     endTSTimer();
   }
 
-  private async generateTypescript(target: ConstructsMakerTarget) {
+  private async generateTypescript(
+    target: ConstructsMakerTarget,
+    schemas: Awaited<ReturnType<typeof readSchema>>
+  ) {
+    // TODO: maybe accessing should go into the provider schema package
```
ansgarm, line 219: leftover todo?

packages/@cdktf/provider-generator/lib/get/constructs-maker.ts
ansgarm: This seems to be copied old code that probably predated us using TS 4.5+ 😂

packages/@cdktf/provider-schema/README.md
```diff
@@ -0,0 +1,3 @@
+# @cdktf/commons
+
+This package is meant to contain all code shared across the packages, like logging, types, configuration, etc. It is not meant for public usage and will not have a stable API.
```
ansgarm, line 3: needs an update 😉

packages/@cdktf/provider-schema/ambient.d.ts
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) HashiCorp, Inc
+// Licensed under the Mozilla Public License 2.0
+declare module '@skorfmann/ink-confirm-input';
```
ansgarm, line 3:
```suggestion
// Copyright (c) HashiCorp, Inc
// Licensed under the Mozilla Public License 2.0
```

packages/@cdktf/provider-schema/ambient.d.ts
```diff
@@ -0,0 +1,12 @@
...
+declare module '@npmcli/ci-detect';
+
+declare module 'child_process' {
+    interface SpawnOptions  {
+        signal?: AbortSignal;
+    }
+}
+
+declare module 'tunnel-agent';
```
ansgarm, line 12:
```suggestion
```

packages/@cdktf/provider-schema/ambient.d.ts
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) HashiCorp, Inc
+// Licensed under the Mozilla Public License 2.0
+declare module '@skorfmann/ink-confirm-input';
+declare module '@npmcli/ci-detect';
```
ansgarm, line 4: Unsure about this one, but we might not need it either
```suggestion
```

packages/@cdktf/provider-schema/package.json
```diff
@@ -0,0 +1,89 @@
...
+    "jest-watch": "jest --watch",
+    "package": "./package.sh",
+    "dist-clean": "rm -rf dist"
+  },
+  "main": "src/index.js",
+  "types": "src/index.d.ts",
+  "repository": {
+    "type": "git",
+    "url": "git://github.com/hashicorp/terraform-cdk.git",
+    "directory": "packages/@cdktf/commons"
```
ansgarm, line 27:
```suggestion
    "directory": "packages/@cdktf/provider-schema"
```

packages/@cdktf/provider-schema/src/cache.ts
```diff
@@ -0,0 +1,45 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import { ConstructsMakerTarget, logger } from "@cdktf/commons";
+import * as fs from "fs-extra";
+
+// We keep this very simple since the caching feature is experimental
+// we might need to do housekeeping / include terraform / cdktf version in the future
+function cacheKey(input: ConstructsMakerTarget): string {
+  return `${encodeURIComponent(input.fqn)}@${input.version}`;
```
ansgarm, line 12: I think we need to encode the version as well, as it may contain e.g. `>=5.0.0` and e.g. Windows does not allow the `>` in file names.

ansgarm: I know this is experimental, but I think it's easy to improve and might save us a headache in case we take this out of experimental later (since we probably won't test this weird case on windows)

packages/@cdktf/provider-schema/src/cache.ts
```diff
@@ -0,0 +1,45 @@
...
+}
+
+export function cachedAccess<I extends ConstructsMakerTarget, O>(
+  producer: (input: I) => Promise<O>,
+  cacheDir?: string | null
+): (input: I) => Promise<O> {
+  const cacheEnabled = Boolean(
+    cacheDir && fs.lstatSync(cacheDir).isDirectory()
```
ansgarm, line 20: I'd rather error if a dir was passed but is not a valid directory. Makes it easier to spot why caching is not happening.

packages/@cdktf/provider-schema/src/cache.ts
```diff
@@ -0,0 +1,45 @@
...
+    logger.debug(`Provider Schema Cache disabled`);
+    return (input) => {
+      return producer(input);
+    };
+  }
+
+  logger.debug(`Provider Schema Cache enabled, caching at ${cacheDir}`);
+  return async (input) => {
+    const key = cacheKey(input);
+    const cachePath = `${cacheDir}/${key}.json`;
```
ansgarm, line 33: Let's use `path.join()` so it works on Windows as well

packages/@cdktf/provider-schema/src/cache.ts
```diff
@@ -0,0 +1,45 @@
...
+import { ConstructsMakerTarget, logger } from "@cdktf/commons";
+import * as fs from "fs-extra";
+
+// We keep this very simple since the caching feature is experimental
+// we might need to do housekeeping / include terraform / cdktf version in the future
+function cacheKey(input: ConstructsMakerTarget): string {
+  return `${encodeURIComponent(input.fqn)}@${input.version}`;
+}
+
+export function cachedAccess<I extends ConstructsMakerTarget, O>(
```
ansgarm, line 15: I _really_ like the simplicity of this function 👏

ansgarm (changes requested):
Found some todos mostly, a few nits (<insert 2000 nits apple watch PR meme here>) and two small improvements that would make this work on Windows

packages/@cdktf/provider-schema/src/cache.ts
DanielMSchmidt: Yeah good point, silently failing is never good

packages/@cdktf/provider-schema/README.md
```diff
@@ -0,0 +1,3 @@
+# @cdktf/schema
```
ansgarm, line 1:
```suggestion
# @cdktf/provider-schema
```

ansgarm: 🤓

ansgarm (approved):
Only one consistency nit left in the readme 😉 But good to go from my point of view!

PR #3138, opened 2023-09-20
fix: make sure noColor option is honored

DanielMSchmidt:
### Related issue

CI run

### Description

Makes sure no process can add color to output if noColor option is set

### Checklist

### DIFF
--- a/test/csharp/no-color-command-option-test/test.ts
+++ b/test/csharp/no-color-command-option-test/test.ts
@@ -22,6 +22,8 @@ describe("no-color option for cdktf deploy, diff, destroy", () => {
         cwd: driver.workingDirectory,
       }
     );
+    // These tests are sometimes flaky, therefore we log the result here to ensure we can debug it properly
+    console.log(result.stdout);
     expect(hasAnsi(result.stdout)).toBe(false);
   });
   onPosix("contains no color formatting in cdktf diff", async () => {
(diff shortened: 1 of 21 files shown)


ansgarm (approved):

test/csharp/no-color-command-option-test/test.ts
```diff
@@ -22,13 +22,17 @@ describe("no-color option for cdktf deploy, diff, destroy", () => {
...
+    console.log(result.stdout);
     expect(hasAnsi(result.stdout)).toBe(false);
   });
   onPosix("contains no color formatting in cdktf diff", async () => {
     const result = await execa("cdktf", ["diff", "--no-color"], {
       env: driver.env,
       cwd: driver.workingDirectory,
     });
+    // These tests are sometimes flaky, therefore we log the result here to ensure we can debug it properly
+    console.log(result.stdout);
```
Maed223, line 35: Are we sure that the console.logs would preserve the color of `result.stdout`? I guess if the bash color flags are coming through It should be fine– more of a question than concern 😁.

Maed223 (approved):
Looks good 💯

Have a curiosity about the console.logs added to the tests, but its non-blocking.

test/csharp/no-color-command-option-test/test.ts
ansgarm: Yeah, they did in past runs that failed we could see the color in these logs 👍

PR #3152, opened 2023-09-28
feat(lib): add resource move functionality

### LINKED ISSUE #219: Improve Experience of Renaming / Moving Constructs
### Community Note

### Description

Given the following Construct

```ts
  const api = new AppsyncGraphqlApi(this, 'GraphQL', {
      authenticationType: 'AMAZON_COGNITO_USER_POOLS',
      name: 'cdk.dev',
      schema,
      userPoolConfig: [{
        defaultAction: 'DENY',
        userPoolId: userPool.id!
      }]
    })
```

When renaming its Construct id from `GraphQL` to `GraphQL1` the resource gets recreated by Terraform since the Terraform resource name changes from `cdkdev_GraphQL_E4BC4DB4` to `cdkdev_GraphQL1_A79E37DD`. This applies not only to renaming, but also when refactoring and moving a Construct into another Construct and therefore changing its position in the Construct tree.

<details><summary>Terraform Plan excerpt</summary><p>

```json
    {
      "address": "aws_appsync_graphql_api.cdkdev_GraphQL1_A79E37DD",
      "mode": "managed",
      "type": "aws_appsync_graphql_api",
      "name": "cdkdev_GraphQL1_A79E37DD",
      "provider_name": "aws",
      "change": {
        "actions": [
          "create"
        ],
        "before": null,
        "after": {
          "additional_authentication_provider": [],
          "authentication_type": "AMAZON_COGNITO_USER_POOLS",
          "log_config": [],
          "name": "cdk.dev",
          "openid_connect_config": [],
          "schema": "\n      schema {\n        query: Query\n      }\n\n      type Query {\n        test: Int,\n        snippet: Snippet!\n      }\n\n      type Snippet {\n        id: ID!\n        author: String!\n        title: String!\n        content: String!\n      }\n  ",
          "tags": null,
          "user_pool_config": [
            {
              "app_id_client_regex": null,
              "default_action": "DENY",
              "user_pool_id": "eu-central-1_Uieh1GWj0"
            }
          ],
          "xray_enabled": null
        },
        "after_unknown": {
          "additional_authentication_provider": [],
          "arn": true,
          "id": true,
          "log_config": [],
          "openid_connect_config": [],
          "uris": true,
          "user_pool_config": [
            {
              "aws_region": true
            }
          ]
        }
      }
    },
    {
      "address": "aws_appsync_graphql_api.cdkdev_GraphQL_E4BC4DB4",
      "mode": "managed",
      "type": "aws_appsync_graphql_api",
      "name": "cdkdev_GraphQL_E4BC4DB4",
      "provider_name": "aws",
      "change": {
        "actions": [
          "delete"
        ],
        "before": {
          "additional_authentication_provider": [],
          "arn": "arn:aws:appsync:eu-central-1:412153864843:apis/xxx",
          "authentication_type": "AMAZON_COGNITO_USER_POOLS",
          "id": "xxx",
          "log_config": [],
          "name": "cdk.dev",
          "openid_connect_config": [],
          "schema": "\n      schema {\n        query: Query\n      }\n\n      type Query {\n        test: Int,\n        snippet: Snippet!\n      }\n\n      type Snippet {\n        id: ID!\n        author: String!\n        title: String!\n        content: String!\n      }\n  ",
          "tags": {},
          "uris": {
            "GRAPHQL": "https://xxx.appsync-api.eu-central-1.amazonaws.com/graphql",
            "REALTIME": "wss://xxx.appsync-realtime-api.eu-central-1.amazonaws.com/graphql"
          },
          "user_pool_config": [
            {
              "app_id_client_regex": "",
              "aws_region": "eu-central-1",
              "default_action": "DENY",
              "user_pool_id": "eu-central-1_Uieh1GWj0"
            }
          ],
          "xray_enabled": false
        },
        "after": null,
        "after_unknown": {}
      }
    },
```
<p></details>

I think it would desirable to make this a smoother experience. There are various approaches to this, here are two which came to my mind:

- Provide a helper to rename resources easily 
- Autodetect moved resources (similar to what Git does with renamed files)

### LINKED ISSUE #1292: Resource Refactoring - Moved State
### Community Note

### Description

With Terraform v1.1 there will be a new addition targeting refactorings: [moved blocks](https://github.com/hashicorp/terraform/blob/14a8668486007d9007a9b8ae851d5aa987087c7f/website/docs/language/modules/develop/refactoring.html.md)  

```hcl
moved {
  from = aws_instance.a
  to   = aws_instance.b
}
```

This allows moving the state of resources via configuration. Which means, stateful refactorings can become a part of the development cycle - including testing, reviews and multi stage deployments.

That's generally quite interesting for us, since we're generating our resource names dynamically. For complex stacks with multiple levels of constructs, it will get challenging to identify a resource which gets moved around just by looking at the code. That's where the new `move` block could help us to provide a better UX to deal with this.

Once Terraform v1.1. gets released, the `moved` block can be used immediately via [escape hatches](https://www.terraform.io/docs/cdktf/concepts/providers-and-resources.html#escape-hatch). How exactly `moved` might look like as a full feature on our end is yet to be determined. 

### References

- https://github.com/hashicorp/terraform-cdk/issues/1102
- https://github.com/hashicorp/terraform-cdk/issues/1091
- https://github.com/hashicorp/terraform-cdk/issues/219
- https://github.com/hashicorp/terraform-cdk/issues/1739

Maed223:
### Related issue

Fixes #1292
Fixes #219

## Description

PR adds support for `moved` block functionality– though only in function not form.

Instead of directly porting over the form of Terraform’s moved block into CDKTF, use case specific functions that leverage the functionality of the moved block have been created. Additionally, instead of using a direct reference, or directly specifying the resource Ids to move from/to, a resource targeting system is used to more simply mark and later identify the location of the move.

#### `addMoveTarget(moveTarget: string)`: 

Adds a string moveTarget to the resource it is called upon.
```ts
new S3Bucket(this, "test-bucket", {
  bucket: "test-move-bucket-name",
}).addMoveTarget("move-s3");
```

This `moveTarget` can then be accessed in any place within the stack, including nested Constructs.

#### `moveTo(moveTarget: string, index?: string | number)`: 

Performs the move operation from the resource it is being called upon to the moveTarget specified

```ts
new S3Bucket(this, "test-bucket-rename", {
      bucket: "test-move-bucket-name",
}).moveTo("move-s3")
```

### In Typescript

Screenshot 2023-10-04 at 3 14 40 PM

### In Python

Screenshot 2023-10-11 at 12 30 57 PM

### Other Notes

I've forgone including integration tests as I believe unit testing should be sufficient (see additions to `resource.test.ts`). Seems to me that any issue not caught by them would be a problem on the side of Terraform.


packages/cdktf/lib/terraform-move-addresses.ts
```diff
@@ -0,0 +1,68 @@
...
+    // what if we keep trying to access the scope until it is undefined
+
+    // if scope.scope === undefined we are at the stack level
+
+    let currAbstractionLevel: IConstruct | undefined = scope;
+    let oneHigher: IConstruct | undefined = currAbstractionLevel.node.scope;
+    while (!App.isApp(oneHigher)) {
+      currAbstractionLevel = currAbstractionLevel?.node.scope;
+      oneHigher = currAbstractionLevel?.node.scope;
+    }
```
ansgarm, line 48: There is also `App.of(scope)` to just get the app 😉

ansgarm: That said, `TerraformStack.of(scope)` might be what could simplify this entire method even more 😉

packages/cdktf/lib/terraform-move-addresses.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  // what do we need?
+  /**
+   * map for a tag -> resource path !!!
+   * function to add to the resource path at a tag
+   * function to return the specificed resource address at a certain tag
+   */
+
+  public add(resource: TerraformResource, tag: string): void {
+    this._terraformAddressMap.set(tag, resource);
```
ansgarm, line 28: I think we should throw an error if a tag is already set 🤔

Maed223: Included 👍

packages/cdktf/lib/terraform-move-addresses.ts
Maed223: Bless you for informing me of this 😌

examples/typescript/aws-move/main.ts
```diff
@@ -0,0 +1,283 @@
...
+export class UnNestingMoveStack extends TerraformStack {
+  constructor(scope: Construct, id: string) {
+    super(scope, id);
+
+    new AwsProvider(this, "aws", {
+      region: "us-west-2",
+    });
+    new S3Bucket(this, "test-bucket-1", {
+      bucket: "test-move-bucket-name-1",
+    }).addMoveTarget("move");
```
ansgarm, line 55: I really like the new function name 👍

examples/typescript/aws-move/main.ts
```diff
@@ -0,0 +1,283 @@
...
+
+import { Construct } from "constructs";
+//import { App, TerraformStack } from "cdktf";
+import { App, TerraformStack, TerraformIterator } from "cdktf";
+import { AwsProvider } from "./.gen/providers/aws/provider";
+import { S3Bucket } from "./.gen/providers/aws/s3-bucket";
+
+// UN-NESTING RESOURCE -> resource currently in nested construct, want to move it to higher layer
+
+// FIRST -> deploy, then comment out
```
ansgarm, line 14: (just a note, no need to change anything)
An alternative to commenting out large chunks of code, could be to use an environment variable as a flag in order to change the behavior of the code. That way it's also programmatically accessible to tests that run synth + plan + apply in multiple phases.

packages/cdktf/lib/terraform-resource-targets.ts
```diff
@@ -0,0 +1,47 @@
...
+        }`
+      );
+    }
+    this._resourceTargetMap.set(target, resource);
+  }
+
+  public getResourceAddressByTarget(target: string): TerraformResource {
+    const result = this._resourceTargetMap.get(target);
+    if (!result) {
+      throw new Error(`Tag "${target}" has not been set:
```
ansgarm, line 38:
```suggestion
      throw new Error(`Target "${target}" has not been set:
```

packages/cdktf/lib/terraform-resource.ts
```diff
@@ -231,4 +252,64 @@ export class TerraformResource
...
+   * @param index Optional The index corresponding to the key the resource is to appear in the foreach of a resource to move to
+   */
+  public moveTo(moveTarget: string, index?: string | number) {
+    const stackMoveTargets = this.parentStackResourceTargets();
+    const resourceToMoveTo =
+      stackMoveTargets.getResourceAddressByTarget(moveTarget);
+    if (this.terraformResourceType !== resourceToMoveTo.terraformResourceType) {
+      throw new Error(
+        `The move target "${moveTarget}" corresponding to the resource of type ${resourceToMoveTo.terraformResourceType} to move to differs from the resource of type ${this.terraformResourceType} being moved from`
+      );
```
ansgarm, line 301: I'd preface with a line like "You seem to move a resource to a different type" or something like that.

That said, I suspect there might be cases where this would actually work (i.e. if there are two resource types that target the same cloud vendor resource). However, I'd ignore that for now and call that out of scope.

packages/cdktf/lib/terraform-resource.ts
```diff
@@ -231,4 +252,64 @@ export class TerraformResource
...
+        : `${this.terraformResourceType}.${resourceToMoveTo.friendlyUniqueId}[${index}]`
+      : `${this.terraformResourceType}.${resourceToMoveTo.friendlyUniqueId}`;
+    const movedFromId = `${this.terraformResourceType}.${this.friendlyUniqueId}`;
+    this._moved = { to: movedToId, from: movedFromId, renamed: false };
+    // TODO: add validation of correct Terraform Version
+  }
+
+  /**
+   * Adds a string moveTarget to this resource
+   * @param moveTarget The string move target that corresponds to the address of this resource
```
ansgarm, line 320: I'd adjust the comment a little to describe what this is good for. For example we could explain that this string is a user defined key that should be unique and will later be referenced in a `.moveTo()` call. I'd also give a mention of that other method this method is supposed to be used together with.

ansgarm: same for the other comment of `moveTo`. It could mention `addMoveTarget` as the method via which the moveTarget string was previously defined.

packages/cdktf/lib/terraform-resource.ts
```diff
@@ -231,4 +252,64 @@ export class TerraformResource
...
+
+  /**
+   * Moves this resource to the location denoted by moveTarget.
+   * @param moveTarget The previously set string moveTarget denoting the location of the move
+   * @param index Optional The index corresponding to the key the resource is to appear in the foreach of a resource to move to
+   */
+  public moveTo(moveTarget: string, index?: string | number) {
+    const stackMoveTargets = this.parentStackResourceTargets();
+    const resourceToMoveTo =
+      stackMoveTargets.getResourceAddressByTarget(moveTarget);
```
ansgarm, line 297: Have you tested if there are cases where `moveTo` is evaluated before `addMoveTarget`? E.g. by changing the order of calls to the resource and to the nested construct in the main stack? I suspect we might need to decouple the evaluation here (I can prototype something, might need an aspect to delay validation until all constructs are added).

packages/cdktf/test/resource.test.ts
```diff
@@ -320,3 +327,193 @@ it("maintains the same order of provisioner", () => {
...
+  const nestedContruct = new Construct(construct, "nested-construct");
+
+  new TestResource(nestedContruct, "simple", {
+    name: "foo",
+    provisioners: [
+      { type: "local-exec", command: "echo 'hello' > world.txt" },
+      { type: "local-exec", command: "echo 'hello' > world1.txt" },
+      { type: "local-exec", command: "echo 'hello' > world2.txt" },
+    ],
+  }).moveTo("test");
```
ansgarm, line 417: We should also add a test where `moveTo` is called before `addMoveTarget` because I think (as mentioned above) that this might happen in real world applications and if we can make it work nevertheless, that would improve the user experience quite a bit.

But to be fair, this can also be done in a follow-up PR as that's something we can implement "behind the scenes" without changes to the publicly facing APIs.

website/docs/cdktf/concepts/resources.mdx
```diff
@@ -370,6 +370,84 @@ import imports.kubernetes.namespace.NamespaceMetadata;
 
 </CodeTabs>
 
+## Refactoring & Renaming Resources
+
+By default moving and/or renaming resource will be regarded with the intent to destroy the resource and re-create it. If such behavior must be avoided, but there is need to refactor your infrastructure definitions and/or rename resources in it, the function `moveTo` can serve your purpose.
```
ansgarm, line 375:
```suggestion
By default moving and/or renaming resources will be regarded with the intent to destroy the resource and re-create it. If such behavior must be avoided, but there is need to refactor your infrastructure definitions and/or rename resources in it, the function `moveTo` can serve your purpose.
```

Also: It seems like there's quite a bit of passive voice that I learned we usually try to avoid when writing docs.

packages/cdktf/lib/terraform-element.ts
```diff
@@ -76,12 +76,10 @@ export class TerraformElement extends Construct {
...
-      if (this._logicalIdOverride) {
-        this._friendlyUniqueId = this._logicalIdOverride;
-      } else {
-        this._friendlyUniqueId = this.cdktfStack.getLogicalId(this);
-      }
+    if (this._logicalIdOverride) {
+      this._friendlyUniqueId = this._logicalIdOverride;
+    } else if (!this._friendlyUniqueId) {
+      this._friendlyUniqueId = this.cdktfStack.getLogicalId(this);
     }
```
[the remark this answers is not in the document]
Maed223: To my understanding of what it should be doing, it wasn't working.

So in the new `renameResourceId` function we call `this.overrideLogicalId` to set the new id associated with the rename. Which just adds to a global on TerraformElement:

```ts
public overrideLogicalId(newLogicalId: string) {
    ok(
      !this._fqnToken,
      "Logical ID may not be overridden once .fqn has been requested. Make sure to override the id before passing the construct to other constructs."
    );
    this._logicalIdOverride = newLogicalId;
  }
```

Before those changes to `friendlyUniqueId()` of TerraformElement, the new logical id set by `this.overrideLogicalId` didn't come through in calling `this.friendlyUniqueId` on the instance of the resource as `this.friendlyUniqueId` was already set. 

As `friendlyUniqueId()` was before:
```ts
public get friendlyUniqueId() {
    if (!this._friendlyUniqueId) {
      if (this._logicalIdOverride) {
        this._friendlyUniqueId = this._logicalIdOverride;
      } else {
        this._friendlyUniqueId = this.cdktfStack.getLogicalId(this);
      }
    }
```

Figured it just wasn't working as it was suppose to as I couldn't figure out how the id override could otherwise come through. And I'd imagine we'd still want to be able to override it even it is does currently have an id.

Though correct me if I'm wrong 😅

examples/typescript/aws-move/main.ts
Maed223: That's high praise coming from you 😁

packages/cdktf/lib/terraform-element.ts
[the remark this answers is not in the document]
Maed223: A safer route could be to make a separate private function that does this, might avoid adding a potential footgun.

packages/cdktf/test/resource.test.ts
Maed223: Yea there isn't a test as that currently does not work 😅. Though now that you mention aspects (which I admittedly don't have a great understanding of) I think it'd make sense to include them here to avoid that pain point. I'll give it a go!

examples/typescript/aws-move/main.ts
```diff
@@ -0,0 +1,283 @@
...
+      forEach: iterator,
+      bucket: iterator.value,
+    }).addMoveTarget("moveToResourceWithListIterator");
+
+    new S3Bucket(this, "moved-bucket-complex-iterator", {
+      bucket: "website-static-file-list-iterator-one",
+    }).moveTo(
+      "moveToResourceWithListIterator",
+      "website-static-file-list-iterator-one"
+    );
```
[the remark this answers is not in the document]
Maed223: 😊

packages/cdktf/lib/terraform-resource-targets.ts
```diff
@@ -0,0 +1,47 @@
...
+    return [...this._resourceTargetMap.entries()].reduce(
+      (accum, curr) => accum + "\n" + `${curr[0]}: ${curr[1].friendlyUniqueId}`,
+      ""
+    );
+  }
+
+  public addResourceTarget(resource: TerraformResource, target: string): void {
+    if (this._resourceTargetMap.has(target)) {
+      throw new Error(
+        `Tag "${target}" has already been set at ${
```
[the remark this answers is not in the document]
Maed223: 👍

packages/cdktf/lib/terraform-resource-targets.ts
Maed223: 👍

packages/cdktf/lib/terraform-resource.ts
```diff
@@ -220,6 +236,11 @@ export class TerraformResource
       overrides: {
         [this.terraformResourceType]: Object.keys(this.rawOverrides),
       },
+      moved: this._moved
+        ? {
+            [this.terraformResourceType]: [this.friendlyUniqueId], // TODO: anything more we need here?
```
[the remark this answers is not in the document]
Maed223: 👍

packages/cdktf/lib/terraform-resource-targets.ts
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  public getResourceAddressByTarget(target: string): TerraformResource {
+    const result = this._resourceTargetMap.get(target);
+    if (!result) {
+      throw new Error(`Tag "${target}" has not been set:
+      
+            Current Tag Entries:\n ${this.prettyPrintEntries()}
+      
+            Make sure the resource being moved to has been tagged before .movedTo() has been called
```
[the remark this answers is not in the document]
Maed223: 👍

packages/cdktf/lib/terraform-resource.ts
Maed223: 👍

packages/cdktf/lib/terraform-resource.ts
Maed223: 👍 included

website/docs/cdktf/concepts/resources.mdx
```diff
@@ -370,6 +370,78 @@ import imports.kubernetes.namespace.NamespaceMetadata;
 
 </CodeTabs>
 
+## Refactoring & Renaming Resources
+
+When working with your infrastructure definitions and the need arises to refactor or rename resources without destroying and recreating them, you can leverage the `moveTo` function. This section outlines how to perform resource moves, rename resources, and enable foreach on existing resources in your CDKTF project.
+
+### Performing Resource Moves
+
+The `moveTo` function is available on all resources and is used for relocating a resource to the location specified by the string target `moveTarget`. To set the `moveTarget` for a resource, utilize the `addMoveTarget` function that is present on the resource to move to.
```
ansgarm, line 379: Maybe add something along these lines?
```suggestion
The `moveTo` function is available on all resources and is used for relocating a resource to the location specified by the string target `moveTarget`. To set the `moveTarget` for a resource, utilize the `addMoveTarget` function that is present on the resource to move to. The `moveTarget` can be a string of your choice but has to be unique within your stack.
```

website/docs/cdktf/concepts/resources.mdx
```diff
@@ -370,6 +370,78 @@ import imports.kubernetes.namespace.NamespaceMetadata;
...
+new S3Bucket(this, "bucket-id", {
+  bucket: "bucket-rename",
+}).renameResourceId("bucket-new-id");
+```
+
+After deploying the resource, you can update your resource to use the newly assigned resource ID, and the `renameResourceId` call can be safely removed.
+
+### Enabling `foreach` on a Resource
+
+If you wish to incorporate a deployed resource into a composition using `foreach`, you can achieve this without destroying the resource by specifying the desired index using the `moveTo` function. The index should correspond to the key found in the `TerraformIterator` named iterator
```
ansgarm, line 420:
```suggestion
If you wish to incorporate a deployed resource into a composition using `foreach`, you can achieve this without destroying the resource by specifying the desired index as a second argument for the `moveTo` function. The index should correspond to the key found in the `TerraformIterator` named `iterator`
```

ansgarm: Not to happy with my "for the moveTo function" wording, maybe you have some better idea to include that there?
(I'd like to make clear that the index is not an alternative to the moveTarget but rather an additional param)

ansgarm (commented):
Just read the docs. Good work 💪
Got two small additions and a few re-wordings to clarify things.

website/docs/cdktf/concepts/resources.mdx
trujillo-adam, line 379:
```suggestion
The `moveTo` function is available on all resources and is used for relocating a resource to the location specified by the string target `moveTarget`. To set a resource's `moveTarget`, use the `addMoveTarget` function that is present on the resource to move to. You can specify an arbitrary string for the `moveTarget`, but it must be unique within your stack.
```
I'm having trouble parsing "use the `addMoveTarget` function that is present on the resource to move to". It might be helpful to use a more concrete verb to describe what you do with the `addMoveTarget` function, for example "add the `addMoveTarget` function to the destination resource. Terraform moves the resource configured with the `moveTo` function to the target during the `apply` operation." I just made that up, of course, but it I think it would be helpful to include that level of (accurate) detail.

website/docs/cdktf/concepts/resources.mdx
```diff
@@ -370,6 +370,78 @@ import imports.kubernetes.namespace.NamespaceMetadata;
...
+new S3Bucket(this, "test-bucket-move-from", {
+  bucket: "move-bucket-name",
+}).moveTo("move-s3");
+```
+
+After deployment, both the resource being moved and the move target on the destination resource can be removed.
+
+#### Move Targets
+
+A `moveTarget` is accessible anywhere within the context of the stack where it is created. Whether it's set at the root of the stack or within a nested construct, you can access it accordingly.
```
trujillo-adam, line 395:
```suggestion
A `moveTarget` is accessible anywhere within the context of the stack where it is created, including the root of the stack and within a nested construct. Moving resources to a different stack is not supported. 
```
Is that more concise? Per the style guide, temporal language, such as "in a future version . . .", "at this time", etc. should be omitted.

website/docs/cdktf/concepts/resources.mdx
trujillo-adam, line 420:
```suggestion
To incorporate a deployed resource into a composition without destroying the resource, use the `foreach` function and specify an index for the `moveTo` function. The index should correspond to the key in the `TerraformIterator` named `iterator`. 
```
Is that clearer?

trujillo-adam (commented):
Left a few suggestions for the documentation.

website/docs/cdktf/concepts/resources.mdx
Maed223: Huge improvement on that wording, [included 👍].(https://github.com/hashicorp/terraform-cdk/pull/3152/commits/70e4146b79b73f51e3b568df7fd5918366cb5b49)

website/docs/cdktf/concepts/resources.mdx
Maed223: Will note to not use such language in the future, [included](https://github.com/hashicorp/terraform-cdk/pull/3152/commits/70e4146b79b73f51e3b568df7fd5918366cb5b49) 👍

website/docs/cdktf/concepts/resources.mdx
Maed223: Far clearer, changed your suggestion to use some more correct wording but otherwise [included](https://github.com/hashicorp/terraform-cdk/pull/3152/commits/70e4146b79b73f51e3b568df7fd5918366cb5b49) 👍

Maed223: included 👍

website/docs/cdktf/concepts/resources.mdx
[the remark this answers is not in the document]
Maed223: included 👍

website/docs/cdktf/concepts/resources.mdx
Maed223: included 👍

packages/cdktf/lib/terraform-element.ts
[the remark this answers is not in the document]
Maed223: Descoped `renameResourceId`, related changes to id behavior has been reverted.

packages/cdktf/lib/terraform-resource.ts
```diff
@@ -342,6 +323,14 @@ export class TerraformResource
...
    */
   public moveTo(moveTarget: string, index?: string | number) {
+    if (this._moved) {
+      throw new Error(
+        `The resource ${this.friendlyUniqueId} has been given two moveTargets: "${this._moved.moveTarget}" and "${moveTarget}"
+
+        A resource can only be moved once per plan/apply
+        `
+      );
+    }
```
ansgarm, line 333: Nice 👏

packages/cdktf/test/resource.test.ts
```diff
@@ -585,6 +558,67 @@ it("moves multiple resources", () => {
...
+});
 
-  expect(Testing.synth(stack)).toMatchSnapshot();
+it("moves correctly when target set after call to moveTo", () => {
+  const app = Testing.app();
+  const stack = new TerraformStack(app, "test");
+  new TestProvider(stack, "provider", {});
+
+  const construct = new Construct(stack, "construct");
+  const nestedContruct = new Construct(construct, "nested-construct");
```
ansgarm, line 592: Nice way of using a nested construct in a test 💯

ansgarm (approved):
Nice work, @Maed223 👋
This can be merged from my point of view 💯

One thing we should still do (don't know if you've done that already), is to test this manually in a JSII enabled language. I don't expect anything to not work there, but I'd like to be safe 😄

packages/cdktf/lib/terraform-resource-targets.ts
```diff
@@ -0,0 +1,47 @@
...
+    }
+    this._resourceTargetMap.set(target, resource);
+  }
+
+  public getResourceByTarget(target: string): TerraformResource {
+    const result = this._resourceTargetMap.get(target);
+    if (!result) {
+      throw new Error(`Target "${target}" has not been set:
+      
+            Current Target Entries:\n ${this.prettyPrintEntries()}
```
DanielMSchmidt, line 40: I'd change this a bit to something that says these are the currently available options. Also not sure but can this run into a race condition depending on the order of constructs?

DanielMSchmidt: I think we could solve this by accessing it and throwing within a [LazyToken](https://github.com/hashicorp/terraform-cdk/blob/main/packages/cdktf/lib/tokens/lazy.ts#L110).

packages/cdktf/lib/terraform-resource.ts
```diff
@@ -260,4 +282,69 @@ export class TerraformResource
...
+    );
+  }
+
+  private _buildMovedBlock() {
+    if (!this._moved) {
+      return undefined;
+    }
+    const { moveTarget, index } = this._moved;
+    const resourceToMoveTo = this._getResourceTarget(moveTarget);
+    if (this.terraformResourceType !== resourceToMoveTo.terraformResourceType) {
```
DanielMSchmidt, line 303: Nice catch!

website/docs/cdktf/concepts/resources.mdx
```diff
@@ -370,6 +370,66 @@ import imports.kubernetes.namespace.NamespaceMetadata;
 
 </CodeTabs>
 
+## Refactoring & Renaming Resources
```
DanielMSchmidt, line 373: We have a guide on refactoring, we should update it using the move block. (We should also write one for import, if we have time)

DanielMSchmidt (changes requested):
Awesome work! I found a few edge cases, nothing major, but I'd like to have at least the breaking changes to be called out in the changelog / migration guide.

packages/cdktf/lib/terraform-resource-targets.ts
ansgarm: A previous version had the possible bug that you mentioned (it failed if the target was not defined before the move).
However, in the current version we deferred that to the `toTerraform` method, which is only called once the construct tree is complete.

DanielMSchmidt: Ah nice, that's great!

website/docs/cdktf/concepts/resources.mdx
Maed223: Great catch, if anything that section logically belongs in our refactoring guide.

[Moved the bulk of the section](https://github.com/hashicorp/terraform-cdk/pull/3152/commits/9a29db9c7fc3c16d444e7fc727101b56ac829af0) to the refactoring guide. I then have `resources.mdx` mention the functionality and refer to the refactoring guide for more info.

packages/@cdktf/provider-generator/lib/get/generator/models/attribute-model.ts
```diff
@@ -49,6 +49,8 @@ export function escapeAttributeName(name: string) {
   if (name === "tfResourceType") return `${name}Attribute`;
   // `importFrom` has potential for common name collision with providers
   if (name === "importFrom") return `${name}Attribute`;
+  // `move` could have common name collision with providers
+  if (name === "moveTo") return `${name}Attribute`;
```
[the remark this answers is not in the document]
Maed223: Updated the CHANGELOG and added an upgrade guide 👍

packages/cdktf/lib/terraform-resource-targets.ts
```diff
@@ -0,0 +1,47 @@
...
+
+            To add this target, call .addMoveTarget("${target}") on the instance of the resource to move to.
```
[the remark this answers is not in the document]
Maed223: Can't help but to think in TS much of the time 😅. It is a bit of an annoying problem since making it language agnostic may make it more difficult to parse (i.e. using something like `add move target resource instance function with parameter "target"`). 

Maybe we can have something that detects the language being used and translates it to the form of that language (i.e. `.addMoveTarget("target")` -> `.add_move_target("target")` in the case of Python). Though that supposes the config is accessible where the error is being thrown. Just an idea 😁

DanielMSchmidt (approved):

PR #3154, opened 2023-09-29
fix: fix Github workflow caches to restore successfully more often

ansgarm:
This PR
* Stops using random temporary directories for caches, as these make cache hits impossible (the cached paths are part of the "cache version" which needs to match in order for caches to be restored)
* Starts using the same cache directories for caches of the same type (e.g. `/usr/local/share/.cache/go` for the Go cache). This allows us to get cache hits between different jobs as a fallback
* Starts using distinct cache directories for all caches of the same type (types being `terraform`, `go`, `yarn` and `providerSchema`).
* Re-names the cache keys to allow cache hits using a hierarchical restore-keys definition
* Defines `restore-keys` for caches of the same type. This allows us to restore e.g. a yarn cache from a different job if there's no cache already for the current job (while the cache might not match 100%, it will still improve the results)
* fixes the specified Terraform version for provider integration tests (used to access a matrix variable that does not exist for these tests; probably a copy&paste error)
* removes the usage of `YARN_CACHE_FOLDER` which is not necessary, since our cache is in the yarn cache dir already
* removes `GOCACHE` for jobs that don't require it

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -44,8 +44,9 @@ jobs:
       - uses: actions/cache@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8
         with:
           path: ${{ steps.global-cache-dir-path.outputs.go }}
-          key: go-${{ runner.os }}-build
+          key: go-${{ runner.os }}-${{ hashFiles('**/go.sum') }}-build
           restore-keys: |
+            go-${{ runner.os }}-${{ hashFiles('**/go.sum') }}-
             go-${{ runner.os }}-
       - name: install dependencies
         run: yarn install
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,14 +27,26 @@ jobs:
         run: git config --global --add safe.directory /__w/terraform-cdk/terraform-cdk
       - name: ensure correct user
         run: chown -R root /__w/terraform-cdk
-      - name: Get yarn cache directory path
+      # Setup caches for yarn, and go
+      - name: Get cache directory paths
         id: global-cache-dir-path
-        run: echo "dir=$(yarn cache dir)" >> $GITHUB_OUTPUT
+        run: |
+          echo "yarn=$(yarn cache dir)" >> $GITHUB_OUTPUT
+          mkdir -p /usr/local/share/.cache/go
+          echo "go=/usr/local/share/.cache/go" >> $GITHUB_OUTPUT
       - uses: actions/cache@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8
-        id: global-cache # use this to check for `cache-hit` (`steps.global-cache.outputs.cache-hit != 'true'`)
         with:
-          path: ${{ steps.global-cache-dir-path.outputs.dir }}
-          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          path: ${{ steps.global-cache-dir-path.outputs.yarn }}
+          key: yarn-${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}-build
+          restore-keys: |
+            yarn-${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}-
+            yarn-${{ runner.os }}-
+      - uses: actions/cache@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8
+        with:
+          path: ${{ steps.global-cache-dir-path.outputs.go }}
+          key: go-${{ runner.os }}-build
+          restore-keys: |
+            go-${{ runner.os }}-
       - name: install dependencies
         run: yarn install
       - name: build and package
@@ -43,5 +55,4 @@ jobs:
           yarn package
         env:
           SENTRY_DSN: ${{ secrets.SENTRY_DSN }}
-          TF_PLUGIN_CACHE_DIR: ${{ steps.global-cache-dir-path.outputs.dir }}/terraform-plugins
-          GOCACHE: ${{ steps.global-cache-dir-path.outputs.dir }}/go-cache
+          GOCACHE: ${{ steps.global-cache-dir-path.outputs.go }}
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -27,14 +27,26 @@ jobs:
...
+          echo "yarn=$(yarn cache dir)" >> $GITHUB_OUTPUT
+          mkdir -p /usr/local/share/.cache/go
+          echo "go=/usr/local/share/.cache/go" >> $GITHUB_OUTPUT
+      - uses: actions/cache@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8
+        with:
+          path: ${{ steps.global-cache-dir-path.outputs.yarn }}
+          key: yarn-${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}-build
+          restore-keys: |
+            yarn-${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}-
+            yarn-${{ runner.os }}-
```
mutahhir, line 43: Not a huge fan of this one as it causes false negatives. If we don't have a matching yarn.lock file, let's not use a cache.

.github/workflows/build.yml
```diff
@@ -27,14 +27,26 @@ jobs:
...
       - uses: actions/cache@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8
-        id: global-cache # use this to check for `cache-hit` (`steps.global-cache.outputs.cache-hit != 'true'`)
         with:
-          path: ${{ steps.global-cache-dir-path.outputs.dir }}
-          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          path: ${{ steps.global-cache-dir-path.outputs.go }}
+          key: go-${{ runner.os }}-build
+          restore-keys: |
+            go-${{ runner.os }}-
```
mutahhir, line 49: I forgot this feedback in the last time we modified it, but we can / should also use the go.sum hash

```suggestion
            go-${{ runner.os }}-${{ hashFiles('**/go.sum') }}-
```

+ Same goes for the above `key` field

.github/workflows/examples.yml
```diff
@@ -50,16 +50,34 @@ jobs:
...
+          echo "terraform=/usr/local/share/.cache/terraform" >> $GITHUB_OUTPUT
+          mkdir -p /usr/local/share/.cache/go
+          echo "go=/usr/local/share/.cache/go" >> $GITHUB_OUTPUT
+      - uses: actions/cache@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8
+        with:
+          path: ${{ steps.global-cache-dir-path.outputs.yarn }}
+          key: yarn-${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}-examples
+          restore-keys: |
+            yarn-${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}-
+            yarn-${{ runner.os }}-
```
mutahhir, line 68: Same feedback as above.

.github/workflows/examples.yml
```diff
@@ -50,16 +50,34 @@ jobs:
...
+          key: terraform-${{ runner.os }}-${{ matrix.terraform }}-examples
+          restore-keys: |
+            terraform-${{ runner.os }}-${{ matrix.terraform }}
       - uses: actions/cache@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8
-        id: global-cache # use this to check for `cache-hit` (`steps.global-cache.outputs.cache-hit != 'true'`)
         with:
-          path: ${{ steps.global-cache-dir-path.outputs.dir }}
-          key: ${{ runner.os }}-examples-yarn-${{ hashFiles('**/yarn.lock') }}
+          path: ${{ steps.global-cache-dir-path.outputs.go }}
+          key: go-${{ runner.os }}-examples
```
mutahhir, line 78: Needs go.sum hashes too

.github/workflows/unit.yml
```diff
@@ -41,25 +41,37 @@ jobs:
...
+          restore-keys: |
+            yarn-${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}-unit-
+            yarn-${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}-
+            yarn-${{ runner.os }}-
       - uses: actions/cache@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8
         with:
           path: ${{ steps.global-cache-dir-path.outputs.terraform }}
-          key: terraform-${{ runner.os }}-${{ inputs.terraform_version }}
+          key: terraform-${{ runner.os }}-${{ inputs.terraform_version }}-unit-${{ inputs.package }}
+          # restore-keys: |
```
mutahhir, line 62: Stray comment

mutahhir: Maybe accidental comment now that I read it?

mutahhir (approved):
Approving, with comments regarding improvements

.github/workflows/build.yml
ansgarm: What do you mean by false negatives? If e.g. one node dependency was changed, all other packages in the cache should still be valid, shouldn't they? From my point of view, it's still better to have e.g. 90% of the packages in the restored cache than to start net new. Am I missing something?

.github/workflows/build.yml
ansgarm: Good catch! Makes sense!

.github/workflows/unit.yml
ansgarm: It was intended and did fix the unit tests. I should remove it entirely instead of just commenting it out though (as the fix worked).

.github/workflows/build.yml
mutahhir: Hmm, I didn't think of it that way. Yes, yarn should deal with the changes and not use the wrong dependency just because the cache has it. Nevermind then, ignore this comment.

ansgarm: One reason I could see, is a cache containing old dependencies and thus growing over time 🤔😄
But if that happens, we can remove these restore-keys :)

.github/workflows/build.yml
ansgarm: Added this to all Go caches!

.github/workflows/unit.yml
ansgarm: replaced it with a comment explaining why we don't have restore-keys for unit tests

PR #3162, opened 2023-10-02
fix: lower cache usage by reusing more of the yarn caches

ansgarm:
Github showed ~500MB per yarn cache, so let's not use one distinct cache per integration test or unit test since that definitely causes evictions at 10GB total maximum for caches 😅

image


DanielMSchmidt (approved):

mutahhir (approved):

.github/workflows/integration.yml
```diff
@@ -114,13 +114,12 @@ jobs:
           echo "terraform=/usr/local/share/.cache/terraform" >> $GITHUB_OUTPUT
           mkdir -p /usr/local/share/.cache/go
           echo "go=/usr/local/share/.cache/go" >> $GITHUB_OUTPUT
+      # only restore as an individual cache as per matrix explodes our cache usage
       - uses: actions/cache@704facf57e6136b1bc63b828d79edcd491f0ee84 # v3.3.2
```
ansgarm, line 118:
```suggestion
      - uses: actions/cache/restore@704facf57e6136b1bc63b828d79edcd491f0ee84 # v3.3.2
```

.github/workflows/integration.yml
```diff
@@ -187,13 +186,12 @@ jobs:
...
```
ansgarm, line 190:
```suggestion
      - uses: actions/cache/restore@704facf57e6136b1bc63b828d79edcd491f0ee84 # v3.3.2
```

ansgarm (commented):
❌ (Can't request changes, as it's my PR 🙄)

The rebase actually removed some intended part, as can be seen in this old commit of mine: https://github.com/hashicorp/terraform-cdk/commit/f18e9773086c07e6ad61f89c9f9b631d03288a92#diff-82454b2fc887e9985b9348b8f3bca03d2f839645a4c3ba1f3f850014d7da5c8bR118

PR #3231, opened 2023-11-02
feat(lib): add move workflows for specifying ids directly

Maed223:
## Description

Adds a alternative workflow for performing resource moves by specifying resource ids directly. 

The original workflow for resource moves didn't have (as of writing this) any bugs per say, yet some of the tradeoffs made in it's design made certain use cases quite difficult. As such, I made resource instance functions where the id to move to/from can be directly specified. Provides a better experience compared to using escape hatches, while avoiding potentially unhelpful abstraction in the targeting system.

In particular we saw in [migrating the repository manager to use our new id generation](https://github.com/cdktf/cdktf-repository-manager/pull/233) that using the target system was incredibly cumbersome. In response to this use case, I've created an Aspect that will automatically perform the resource moves between the old id generation and the new for ease of upgrading past 0.17.

#### `moveToId(id: string)`

Moves the resource it is called upon to the resource corresponding to `id`

```typescript
new S3Bucket(this, "test-bucket-moved", {
  bucket: "test-move-bucket-move-to-id",
}).moveToId("aws_s3_bucket.test-bucket-moved-to");

new S3Bucket(this, "test-bucket-moved-to", {
  bucket: "test-move-bucket-move-to-id",
});
```

#### `moveFromId(id: string)`

Move the resource corresponding to `id` to the resource it is called upon. Note that the resource being moved from must be marked as moved using the instance function `.hasMoved()`.

```typescript
new S3Bucket(this, "test-bucket-moved-from", {
  bucket: "test-move-bucket-move-from-id",
}).hasMoved();
new S3Bucket(this, "test-bucket", {
  bucket: "test-move-bucket-move-from-id",
}).moveFromId("aws_s3_bucket.test-bucket-moved-from");
```

## Id Migration Aspect

```typescript
/**
 * For migrating past 0.17 where id generation changed
 */
export class MigrateIdsAspect implements IAspect {
  visit(node: IConstruct) {
    // eslint-disable-next-line no-instanceof/no-instanceof
    if (node instanceof TerraformResource) {
      const oldId = allocateLogicalIdOldVersion(node);
      node.moveFromId(`${node.terraformResourceType}.${oldId}`);
    }
  }
}
```

Added an Aspect (`MigrateIdsAspect`) to the lib that migrates ids from prev 0.17 to post 0.17 when the feature flag was removed. Uses the old logic for Id generation to resolve the original resource address and calls `moveFromId` with that old address. 

Should solve our issue in the repository manager quite nicely as well as any users in the same situation of upgrading their CDKTF version.

### A screenshot of the new Aspect in action

Screenshot 2023-11-14 at 9 30 23 AM

### DIFF
--- a/examples/typescript/aws-move/main.ts
+++ b/examples/typescript/aws-move/main.ts
@@ -176,9 +176,59 @@ export class ComplexIteratorMoveStack extends TerraformStack {
 }
 // MOVE INTO RESOURCE USING COMPLEX ITERATOR
 
+export class MoveToIDStack extends TerraformStack {
+  constructor(scope: Construct, id: string) {
+    super(scope, id);
+
+    new AwsProvider(this, "aws", {
+      region: "us-west-2",
+    });
+    if (process.env.STEP_2) {
+      new S3Bucket(this, "test-bucket-move-by", {
+        bucket: "test-move-bucket-move-to-id",
+      }).moveToId("aws_s3_bucket.test-bucket-1");
+
+      new S3Bucket(this, "test-bucket-1", {
+        bucket: "test-move-bucket-move-to-id",
+      });
+    }
+    if (process.env.STEP_1) {
+      new S3Bucket(this, "test-bucket-move-by", {
+        bucket: "test-move-bucket-move-to-id",
+      });
+    }
+  }
+}
+
+export class MoveFromIDStack extends TerraformStack {
+  constructor(scope: Construct, id: string) {
+    super(scope, id);
+
+    new AwsProvider(this, "aws", {
+      region: "us-west-2",
+    });
+    if (process.env.STEP_2) {
+      new S3Bucket(this, "test-bucket-move-by", {
+        bucket: "test-move-bucket-move-from-id",
+      }).hasMoved();
+
+      new S3Bucket(this, "test-bucket-1", {
+        bucket: "test-move-bucket-move-from-id",
+      }).moveFromId("aws_s3_bucket.test-bucket-move-by");
+    }
+    if (process.env.STEP_1) {
+      new S3Bucket(this, "test-bucket-move-by", {
+        bucket: "test-move-bucket-move-from-id",
+      });
+    }
+  }
+}
+
 const app = new App();
-new UnNestingMoveStack(app, "un-nesting-move-stack");
-new NestingMoveStack(app, "nesting-move-stack");
-new ListIteratorMoveStack(app, "list-iterator-move-stack");
-new ComplexIteratorMoveStack(app, "complex-iterator-move-stack");
+new MoveToIDStack(app, "move-to-id-stack");
+new MoveFromIDStack(app, "move-from-id-stack");
+//new UnNestingMoveStack(app, "un-nesting-move-stack");
+//new NestingMoveStack(app, "nesting-move-stack");
+//new ListIteratorMoveStack(app, "list-iterator-move-stack");
+//new ComplexIteratorMoveStack(app, "complex-iterator-move-stack");
 app.synth();
--- a/examples/typescript/aws-move/main.ts
+++ b/examples/typescript/aws-move/main.ts
@@ -227,8 +227,8 @@ export class MoveFromIDStack extends TerraformStack {
 const app = new App();
 new MoveToIDStack(app, "move-to-id-stack");
 new MoveFromIDStack(app, "move-from-id-stack");
-//new UnNestingMoveStack(app, "un-nesting-move-stack");
-//new NestingMoveStack(app, "nesting-move-stack");
-//new ListIteratorMoveStack(app, "list-iterator-move-stack");
-//new ComplexIteratorMoveStack(app, "complex-iterator-move-stack");
+new UnNestingMoveStack(app, "un-nesting-move-stack");
+new NestingMoveStack(app, "nesting-move-stack");
+new ListIteratorMoveStack(app, "list-iterator-move-stack");
+new ComplexIteratorMoveStack(app, "complex-iterator-move-stack");
 app.synth();
--- a/examples/typescript/aws-move/main.ts
+++ b/examples/typescript/aws-move/main.ts
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: MPL-2.0
  */
 
-import { Construct, IConstruct } from "constructs";
+import { Construct } from "constructs";
 import { App, TerraformStack, TerraformIterator } from "cdktf";
 import { AwsProvider } from "./.gen/providers/aws/provider";
 import { S3Bucket } from "./.gen/providers/aws/s3-bucket";
--- a/examples/typescript/aws-move/main.ts
+++ b/examples/typescript/aws-move/main.ts
@@ -208,10 +208,6 @@ export class MoveFromIDStack extends TerraformStack {
       region: "us-west-2",
     });
     if (process.env.STEP_2) {
-      new S3Bucket(this, "test-bucket-move-by", {
-        bucket: "test-move-bucket-move-from-id",
-      }).hasMoved();
-
       new S3Bucket(this, "test-bucket-1", {
         bucket: "test-move-bucket-move-from-id",
       }).moveFromId("aws_s3_bucket.test-bucket-move-by");
--- a/examples/typescript/aws-move/main.ts
+++ b/examples/typescript/aws-move/main.ts
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: MPL-2.0
  */
 
-import { Construct } from "constructs";
+import { Construct, IConstruct } from "constructs";
 import { App, TerraformStack, TerraformIterator } from "cdktf";
 import { AwsProvider } from "./.gen/providers/aws/provider";
 import { S3Bucket } from "./.gen/providers/aws/s3-bucket";
@@ -231,4 +231,5 @@ new UnNestingMoveStack(app, "un-nesting-move-stack");
 new NestingMoveStack(app, "nesting-move-stack");
 new ListIteratorMoveStack(app, "list-iterator-move-stack");
 new ComplexIteratorMoveStack(app, "complex-iterator-move-stack");
+
 app.synth();
--- a/packages/cdktf/lib/terraform-resource.ts
+++ b/packages/cdktf/lib/terraform-resource.ts
@@ -335,7 +335,7 @@ export class TerraformResource
         to: ${this._movedById.to}
       }
 
-      Only one move operation can occur at a time. Remove one of the operations.
+      Only one move operation can occur plan/apply. Remove one of the operations.
       `);
     } else if (this._movedByTarget) {
       const movedBlockByTarget = this._buildMovedBlockByTarget(
@@ -400,7 +400,7 @@ export class TerraformResource
         to: ${this.terraformResourceType}.${this.friendlyUniqueId} (Resource calling the move to operation)
       }
 
-      Only one move operation can occur at a time. Remove one of the operations.
+      Only one move operation can occur per plan/apply. Remove one of the operations.
       `);
     }
     this._movedById = {
@@ -434,7 +434,7 @@ export class TerraformResource
         to: ${id}
       }
 
-      Only one move operation can occur at a time. Remove one of the operations.
+      Only one move operation can occur plan/apply. Remove one of the operations.
       `);
     }
     this._movedById = {
--- a/packages/cdktf/lib/terraform-resource.ts
+++ b/packages/cdktf/lib/terraform-resource.ts
@@ -303,7 +303,9 @@ export class TerraformResource
     );
   }
 
-  private _buildMovedBlockByTarget(movedTarget: TerraformResourceMoveByTarget) {
+  private _buildMovedBlockByTarget(
+    movedTarget: TerraformResourceMoveByTarget
+  ): { to: string; from: string } {
     const { moveTarget, index } = movedTarget;
     const resourceToMoveTo = this._getResourceTarget(moveTarget);
     if (this.terraformResourceType !== resourceToMoveTo.terraformResourceType) {
@@ -323,8 +325,7 @@ export class TerraformResource
     return { to, from };
   }
 
-  private _buildMovedBlock() {
-    let movedBlock: { to: string; from: string } | undefined;
+  private _buildMovedBlock(): { to: string; from: string } | undefined {
     if (this._movedByTarget && this._movedById) {
       throw new Error(`
       ${this.node.id} has been given two separate moved operations.
@@ -338,16 +339,27 @@ export class TerraformResource
       Only one move operation can occur per plan/apply. Remove one of the operations.
       `);
     } else if (this._movedByTarget) {
+      this.node.addValidation(
+        new ValidateTerraformVersion(
+          ">=1.5",
+          `Resource move functionality is only supported for Terraform >=1.5. Please upgrade your Terraform version.`
+        )
+      );
       const movedBlockByTarget = this._buildMovedBlockByTarget(
         this._movedByTarget
       );
-      movedBlock = { to: movedBlockByTarget.to, from: movedBlockByTarget.from };
+      return { to: movedBlockByTarget.to, from: movedBlockByTarget.from };
     } else if (this._movedById) {
-      movedBlock = { to: this._movedById.to, from: this._movedById.from };
+      this.node.addValidation(
+        new ValidateTerraformVersion(
+          ">=1.5",
+          `Resource move functionality is only supported for Terraform >=1.5. Please upgrade your Terraform version.`
+        )
+      );
+      return { to: this._movedById.to, from: this._movedById.from };
     } else {
-      movedBlock = undefined;
+      return undefined;
     }
-    return movedBlock;
   }
 
   /**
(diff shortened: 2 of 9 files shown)


Maed223:
Still need to shout out the new `MigrateIdsAspect` somewhere. Figure the 0.17 upgrade guide would be the logical place, but to me it doesn't seem to be the most visible place. Curious to hear opinions on this 😁

examples/typescript/aws-move/main.ts
```diff
@@ -176,9 +176,60 @@ export class ComplexIteratorMoveStack extends TerraformStack {
...
+  constructor(scope: Construct, id: string) {
+    super(scope, id);
+
+    new AwsProvider(this, "aws", {
+      region: "us-west-2",
+    });
+    if (process.env.STEP_2) {
+      new S3Bucket(this, "test-bucket-move-by", {
+        bucket: "test-move-bucket-move-from-id",
+      }).hasMoved();
```
ansgarm, line 213: curious: Would an alternative be to remove this code? Or does it has to exist until the move is done?

packages/cdktf/lib/terraform-resource.ts
```diff
@@ -317,21 +323,49 @@ export class TerraformResource
     return { to, from };
   }
 
+  private _buildMovedBlock() {
+    let movedBlock: { to: string; from: string } | undefined;
```
ansgarm, line 327: I think you could simplify this method a little by using `{ to: string; from: string } | undefined` as a return type, removing this variable and instead directly return the object wherever you current assign to this variable 🤓

packages/cdktf/lib/terraform-resource.ts
```diff
@@ -347,4 +381,84 @@ export class TerraformResource
...
+      new ValidateTerraformVersion(
+        ">=1.5",
+        `Resource move functionality is only supported for Terraform >=1.5. Please upgrade your Terraform version.`
+      )
+    );
+  }
+
+  /**
+   * Marks this resource as moved. Required to be present on a resource being moved from.
+   */
```
ansgarm, line 454: I see that the Aspect might require this, as it can't delete resources after they've been added to the construct tree.
If the user controls this, however, it would also be possible to remove the resource, right? We could hint at that other option in this doc string (if true)
```suggestion
  /**
   * Marks this resource as moved. Required to be present on a resource being moved from.
   * You can also remove the entire resource instead, if you don't need to programmatically decide whether to move it.
   * This gets set automatically if you use targets to move resources, no need to call this method in that case.
   */
```

ansgarm: (just a suggestion, feel free to put your own spin to it)

packages/cdktf/lib/terraform-resource.ts
```diff
@@ -347,4 +381,84 @@ export class TerraformResource
   public addMoveTarget(moveTarget: string) {
     this._addResourceTarget(moveTarget);
   }
+
+  /**
+   * Moves this resource to the resource corresponding to "id"
+   * @param id Full id of resource to move to
```
ansgarm, line 387:
```suggestion
   * @param id Full id of resource to move to, e.g. "aws_s3_bucket.example"
```

packages/cdktf/lib/terraform-resource.ts
```diff
@@ -317,21 +323,49 @@ export class TerraformResource
...
+      throw new Error(`
+      ${this.node.id} has been given two separate moved operations.
+
+      Move target: "${this._movedByTarget.moveTarget}"
+      Move by id: {
+        from: ${this._movedById.from}
+        to: ${this._movedById.to}
+      }
+
+      Only one move operation can occur plan/apply. Remove one of the operations.
```
ansgarm, line 338:
```suggestion
      Only one move operation can occur per plan/apply. Remove one of the operations.
```

packages/cdktf/lib/terraform-resource.ts
```diff
@@ -347,4 +381,84 @@ export class TerraformResource
...
+      new ValidateTerraformVersion(
+        ">=1.5",
+        `Resource move functionality is only supported for Terraform >=1.5. Please upgrade your Terraform version.`
+      )
+    );
+  }
+
+  /**
+   * Move the resource corresponding to "id" to this resource. Note that the resource being moved from must be marked as moved using it's instance function.
+   * @param id Full id of resource being moved from
```
ansgarm, line 421:
```suggestion
   * @param id Full id of resource being moved from, e.g. "aws_s3_bucket.example"
```

packages/cdktf/lib/upgrade-id-aspect.ts
```diff
@@ -0,0 +1,171 @@
...
+
+  // This is the previous behavior, which we want for now.
+  const stackIndex = 0;
+
+  const components = node.scopes.slice(stackIndex + 1).map((c) => c.node.id);
+  return components.length > 0 ? makeUniqueIdOldVersion(components, false) : "";
+}
+
+/**
+ * For migrating past 0.17 where id generation changed
```
ansgarm, line 161:
```suggestion
 * For migrating past 0.17 where the feature flag for the old id generation logic was removed after being deprecated since 0.15
```

packages/cdktf/test/resource.test.ts
```diff
@@ -622,3 +623,86 @@ it("moves correctly when target set after call to moveTo", () => {
     "simple"
   );
 });
+
+it.only("override logical ID - before moveTo", () => {
```
ansgarm, line 627:
```suggestion
it("override logical ID - before moveTo", () => {
```

packages/cdktf/test/resource.test.ts
```diff
@@ -622,3 +623,86 @@ it("moves correctly when target set after call to moveTo", () => {
...
+  expect(Object.keys(synthedStack.resource.test_resource)).toContain(
+    "construct_nested-construct_simple_2C3755B0"
+  );
+  // Must not include old resource being moved from
+  expect(Object.keys(synthedStack.resource.test_resource)).not.toContain(
+    "old_logical_id"
+  );
+});
+
+it.only("override logical ID - before addTarget", () => {
```
ansgarm, line 669:
```suggestion
it("override logical ID - before addTarget", () => {
```

DanielMSchmidt:
@Maed223 I think the upgrade guide is the best place for that 👍

website/docs/cdktf/examples-and-guides/refactoring.mdx
```diff
@@ -141,6 +141,48 @@ new S3Bucket(this, "moved-bucket", {
...
+new S3Bucket(this, "bucket-new", {
+  bucket: "move-bucket",
+}).moveFromId("aws_s3_bucket.bucket-moved");
+```
+
+Note that in moving from a specified resource address, the resource being moved from must be marked as moved.
+
+#### Nested Constructs
+
+Resource addresses in the context of nested constructs will not simply be the specified id given to the resource. When dealing with moving resources by their address to/from nested constructs, run `cdktf synth` and refer to `cdktf.out/stacks/'your stack name'` to find the exact address to move to/from.
```
ansgarm, line 176:
```suggestion
Resource addresses in the context of nested constructs will not simply be the specified id given to the resource. When dealing with moving resources by their address to/from nested constructs, run `cdktf synth` and refer to `cdktf.out/stacks/<stack name>/cdk.tf.json` to find the exact address to move to/from.
```

website/docs/cdktf/examples-and-guides/refactoring.mdx
```diff
@@ -141,6 +141,48 @@ new S3Bucket(this, "moved-bucket", {
...
+
+```typescript
+new S3Bucket(this, "bucket-moved", {
+  bucket: "bucket-moved",
+}).moveToId("aws_s3_bucket.nested-construct_bucket-new_5A0C9225"); // moving to S3Bucket "bucket-new" in NestedConstruct "nested-construct"
+
+new NestedConstruct(this, "nested-construct", {});
```
ansgarm, line 183:
```suggestion
new NestedConstruct(this, "nested-construct"); // contains the moved S3 bucket resource with the id "bucket-new"
```

ansgarm (commented):
Nice work! Looks good to me 🚀
Got a few cleanup items / docs improvements, but would've been an approve otherwise ✅

packages/cdktf/lib/terraform-resource.ts
```diff
@@ -103,7 +108,9 @@ export class TerraformResource
     FileProvisioner | LocalExecProvisioner | RemoteExecProvisioner
   >;
   private _imported?: TerraformResourceImport;
-  private _moved?: TerraformResourceMove;
+  private _movedByTarget?: TerraformResourceMoveByTarget;
```
DanielMSchmidt, line 111: I think one could keep the _moved attribute and combine both types of moves into it with a union.
As you see in the current solution there is now extra book keeping if sth has moved, what type of move it is etc. This is not bad pre-se but can lead to errors due to oversights when next touching this code.

```suggestion
  private _moved?: TerraformResourceMoveByTarget | TerraformResourceMoveById;
```

I would suggest refactoring this in a follow-up PR if there are no strong reasons to keep this design :)

packages/cdktf/lib/terraform-resource.ts
```diff
@@ -347,4 +381,84 @@ export class TerraformResource
...
+
+      Only one move operation can occur plan/apply. Remove one of the operations.
+      `);
+    }
+    this._movedById = {
+      to: `${this.terraformResourceType}.${this.friendlyUniqueId}`,
+      from: id,
+    };
+    this.node.addValidation(
+      new ValidateTerraformVersion(
```
DanielMSchmidt, line 445: I wonder if a single instance of this specific validation as e.g. moveBlockValidation would be possible, DRYing this up so we can keep the messaging concise :)

packages/cdktf/test/resource.test.ts
DanielMSchmidt, line 627:
```suggestion
it("override logical ID - before moveTo", () => {
```

packages/cdktf/test/resource.test.ts
DanielMSchmidt, line 669:
```suggestion
it("override logical ID - before addTarget", () => {
```

website/docs/cdktf/examples-and-guides/refactoring.mdx
```diff
@@ -141,6 +141,48 @@ new S3Bucket(this, "moved-bucket", {
...
+In instances where the move target workflow does not easily fit your use case, resource moves can be performed by directly specifying the full resource address to be moved to/from.
+
+#### Move To
+
+```typescript
+new S3Bucket(this, "bucket-moved", {
+  bucket: "move-bucket",
+}).moveToId("aws_s3_bucket.bucket-new");
+
+new S3Bucket(this, "bucket-new", {
```
DanielMSchmidt, line 155: Do I still need to specify the same bucket twice? I thought the new design allows us to "in-place" replace it. The aspect also seems to work this way without adding sth to the construct tree so either the example or aspect is wrong.

DanielMSchmidt (commented):
The only "blocking" one is my question in the refactoring examples :)

packages/cdktf/lib/terraform-resource.ts
Maed223: Good point I like that much better– included 👍

examples/typescript/aws-move/main.ts
Maed223: Would be a completely valid way to do it. My main reasoning at the time for including it was to keep some semblance of consistency between all the ways of moving resources. Or in other words, it felt odd to have users delete resource definitions in this workflow where they wouldn't in others.

Must admit that I was thinking too narrowly, as the old resource will most likely be deleted anyways. So the use of `hasMoved` is redundant. 

Going to just remove the convention 👍

packages/cdktf/lib/terraform-resource.ts
Maed223: Aspect actually doesn't require it. 

The idea being that the user upgrade their CDKTF version, which in turn would give different resource addresses for everything. In the Aspect we use the old id generation logic to then move everything in place from those old ids that are now (in light of the upgrade) no longer present. Meaning `hasMoved` isn't actually needed in anyway.

Going to remove the convention as it's redundant.

website/docs/cdktf/examples-and-guides/refactoring.mdx
Maed223: In short, moving can be done "in-place" (in a certain sense of the phrase) when moving from a resource id but not in moving to a resource id. At least if I'm understanding your point correctly.

For `moveToId` we of course need the destination resource (or in other words the place we want the resource to be after the move) to be present in the config.

For `moveFromId` I've since realized that the convention of marking the original resource with `hasMoved` to be redundant. So, yes the move can be done "in-place"

CC my response to Ansgar asking if the original resource can just be removed:
```
Would be a completely valid way to do it. My main reasoning at the time for including it was to keep some semblance of consistency between all the ways of moving resources. Or in other words, it felt odd to have users delete resource definitions in this workflow where they wouldn't in others.

Must admit that I was thinking too narrowly, as the old resource will most likely be deleted anyways. So the use of hasMoved is redundant.

Going to just remove the convention 👍
```

On another note, added both `moveToId` and `moveFromId` to just cover all potential ways a move can be perform. In general I see these workflows as something to have in between the move target workflow and using escape hatches. So hopefully in the odd case that the move target way of doing things doesn't fit a use case, there is some native way of doing it.

packages/cdktf/lib/terraform-resource.ts
Maed223: Widdled it down to just two instances of the validation– seems that taking it down to one would require some more serious refactoring, at least if the validation were to occur in the `TerraformResource` class. Going to see if the validation can instead take place on the level of the stack to avoid the potential onslaught of errors.

Maed223: Was able to put the validation as a part of the synth process.

In `synthesize` of `StackSynthesizer`:

```typescript
if (this.stack.hasResourceMove()) {
      this.stack.node.addValidation(
        new ValidateTerraformVersion(
          ">=1.5",
          `Resource move functionality is only supported for Terraform >=1.5. Please upgrade your Terraform version.`
        )
      );
    }
```

Then in `TerraformStack` we just go through the elements of the stack checking if some resource of the stack has a resource move:

```typescript
public hasResourceMove(): boolean {
    return terraformElements(this).some((e) => {
      if (TerraformResource.isTerraformResource(e) && e.hasResourceMove()) {
        return true
      }
      return false;
    })
  }
```

**Andddd it works as expected**

_in the aws-move example_: Using `TERRAFORM_BINARY_NAME=../../../../terraform_1.4.7 STEP_2=1 cdktfl synth`

we get the following failure:
Screenshot 2023-11-17 at 12 02 21 PM

Could be a point of improvement for import as well in a follow-up PR

packages/cdktf/lib/terraform-resource.ts
Maed223: This was actually my initial approach for this. Ended up going the way I did because I found that the type checking was quite annoying to work with, and that the different methods of moving required pretty different sets of logic. Though this might just be personal preference 😅. Willing to give it another look in a follow-up PR.

packages/cdktf/lib/terraform-resource.ts
DanielMSchmidt: Jut to be clear: I mean put `const moveValidation = new ValidateTerraformVersion(...)` on the top of the file and use ` this.stack.node.addValidation(moveValidation)` in the places you have it right now. But yeah nothing urgent

website/docs/cdktf/examples-and-guides/refactoring.mdx
DanielMSchmidt: > we of course need the destination resource

I beg to differ, if you know the new id you can change the toTerraform function to put your resource under this name versus the friendlyUniqueId that is currently there.

Maed223: Ahh I see what you mean now. Had a similar thought during the initial implementation with the `rename` function. Would require some rework of how `TerraformElement` overrides ids.

We come to the problem where if we override the id, we can't get the original `friendlyUniqueId`– and if we access the `friendlyUniqueId`, we then can't override the id.

Per the following code:

```typescript
private _logicalIdOverride?: string;
 private _friendlyUniqueId?: string;

public get friendlyUniqueId() {
    if (!this._friendlyUniqueId) {
      if (this._logicalIdOverride) {
        this._friendlyUniqueId = this._logicalIdOverride;
      } else {
        this._friendlyUniqueId = this.cdktfStack.getLogicalId(this);
      }
    }
    return this._friendlyUniqueId;
  }

public overrideLogicalId(newLogicalId: string) {
    ok(
      !this._fqnToken,
      "Logical ID may not be overridden once .fqn has been requested. Make sure to override the id before passing the construct to other constructs."
    );
    this._logicalIdOverride = newLogicalId;
  }
```

In either case we are only able to get half the information needed for the moved block when doing such an operation in place, as it currently stands.

packages/cdktf/lib/terraform-resource.ts
Maed223: Ah I understand now. Thought you were referring to the issue where the validation error pops up a bunch of times in the case where there are a bunch of move statements.

DanielMSchmidt (approved):

PR #3255, opened 2023-11-07
fix(cli): Don't send telemetry events if request body is larger than 8KB

### LINKED ISSUE #3198: CDKTF-CLI: Hangs on synth/plan
### Expected Behavior

1. Run cdktf synth
2. It shows a list of stacks
3. Gives me control of the terminal

### Actual Behavior

Just hangs after showing all the stacks for about 30 minutes.

### Steps to Reproduce

Haven't been able to cleanly reproduce, adding a 26th stack seems to my app will cause it, even if the 26th stack is empty, but if all stacks are empty it works.

### Versions

cdktf debug
language: go
cdktf-cli: 0.18.0
node: v20.4.0
cdktf: 0.18.0
constructs: 10.2.70
jsii: 1.88.0
terraform: 1.5.3
arch: arm64
os: darwin 23.0.0
go: go version go1.20.3 darwin/arm64

### Providers

erik@Erik-Ahrends-MacBook-Pro cloud-infrastructure % cdktf provider list
Pre-built provider information not found
Error: Pre-built provider information not found

### Gist

https://gist.github.com/eahrend/410069fa4059823813cc1ab36c1c365f

### Possible Solutions

No clue, going through the CDKTF CLI code doesn't show any references to an http 413 error, nor do I see any obvious place where it would even be making an HTTP call, let alone where the request entity would be too large during the synth process.

I tried setting sendCrashReports to false in cdktf.json as well, in the hopes that maybe there was some http call that cdktf was doing and this didn't work either.

Tried doing `export NODE_OPTIONS=--max-old-space-size=8192` in the hopes that maybe it was some memory issue, but doubling and even tripling the available memory to node didn't resolve this either.
https://github.com/hashicorp/terraform-cdk/blob/v0.18.0/packages/%40cdktf/cli-core/src/lib/synth-stack.ts#L98-L100

### Workarounds

Comment out a stack, then rerun. 

### Anything Else?

Nothing out of the ordinary.

### References

_No response_

### Help Wanted

- [ ] I'm interested in contributing a fix myself

### Community Note

ansgarm:
Resolves #3198

### DIFF
--- a/packages/@cdktf/commons/src/checkpoint.ts
+++ b/packages/@cdktf/commons/src/checkpoint.ts
@@ -15,6 +15,8 @@ const BASE_URL = `https://checkpoint-api.hashicorp.com/v1/`;
 
 const VALID_STATUS_CODES = [200, 201];
 
+const MAX_REQUEST_BODY_SIZE = 8192;
+
 function homeDir() {
   return process.env.CDKTF_HOME
     ? path.resolve(process.env.CDKTF_HOME)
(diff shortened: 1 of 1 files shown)


DanielMSchmidt (approved):
I like this as a stopgap measure, but folks with huge amounts telemetry data will see this warning all the time without being able to change it. I think we need to address the root cause and truncate the payloads, maybe track what brings the data to this size. Also as telemetry is for us I could see that an info or verbose level is more appropriate :)

ansgarm:
> will see this warning all the time

Tested this locally – the warning will only be shown if the default log level is increased 👍

> maybe track what brings the data to this size

That's a good idea for a later improvement 👍

PR #3257, opened 2023-11-07
chore: disable provider schema tests

DanielMSchmidt:


ansgarm (approved):
Can we create an issue to re-enable them and put that into the next release milestone? 😇

PR #3272, opened 2023-11-17
feat(lib): allow chaining of TerraformIterator created resources

### LINKED ISSUE #1940: Support passing a primitive value (e.g. a string) to `iterator.dynamic()`
### Community Note

### Description

https://github.com/hashicorp/terraform-cdk/pull/1882 introduced block level iterators with a design that can also be passed to properties that are lists and not blocks. Currently we allow passing an object (which enables users to create arrays of objects) only. But we could also allow e.g. strings, which would enable the following:
```ts
new ArchiveProvider.DataArchiveFile(this, "dir", {
      type: "zip",
      outputPath: "${path.module}/out1.zip",
      sourceDir: files.path,
      excludes: it.dynamic(`${it.value}.txt`), // <- this is not possible yet but would be neat
});
```

One downside to this would be the type handling through JSII, as `object | string` would appear as `interface{}` (aka `any`) in languages like e.g. Golang.

### References

* https://github.com/hashicorp/terraform-cdk/blob/4da8eaeccb62023806cfbbad0a4af868fd5d26db/test/typescript/iterators/main.ts#L34

### LINKED ISSUE #2024: Support accessing resources created by an iterator

DanielMSchmidt:
This PR brings both chaining iterators to the table as well as a mechanism to craft for expressions from iterators.

Closes #2024 
Closes #1940


packages/cdktf/lib/terraform-iterator.ts
```diff
@@ -80,6 +81,26 @@ export abstract class TerraformIterator implements ITerraformIterator {
     return new MapTerraformIterator(map);
   }
 
+  /**
+   * Creates a new iterator from a resource that
+   * has been created with the `for_each` argument.
+   */
+  public static fromResources(
```
ansgarm, line 88: This method is named in plural while the next is singular – I don't have any opinion but I'd want consistency 😄

DanielMSchmidt: Should be plural on both indicating we are talking about lists of resources. Where lists means they were created through for each

packages/cdktf/lib/terraform-iterator.ts
```diff
@@ -80,6 +81,26 @@ export abstract class TerraformIterator implements ITerraformIterator {
...
+  ): ResourceTerraformIterator {
+    return new ResourceTerraformIterator(resource);
+  }
+
+  /**
+   * Creates a new iterator from a data source that
+   * has been created with the `for_each` argument.
+   */
+  public static fromDataSources(
+    resource: ITerraformResource
```
ansgarm, line 99: Is there a specific type for DataSources? Or is the method just doubled for keeping the name simple?

packages/cdktf/lib/terraform-iterator.ts
```diff
@@ -284,3 +305,47 @@ export class MapTerraformIterator extends TerraformIterator {
...
+      throw new Error(
+        "Cannot create iterator from resource without for_each argument"
+      );
+    }
+  }
+
+  /**
+   * Returns the currenty entry in the list or set that is being iterated over.
+   * For lists this is the same as `iterator.value`. If you need the index,
+   * use count using the escape hatch:
```
ansgarm, line 331: > using the escape hatch

Don't we already have support for `TerraformCount` nowadays?

website/docs/cdktf/concepts/iterators.mdx
```diff
@@ -434,6 +434,212 @@ new TeamMembers(this, "members", {
 
 </CodeTabs>
 
+## Chaining Iterators
+
+Sometimes the need arises to loop over resources created through an iterator; commonly referrred as chaining.
```
ansgarm, line 439:
```suggestion
Sometimes the need arises to loop over resources created through an iterator; commonly referred to as chaining.
```

ansgarm (commented):
Looks good so far, only a few small things 👍

packages/cdktf/lib/terraform-iterator.ts
DanielMSchmidt: Yeah, I'll fix the comment from where I copied it over from as well 👍

packages/cdktf/lib/terraform-iterator.ts
DanielMSchmidt: It's just an alias so that we don't get people confused on if it's only supported for Resources :)

packages/cdktf/lib/terraform-iterator.ts
```diff
@@ -222,6 +251,88 @@ export abstract class TerraformIterator implements ITerraformIterator {
...
     );
   }
+
+  /**
+   * Creates a for expression that maps the iterators to its keys.
+   * For lists these would be the indices, for maps the keys.
+   * As this returns an IResolvable you might need to wrap the output in
+   * a Token, e.g. `Token.asString`.
+   */
+  public mapToKey(): IResolvable {
```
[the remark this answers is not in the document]
ansgarm: That said, `mapToValuePropertyList` would feel quite off, so I'm unsure here 😅 Still not entirely happy with the name, but not a blocker

packages/cdktf/lib/tfExpression.ts
```diff
@@ -234,8 +263,12 @@ class PropertyAccess extends TFExpression {
 }
 
 // eslint-disable-next-line jsdoc/require-jsdoc
-export function propertyAccess(target: Expression, args: Expression[]) {
-  return new PropertyAccess(target, args) as IResolvable;
+export function propertyAccess(
+  target: Expression,
+  args: Expression[],
+  forceNoBraces = false
```
ansgarm, line 269: I don't see any code setting this to true – is this change still required? It seems like you solved the problem with the `resolveExpressionPart` change?

packages/cdktf/lib/tfExpression.ts
```diff
@@ -131,6 +151,11 @@ export function rawString(str: string): IResolvable {
   return new RawString(str);
 }
 
+// eslint-disable-next-line jsdoc/require-jsdoc
+export function unescapedString(str: string): IResolvable {
+  return new UnescapedString(str);
+}
```
ansgarm, line 157: This function and the class don't seem to be called / documented or tested anywhere? Is this still required?

packages/cdktf/lib/terraform-iterator.ts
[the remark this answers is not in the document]
DanielMSchmidt: Maybe `mapToKey` could become `keys`. I'm not sure where to go with the others, I see that map is used in multiple contexts, but map is also the most common word for the kind of transformation it does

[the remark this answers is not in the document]
DanielMSchmidt: I renamed `mapToKeys` and `mapToValues` to `keys` and `values` and for `mapToValuePropertyList` I went into underscore.js which has the `pluck` method that does the same so I renamed it to that

packages/cdktf/lib/tfExpression.ts
DanielMSchmidt: Yeah that was old, removed it.

packages/cdktf/lib/terraform-iterator.ts
```diff
@@ -222,6 +251,88 @@ export abstract class TerraformIterator implements ITerraformIterator {
...
+        this._getForEachExpression(),
+        propertyAccess(FOR_EXPRESSION_VALUE, [property])
+      )
+    );
+  }
+
+  /**
+   * Creates a for expression that results in a list.
+   * This method allows you to create every possible for expression, but requires more knowledge about
+   * Terraforms for expression syntax.
```
mutahhir, line 297:
```suggestion
   * Terraform's for expression syntax.
```

packages/cdktf/lib/terraform-iterator.ts
```diff
@@ -222,6 +251,88 @@ export abstract class TerraformIterator implements ITerraformIterator {
...
+   * For the most common use cases you can use keys(), values(), and pluckProperty() instead.
+   *
+   * You may write any valid Terraform for each expression, it will result
+   * in `[ for key, val in var.myIteratorSource: <expression> ]`.
+   *
+   * As this returns an IResolvable you might need to wrap the output in
+   * a Token, e.g. `Token.asString`.
```
mutahhir, line 304: Would be nice to accompany the above terraform example with a CDKTF call showing how to use this API

mutahhir (approved):
Very nice!

Issue #2001, opened 2022-08-05
Incorrectly generated iterator for ComplexList


ignaloidas:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### cdktf & Language Versions

cdktf 0.12.0
Python 3.9.12

### Debug Output

[Debug output](https://gist.github.com/ignaloidas/2764d3345233bf562cce093db5f9df29)

### Expected Behavior

A for loop generated to create a mapping to access properties of the objects in the list

### Actual Behavior

The list converted into a set of objects, which is an invalid argument for `for_each`

### Steps to Reproduce

```python
#!/usr/bin/env python
from constructs import Construct
from cdktf import App, TerraformStack, TerraformIterator
from cdktf_cdktf_provider_aws import AwsProvider, acm, route53


class MyStack(TerraformStack):
    def __init__(self, scope: Construct, id_: str) -> None:
        super().__init__(scope, id_)
        self.root_domain_name = "example.com"
        self.zone_id = "EXAMPLE_ID"
        self.aws_provider = AwsProvider(self, "AwsProvider")

        acm_certificate = acm.AcmCertificate(
            self,
            "AcmCertificate",
            domain_name=f"*.{self.root_domain_name}",
            subject_alternative_names=[self.root_domain_name],
            validation_method="DNS",
        )

        iterator = TerraformIterator.from_list(
            acm_certificate.domain_validation_options
        )

        validation_records = route53.Route53Record(
            self,
            "ValidationRecords",
            for_each=iterator,
            name=iterator.get_string("resource_record_name"),
            records=[iterator.get_string("resource_record_value")],
            type=iterator.get_string("resource_record_type"),
            allow_overwrite=True,
            ttl=300,
            zone_id=self.zone_id,
        )


app = App()
MyStack(app, "cdktf_problem")
app.synth()
```

### References
Prior references on how to work around this:
https://discuss.hashicorp.com/t/issue-with-acm-certs-and-route53-in-cdktf/22270
https://github.com/hashicorp/terraform-cdk/issues/430#issuecomment-831511019

jsteinich:
[...]

Issue #1393, opened 2021-12-06
Acm certificate validation not handled properly


Blakeinstein:
https://github.com/hashicorp/terraform-cdk/blob/158d8575f520f17a5adf76d46ad9840e5888cca8/examples/typescript/aws-cloudfront-proxy/main.ts#L84

I referenced the code above, and produce something like the following.

```ts
const certificateResource = new ACM.AcmCertificate(this, props.domainName + '_certificate', {
  domainName: props.domainName,
  validationMethod: 'DNS',
  tags: {
    environment: props.environment,
  },
  lifecycle: {
    createBeforeDestroy: true,
  },
});
certificate = certificateResource.arn;
const validationRecord = new Route53.Route53Record(this, props.domainName + '_certificate_record', {
  zoneId: hostedZone.zoneId,
  name: certificateResource.domainValidationOptions('0').resourceRecordName,
  type: certificateResource.domainValidationOptions('0').resourceRecordType,
  records: [
    certificateResource.domainValidationOptions('0').resourceRecordValue,
  ],
  ttl: 60,
});

certificateDependable = new ACM.AcmCertificateValidation(this, props.domainName + '_certificate_validation', {
  certificateArn: certificateResource.arn,
  validationRecordFqdns: [
    validationRecord.fqdn,
  ],
});
```

Raises the following error
```
 on cdk.tf.json line 65, in resource.aws_route53_record.typescript-aws_examplecom_certificate_record_67B1E051:
│   65:         "type": "${aws_acm_certificate.typescript-aws_examplecom_certificate_A8CEFE35.domain_validation_options.0.resource_record_type}",
Elements of a set are identified only by their value and don't have any
separate index or key to select with, so it's only possible to perform
operations across all elements of the set.
```

Futhermore I would like add a few alternative names as discussed [here](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/guides/version-3-upgrade#resource-aws_acm_certificate) and would like to know the syntax for the same.

jsteinich:
I'm sure there's an example somewhere for this, but the closet I could find was https://github.com/hashicorp/terraform-cdk/issues/42#issuecomment-950333267

Blakeinstein:
> I'm sure there's an example somewhere for this

I don't think that will work since as suggested by the comment, `domainValidationOptions` is now a function

Blakeinstein:
A solution that could work is printing the appropriate hcl code via override.

[...]

Issue #430, opened 2020-11-17
Iterating computed values


hampus77:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### cdktf & Language Versions

cdktf-cli@0.0.18
cdktf@0.0.18
typescript@4.0.5

### Description
I´m trying to figure out how to do work with computed values and loops with the kbst/kustomize provider where they provide an ids list which contain computed values which in turn will be used to get the kustomized manifests.

data "kustomization" "test" {
  provider = kustomization
  path = "test_kustomizations/basic/initial"
}

resource "kustomization_resource" "test" {
  provider = kustomization
  for_each = data.kustomization.test.ids
  manifest = data.kustomization.test.manifests[each.value]
}

How would I go about and do this in the cdk?

Regards,
Hampus

skorfmann:
There's no abstraction over loops yet. However, you can achieve this by using [escape hatches](https://cdk.tf/escape-hatch) for now

skorfmann:
In the context of a cert validation in AWS (see [Terraform docs](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/acm_certificate_validation)) here's a working example for using `for_each`:

Given this HCL 

```hcl
resource "aws_route53_record" "example" {
  for_each = {
    for dvo in aws_acm_certificate.example.domain_validation_options : dvo.domain_name => {
      name   = dvo.resource_record_name
      record = dvo.resource_record_value
      type   = dvo.resource_record_type
    }
  }

  allow_overwrite = true
  name            = each.value.name
  records         = [each.value.record]
  ttl             = 60
  type            = each.value.type
  zone_id         = aws_route53_zone.example.zone_id
}

resource "aws_acm_certificate_validation" "example" {
  certificate_arn         = aws_acm_certificate.example.arn
  validation_record_fqdns = [for record in aws_route53_record.example : record.fqdn]
}
```

We can use escape hatches to make JSON understand the `for_each` concept:

```typescript
const record = new aws.Route53Record(this, 'validation-record', {
  name: "${each.value.name}",
  type: "${each.value.type}",
[...]

PR #3273, opened 2023-11-17
feat(lib): Add DynamicListTerraformIterator with support for lists or sets with objects that have properties that are only known at apply time

### LINKED ISSUE #2178: how to implement AcmCertificateValidation in typescript?
### Community Note

### Description

I don't believe it is currently possible (CDKTF 0.13.0) in Typescript to generate a ACM Certificate because of the complex inputs needed in an [acm_certificate_validation](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/acm_certificate_validation#dns-validation-with-route-53) operation. Setting up a certificate is a fundamental piece before creating more infrastructure.

Please let me know if I have somehow missed an example that shows it is possible to validate a certificate with DNS validation.

There is some discussion around [iterators](https://www.terraform.io/cdktf/concepts/iterators). It seems to me the biggest problem is trying to re-use `for_each` from HCL instead of providing a [proper iterator](https://github.com/hashicorp/terraform-cdk/issues/2001#issuecomment-1275344179) type in higher level languages.

### References

- https://github.com/hashicorp/terraform-cdk/issues/430
- https://github.com/hashicorp/terraform-cdk/issues/2001

ansgarm:
> Note! This PR is based on #3272 

## Problem
Terraform does not support iterating over lists without turning them into a set first. Sets that contain computed values need to be applied using `-target` in order for Terraform to know about these values before a plan is made. Using `-target` and applying changes in two passes is not desirable. However, there’s a workaround by turning them into maps which is described below.

### Workaround by using Maps
A workaround for this case is to turn the set of objects into a map with keys that are based on arguments that are known during the plan stage. This is done for example in the [HCL docs](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/acm_certificate) snippet for the `aws_acm_certificate_validation` resource:

image

In the snippet above, the set of objects stored in `domain_validation_options` is converted into a map with the `domain_name` of each object as the key.

If one instead tries to directly use `domain_validation_options`, the following error is printed (hinting at using `-target`, as described in the problem section above):

image

## Solution
Allow the user to pass the attribute name to be used as the key in the map and convert the set/list into a map using a for expression.

Closes #2178
Related to #2001, #1393, #430 (some already closed with workarounds)


packages/cdktf/lib/terraform-iterator.ts
```diff
@@ -66,6 +70,22 @@ export abstract class TerraformIterator implements ITerraformIterator {
...
   }
 
+  /**
+   * Creates a new iterator from a complex list. One example for this would be a list of maps.
+   * The list will be converted into a map with the mapKeyAttributeName as the key.
+   * @param list the list to iterate over
+   * @param mapKeyAttributeName the name of the attribute that should be used as the key in the map
+   *
+   * @example
+   * TODO: put AWS certificate validation example
```
DanielMSchmidt, line 80: Leftover TODO :)

packages/cdktf/lib/terraform-iterator.ts
```diff
@@ -66,6 +70,22 @@ export abstract class TerraformIterator implements ITerraformIterator {
...
+   */
```
DanielMSchmidt, line 81: Maybe we can also link to the not-yet-existing documentation page, I think the topic is complex enough that it warrants a new section in some page

test/typescript/iterators/test.ts
```diff
@@ -60,6 +60,18 @@ describe("iterators integration test", () => {
...
+
+    expect(stack2.byId("record")).toHaveProperty(
+      "for_each",
+      "${{ for key, val in tolist(aws_acm_certificate.cert.domain_validation_options): val.domain_name => val }}"
+    );
+    expect(stack2.byId("record")).toHaveProperty("name", "${each.value.name}");
+  });
+
+  test("non-synth stack is valid Terraform", async () => {
+    await driver.exec("terraform", ["init"]);
```
DanielMSchmidt, line 72: Nit: maybe the init & validate combination could be a new driver method

DanielMSchmidt (commented):
I like the approach overall and to my knowledge it looks good 👍 There are a few things missing for the overall picture from my point of view, let me know what you think:

- [ ] Documentation section about this (detailing the use-case & usage)
- [ ] Hcl2cdk needs to be made aware of this, ideally we would detect dynamic blocks / for_each being used from a complex list and then add the correct fromComplexList invocation. This will allow us to remove yet another escape hatch 🎉

ansgarm:
> I like the approach overall and to my knowledge it looks good 👍 There are a few things missing for the overall picture from my point of view, let me know what you think:
> 
>     * [ ]  Documentation section about this (detailing the use-case & usage)
> 
>     * [ ]  Hcl2cdk needs to be made aware of this, ideally we would detect dynamic blocks / for_each being used from a complex list and then add the correct fromComplexList invocation. This will allow us to remove yet another escape hatch 🎉

Added docs for this and linked them in the docstring as well.

In regards to `hcl2cdk` support: While it might be possible to detect for expression being used to convert set data structures containing computed values, I would put this out of scope for this PR as that seems a bit too complex to add onto this one.

examples/python/documentation/Pipfile
```diff
@@ -7,4 +7,4 @@ verify_ssl = true
 python_version = "3.7"
 
 [packages]
-cdktf = "*"
+cdktf = {path = "./../../../dist/python/cdktf-0.19.1.dev1700818458-py3-none-any.whl"}
```
DanielMSchmidt, line 10: This change is probably not wanted (and I think if you use the reinstall command in the package.json not needed)

DanielMSchmidt (approved):

examples/python/documentation/Pipfile
ansgarm: Whoops, yeah, that must've happened automatically somehow

ansgarm:
Merging even though C# example fails on the `pluckProperty` call, which is a problem that's not directly related to this branch and something that @DanielMSchmidt is looking into.

PR #3285, opened 2023-11-21
fix(provider-generator): fix oci provider by skipping attributes that conflict after being camel cased

ansgarm:


mutahhir (approved):

ansgarm:
> Is there a provider test for OCI?

Good call, @DanielMSchmidt! There is a provider integration test at this point, but as soon as the deprecated attributes are removed, nothing tests this case anymore 😅 I'll add a test in a separate branch in order to get this in quicker as it should unblock our pre-release ✌️

PR #3286, opened 2023-11-21
chore: Run all tests per PR unless skipped

mutahhir:
### Description

Enable PR checks to be run for all PRs, since we now have the merge queue. This prevents us from merging things that break the release pipeline.

### DIFF
--- a/.github/workflows/examples-pr.yml
+++ b/.github/workflows/examples-pr.yml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   examples:
-    if: ${{ contains(github.event.pull_request.labels.*.name, 'ci/run-examples') || contains(github.event.pull_request.labels.*.name, 'ci/run-all') }}
+    if: ${{ !contains(github.event.pull_request.labels.*.name, 'ci/skip-examples') && !contains(github.event.pull_request.labels.*.name, 'ci/unit-only') }}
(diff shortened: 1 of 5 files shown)


ansgarm:
I think that would still not change how the merge queue works 🤔

As far as I understand it, you need to explicitly define what to run in merge queues.
For example we currently have this:
https://github.com/hashicorp/terraform-cdk/blob/9f70e6f82dadb0f616480df6aec4789138e8cf9c/.github/workflows/pr-unit.yml#L3-L4

(and we'd need that for all kinds of checks that we run on PRs)

mutahhir:
> I think that would still not change how the merge queue works 🤔
> 
> As far as I understand it, you need to explicitly define what to run in merge queues. For example we currently have this:
> 
> https://github.com/hashicorp/terraform-cdk/blob/9f70e6f82dadb0f616480df6aec4789138e8cf9c/.github/workflows/pr-unit.yml#L3-L4
> 
> (and we'd need that for all kinds of checks that we run on PRs)
> 
> edit: Or does a merge queue run PR checks by default?

Yeah, just testing for now. I think since the rules are for branch protections, the merge queue should not be able to merge to that branch, but let's see.

DanielMSchmidt (approved):

PR #3289, opened 2023-11-22
fix(tests): Update provider version in C# snapshot test

ansgarm:


DanielMSchmidt (approved):
Why does this change, shouldn't it be pinned? 😕

ansgarm:
The test uses a version constraint that allows patch level upgrades (`~> 3.2.0`).

I don't have a strong opinion whether we should pin it, as it also tests handling constraints that don't pin an exact version 🤔

But the handling is probably not different from what e.g. `=3.2.2` result in (aka the constraint is only printed afaik).

PR #3302, opened 2023-11-28
chore: re-enable operator example

DanielMSchmidt:


packages/cdktf/lib/terraform-operators.ts
```diff
@@ -21,43 +22,43 @@ export class Op {
   /**
    * Renders -expression
    */
-  public static negate(expression: Expression): IResolvable {
-    return new OperatorExpression("-", expression);
+  public static negate(expression: Expression): number {
```
ansgarm, line 25: In theory this could also be used on a boolean, but I think it's fine to have to use Token.asBoolean() then :)

ansgarm (approved):

packages/cdktf/lib/terraform-operators.ts
DanielMSchmidt: Yeah good point, although one would use the not operator (!) there then, right?

PR #3333, opened 2023-12-04
feat(docs): Add note to address having the same name as new Terraform stacks feature

mpron:
Since CDKTF has a primitive called stacks and we also have an upcoming (now announced) feature that is also called "stacks" or "Terraform stacks" - I wrote a sentence with a link to point out the differences.

### DIFF
--- a/website/docs/cdktf/concepts/stacks.mdx
+++ b/website/docs/cdktf/concepts/stacks.mdx
@@ -7,6 +7,8 @@ description: >-
 
 # Stacks
 
+_Note: Stacks in CDKTF are not the same as [Terraform stacks](https://www.hashicorp.com/blog/terraform-stacks-explained), which are a configuration layer that eliminates the need to manually manage cross-configuration dependencies between Terraform modules.
+
 A stack represents a collection of infrastructure that CDK for Terraform (CDKTF) synthesizes as a dedicated Terraform configuration. Stacks allow you to separate the state management for multiple environments within an application.
 
 > **Hands-on:** Try the [Deploy Applications with CDK for Terraform](/terraform/tutorials/cdktf/cdktf-applications?utm_source=WEBSITE&utm_medium=WEB_IO&utm_offer=ARTICLE_PAGE&utm_content=DOCS) tutorial.


mpron:
This is my first contribution to dev dot and I didn't read every single piece of submission guidance, so redirect me if need be, but it is just one line and you can take it from here and rework as you see fit.

website/docs/cdktf/concepts/stacks.mdx
```diff
@@ -7,6 +7,8 @@ description: >-
 
 # Stacks
 
+_Note: Stacks in CDKTF are not the same as [Terraform stacks](https://www.hashicorp.com/blog/terraform-stacks-explained), which are a configuration layer that eliminates the need to manually manage cross-configuration dependencies between Terraform modules.
```
ansgarm, line 10: Thanks for the explaining comment, @mpron! Appreciate it.

A small thing that might cause confusion in this sentence, is the following:
> […] that eliminates the need to manually manage cross-configuration dependencies between Terraform modules

CDKTF Stacks support what we call "Cross Stack References" ([docs](https://developer.hashicorp.com/terraform/cdktf/concepts/stacks#cross-stack-references)), which automatically create outputs and Terraform remote state data sources when references are passed across stacks.

Terraform stacks, however, offer more ease of use, as they also include all the access rights stuff which need to be manually ensured for CDKTF stacks (i.e. you need to allow your workspaces in TFC to access each other).

Maybe there's a way to differentiate Terraform stacks in a different way? E.g. by calling them something like "a TFC native way of creating infrastructure applications consisting of multiple components and environments" (I'm no expert here, dunno if that's correct).

mpron: I was just using the tagline descriptor used in the TF stacks introductory video.  Another way stacks were described is a configuration layer that "helps users automate and optimize the coordination, deployment, and lifecycle management of interdependent Terraform configurations."  Do you like that better? And we can't say it's just for TFC because its being built with core functionality in core terraform (community).

mpron: Feel free to pull in another SME from TF edu or Dan Barr.

ansgarm: > And we can't say it's just for TFC because its being built with core functionality in core terraform (community).

Forgot about that aspect. I checked with the team and we're fine with the current wording 👍

ansgarm (approved):

PR #3340, opened 2023-12-07
chore: pin all dependencies

DanielMSchmidt:


yarn.lock
```diff
@@ -417,6 +417,15 @@
     debug "^4.1.0"
     globals "^11.1.0"
 
+"@babel/types@7.21.4", "@babel/types@^7.21.4":
```
ansgarm, line 420: Quick question: The `@babel/types@^7.21.4` one (with the caret) is a transitive dependency, right? So basically a dependency of ours using that version constraint?

ansgarm (approved):
Looks good (if CI passes 😄), just one small non-blocking question

yarn.lock
DanielMSchmidt: I think so, yeah :)

packages/@cdktf/provider-schema/src/__tests__/provider-schema.test.ts
```diff
@@ -11,17 +12,12 @@ import {
...
-    if (typeof value !== "object" || value === null) return false;
-
-    return value["format_version"] !== undefined;
-  },
-  print: (value: any) => {
-    value["format_version"] = "STUBBED VERSION";
-    return JSON.stringify(value, null, 2);
-  },
-});
+function stubVersion(value: any) {
```
ansgarm, line 15: I'd call this something like "clean" or "sanitize" 🤔

PR #3341, opened 2023-12-07
fix: Terraform Modules relatively referencing each other with generated bindings

### LINKED ISSUE #2460: 'cdktf get' Cannot Import Local Modules Which Reference Other Modules That Use `../`
### Community Note

### cdktf & Language Versions
- cdktf 0.14.3
- Terraform v1.3.7
- Python 3.10.8

### Expected Behavior

Should be able to import modules which are outside the App directory in the file tree which themselves reference modules.

### Actual Behavior

Attempting to import such a module causes and error such as the following:
```
downloading and generating modules and providers...
╷
│ Error: Local module path escapes module package
│ 
│ The given source directory for module.register.module.constants would be
│ outside of its containing package
│ "file:///Users/jamesbarlow/repo/tf.modules/modules/register/v0".
│ Local source addresses starting with "../" must stay within the same
│ package that the calling module belongs to.
╵

╷
│ Error: Local module path escapes module package
│ 
│ The given source directory for module.register.module.constants would be
│ outside of its containing package
│ "file:///Users/jamesbarlow/repo/tf.modules/modules/register/v0".
│ Local source addresses starting with "../" must stay within the same
│ package that the calling module belongs to.
╵
Error: non-zero exit code 1
    at ChildProcess.<anonymous> (/Users/jamesbarlow/homebrew/lib/node_modules/cdktf-cli/bundle/bin/cmds/handlers.js:45:2064)
    at Object.onceWrapper (node:events:628:26)
    at ChildProcess.emit (node:events:513:28)
    at ChildProcess.emit (node:domain:489:12)
    at maybeClose (node:internal/child_process:1098:16)
    at Socket.<anonymous> (node:internal/child_process:456:11)
    at Socket.emit (node:events:513:28)
    at Socket.emit (node:domain:489:12)
    at Pipe.<anonymous> (node:net:305:12)
```

This error appears to be coming from the Terraform CLI itself rather than cdktf, but this pattern works fine when using HCL-based Terraform.

### Steps to Reproduce

- Create a module in a directory such as `/Users/jamesbarlow/repo/tf.modules/modules/register/v0` and have this module itself reference another module with a relative path:
```
module "constants" {
  source = "../../constants"
}
```
So, in this example the `constants` module is located at `/Users/jamesbarlow/repo/tf.modules/modules/constants`

- Create a cdktf App directory at `/Users/jamesbarlow/repo/learn-cdktf-docker`
- Setup the `cdktf.json` to reference the `register` module that we want to generate the code bindings for, and include the `constants` module that is referenced by that module for good measure:

```json
"terraformModules": [
    {
      "name": "constants",
      "source": "../tf.modules/modules/constants"
    },
    {
      "name": "register",
      "source": "../tf.modules/modules/register/v0"
    }
  ]
```
- Run `cdktf get`. The error is emitted.

### Important Factoids

If a module which itself does not reference another module is imported, this works fine. For example, the `constants` module imports fine when specified in isolation in the `terraformModules` as this does not use another module.

### References

According to [here](https://github.com/hashicorp/terraform/issues/23333#issuecomment-1261531854) there is special treatment for relative paths and this error will occur for absolute paths.

I would guess that cdktf is "incorrectly" pre-converting the relative path to an absolute path before passing the path to Terraform, hence causing this error.

ansgarm:
When multiple local terraform modules are mentioned in the `cdktf.json` and they rely on one-another locally with relative paths the previous implementation destroyed the structure of their relative paths. This implementation uses a single asset for all terraform modules, leading to correct relative paths.

Closes #2460

### DIFF
--- a/packages/@cdktf/cli-core/src/lib/synth-stack.ts
+++ b/packages/@cdktf/cli-core/src/lib/synth-stack.ts
@@ -4,9 +4,15 @@ import * as fs from "fs-extra";
 import * as path from "path";
 import * as chalk from "chalk";
 import indentString from "indent-string";
-import { Manifest, StackManifest, TerraformStackMetadata } from "cdktf";
+import {
+  CONTEXT_ENV,
+  Manifest,
+  StackManifest,
+  TerraformStackMetadata,
+} from "cdktf";
 import { performance } from "perf_hooks";
 import { logger, readConfigSync, sendTelemetry, shell } from "@cdktf/commons";
+import { CdktfConfig } from "./cdktf-config";
 
 const chalkColour = new chalk.Instance();
 
@@ -111,6 +117,17 @@ might fail while synthesizing with an out of memory error.`);
       env.NODE_OPTIONS = `${env.NODE_OPTIONS || ""} ${nodeOptsSetting}`.trim();
     }
 
+    const currentContext = process.env[CONTEXT_ENV]
+      ? JSON.parse(process.env.CDKTF_CONTEXT_JSON)
+      : {};
+    const relativeModules = CdktfConfig.read().terraformModules.filter(
+      (mod) => {
+        return typeof mod === "string"
+          ? mod.startsWith("./") || mod.startsWith("../")
+          : mod.source.startsWith("./") || mod.source.startsWith("../");
+      }
+    );
+
     try {
       await shell(command, [], {
         shell: true,
@@ -118,6 +135,10 @@ might fail while synthesizing with an out of memory error.`);
           ...env,
           CDKTF_OUTDIR: outdir,
           CDKTF_CONTINUE_SYNTH_ON_ERROR_ANNOTATIONS: "true", // we want to display the errors ourselves
+          [CONTEXT_ENV]: JSON.stringify({
+            ...currentContext,
+            cdktfRelativeModules: relativeModules,
+          }),
         },
         cwd: workingDirectory,
         signal: abortSignal,
--- a/packages/@cdktf/cli-core/src/lib/synth-stack.ts
+++ b/packages/@cdktf/cli-core/src/lib/synth-stack.ts
@@ -120,13 +120,7 @@ might fail while synthesizing with an out of memory error.`);
     const currentContext = process.env[CONTEXT_ENV]
       ? JSON.parse(process.env.CDKTF_CONTEXT_JSON)
       : {};
-    const relativeModules = CdktfConfig.read().terraformModules.filter(
-      (mod) => {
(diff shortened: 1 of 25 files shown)


test/typescript/modules-relative-paths/cdktf-project/main.ts
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) HashiCorp, Inc
+// SPDX-License-Identifier: MPL-2.0
+import { Construct } from "constructs";
+import { App, TerraformStack, Testing, LocalBackend } from "cdktf";
+
+export class HelloTerra extends TerraformStack {
+  constructor(scope: Construct, id: string) {
+    super(scope, id);
+
+    // TODO: put code here
```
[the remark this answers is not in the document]
ansgarm: Okay, so this fails during get already. Usage of that module should still added though, to confirm the result actually works 😅

packages/@cdktf/cli-core/src/lib/synth-stack.ts
```diff
@@ -243,3 +258,19 @@ Command output on stdout:
     await sendTelemetry("synth", { error: true, synthOrigin });
   }
 }
+
+function getRelativeTerraformModules() {
+  let cfg;
+
+  try {
+    cfg = CdktfConfig.read();
+  } catch {
```
ansgarm, line 267: We should at least add a trace log with the error in order to be able to debug this

packages/cdktf/lib/terraform-module-asset.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    const relativeModules: Array<string | { source: string }> | undefined =
+      this.node.tryGetContext("cdktfRelativeModules");
+    const staticModuleAssetHash: string | undefined = this.node.tryGetContext(
+      "cdktfStaticModuleAssetHash"
+    );
+
+    if (!relativeModules) {
+      throw new Error(
+        "You are trying to use a local module with a relative path, but the cdktfRelativeModules context is not set. It is expected to be an array of strings containing the relative paths to. You either need to supply it in the Apps constructor via the context option or invoke the synthesis through the CLI. We need this information so that assets with relative paths are properly handled when used with assets, so you can also set the skipAssetCreationFromLocalModules to true on your relative modules."
```
ansgarm, line 41: > It is expected to be an array of strings containing the relative paths to.
Seems like something is missing in this sentence?

ansgarm (commented):
Looks good! Only two small things, but feel free to merge after they've been addressed ✅

PR #3363, opened 2024-01-02
chore: update docker base image

DanielMSchmidt:


ansgarm (approved):

ansgarm:
I think you can ignore the tests and merge, they won't run against the new image until it is built anyway 😄

PR #3365, opened 2024-01-02
feat: Synthesis in HCL

### LINKED ISSUE #225: [Feature] Add Support for HCL Output
### Community Note

### Description

Currently the output is JSON, but it would be great if we could output HCL.
While we can use `cd cdktf.out && terraform plan`, but could see real value/interesting solutions created if the output was HCL.

ansgarm:
### Related issue

Fixes #225 

### Description

Add support for CDKTF to synthesize code to HCL Terraform instead of just JSON. While we are trying to ensure that HCL generation is as good as the existing JSON synthesis, care is requested when using it as JSON synthesis is much better tested and robust.

In order to use HCL output, the `--hcl` flag can be passed on to the `synth` command (e.g. `cdktf synth --hcl`). 

> [!WARNING]
> Please note that we currently don't erase / remove existing JSON files, so care has to be exercised when using both `cdktf synth` along with `cdktf synth --hcl` as terraform will try to use both files if present and that will cause problems.

## Missing features / Problems

- Some keywords may be quoted when not necessary. e.g. `when = destroy` will be `when = "destroy"` within the provisioner block.
- Overrides do not contain enough type information, so they may be rendered incorrectly. We try our best to find the right rendering, but it's not always possible.


DanielMSchmidt (approved):
🎉🎉🎉

PR #3379, opened 2024-01-05
chore: run python tests in virtual env

DanielMSchmidt:
Installing packages requires a non-root environment

### DIFF
--- a/test/python/provider-add-command/test.ts
+++ b/test/python/provider-add-command/test.ts
@@ -1,8 +1,7 @@
 // Copyright (c) HashiCorp, Inc
 // SPDX-License-Identifier: MPL-2.0
 import { TestDriver } from "../../test-helper";
-// Disabled, needs pipenv environment to be setup before cdktf provider commands can run
-describe.skip("provider add command", () => {
+describe("provider add command", () => {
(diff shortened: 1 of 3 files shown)


test/python/provider-add-command/test.ts
```diff
@@ -49,9 +48,14 @@ describe.skip("provider add command", () => {
         await driver.setupPythonProject({
           init: { additionalOptions: "--cdktf-version 0.10.4" },
         });
+        driver.setEnv("PIPENV_VERBOSITY", "-1");
```
mutahhir, line 51: Minor/Nit: Leaving a comment here would be useful to recall why we did this.

```
Supress warning that Pipenv is running within a virtual environment
```

mutahhir (approved):
:) Mostly don't get what's going on, but I trust you

DanielMSchmidt:
Just for context: We updated to a newer Python version, which marks the global install location for pip as system managed, throwing an error if against it without a special flag. For our use-cases and the ones of our customers they will most likely will use a virtual env, which we now need to set up in our tests. We updated the python version because the latest (1.93.0) version of JSII dropped support for python 3.7.

PR #3442, opened 2024-01-19
feat(provider-generator): support renaming provider names in order to allow using two providers sharing the same name at the same time

### LINKED ISSUE #1258: Ensure two providers with same name but from different namespaces can be used at the same time
> I'm not sure where the cdktf.json schema is documented, but I tried a few variations of the block syntax
> ```
>    "terraformProviders": [
>     {
>        "name": "google-beta",
>        "source": "google-beta",
>        "version": "4.0.0"
>      }
>    ],
>```
>
> and nothing helped with collisions. I hoped that something similar to terraform's [alias field](https://www.terraform.io>/docs/language/providers/configuration.html#alias-multiple-provider-configurations) would work, since collisions need to be handled somehow (e.g. how do I use both `Runelab/bitbucket` and `aeirola/bitbucket`?), but maybe this is a separate issue.

_Originally posted by @figadore in https://github.com/hashicorp/terraform-cdk/pull/1252#discussion_r743060433_

ansgarm:
Adds support for the Terraform feature described here: https://developer.hashicorp.com/terraform/language/providers/requirements#handling-local-name-conflicts

Resolves #1258

### Open items
- [x] add test case for JSON and HCL output when using two providers with the same name

### DIFF
--- a/packages/@cdktf/provider-generator/lib/get/generator/provider-generator.ts
+++ b/packages/@cdktf/provider-generator/lib/get/generator/provider-generator.ts
@@ -26,18 +26,60 @@ const isMatching = (
 ): boolean => {
   if (target.isModule) return false;
 
+  console.log("isMatching", { target, terraformSchemaName });
+
   const elements = terraformSchemaName.split("/");
 
   if (elements.length === 1) {
     return target.source === terraformSchemaName;
   } else {
-    const [hostname, scope, provider] = elements;
+    const [hostname, namespace, provider] = elements;
 
-    if (!hostname || !scope || !provider) {
+    if (!hostname || !namespace || !provider) {
       throw new Error(`can't handle ${terraformSchemaName}`);
     }
 
-    return target.name === provider || target.source === terraformSchemaName;
+    console.log("target.name", target.name);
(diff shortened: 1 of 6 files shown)


packages/@cdktf/provider-generator/lib/get/generator/provider-generator.ts
```diff
@@ -26,18 +26,60 @@ const isMatching = (
...
-    const [hostname, scope, provider] = elements;
+    const [hostname, namespace, provider] = elements;
 
-    if (!hostname || !scope || !provider) {
+    if (!hostname || !namespace || !provider) {
       throw new Error(`can't handle ${terraformSchemaName}`);
     }
 
-    return target.name === provider || target.source === terraformSchemaName;
+    console.log("target.name", target.name);
```
DanielMSchmidt, line 42: Logs are left here

DanielMSchmidt (approved):

packages/@cdktf/provider-generator/lib/get/generator/provider-generator.ts
ansgarm: Ah! thx

Issue #2535, opened 2023-01-25
TerraformVariable not working


tieto-hemmopaa:
### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

### cdktf & Language Versions

Node v18.13.0
Terraform v1.3.4
cdktf 0.15.1

### Affected Resource(s)

TerraformVariable

### Expected Behavior

Creating a TerraformVariable should get the values of environment variables `TF_VAR_awsAccessKey` and `TF_VAR_awsSecretAccessKey`.

Running `cdktf deploy test-stack-1` or `cdktf deploy test-stack-2` should be successful.

main.ts:

```
import { Construct } from 'constructs';
import { App, TerraformStack, S3Backend, TerraformVariable } from 'cdktf';
import { AwsProvider } from './.gen/providers/aws/provider';
import { S3Bucket } from './.gen/providers/aws/s3-bucket';
import { DynamodbTable } from './.gen/providers/aws/dynamodb-table';

interface MultiStackConfig {
    environment: string;
}
class MultiStack extends TerraformStack {
  constructor(scope: Construct, id: string, config: MultiStackConfig) {
    super(scope, id);
    
    const awsAccessKey = new TerraformVariable(this, 'awsAccessKey', {
      type: 'string',
      description: 'AWS Access Key',
      sensitive: true
    })
    const awsSecretAccessKey = new TerraformVariable(this, 'awsSecretAccessKey', {
      type: 'string',
      description: 'AWS Secret Access Key',
      sensitive: true
    })
    new AwsProvider(this, 'aws', {
      region: 'eu-west-1',
      accessKey: awsAccessKey.value,
      secretKey: awsSecretAccessKey.value
    });

    const { environment } = config;

    new S3Bucket(this, 's3_backend_bucket', {
      bucket: `cdktf-remote-backend-${environment}`,
    });
    new DynamodbTable(this, 'dynamodb_lock_table', {
      name: `cdktf-remote-backend-lock-${environment}`,
      billingMode: 'PAY_PER_REQUEST',
      attribute: [
        {
          'name': 'LockID',
          'type': 'S'
        }
[...]

PR #3446, opened 2024-01-22
chore: Throw error when variables are used within backend configuration

mutahhir:
### Related issue

Fixes https://github.com/hashicorp/terraform-cdk/issues/2535

### Description

Terraform doesn't support variables within backends, but CDKTF doesn't prevent users from going this route. With this PR, we try to detect variable usage within the backend, and will throw an error when encountered.

Note: The proper way of using credentials etc within backends is outlined [here](https://github.com/hashicorp/terraform-cdk/issues/2535), but supporting that within the CDKTF CLI would need more work.

### DIFF
--- a/packages/cdktf/lib/terraform-backend.ts
+++ b/packages/cdktf/lib/terraform-backend.ts
@@ -37,29 +37,95 @@ export abstract class TerraformBackend extends TerraformElement {
   ): TerraformRemoteState;
 
   public toHclTerraform(): any {
+    const backendConfig = deepMerge(
+      this.synthesizeHclAttributes(),
+      this.rawOverrides
+    );
+
+    const variables = this.detectVariables(backendConfig);
+
+    if (variables.length > 0) {
+      throw new Error(
+        [
+          "Terraform does not support interpolated values within backends.",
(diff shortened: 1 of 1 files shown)


packages/cdktf/lib/terraform-backend.ts
```diff
@@ -37,29 +37,95 @@ export abstract class TerraformBackend extends TerraformElement {
...
+    const variables = this.detectVariables(backendConfig);
+
+    if (variables.length > 0) {
+      throw new Error(
+        [
+          "Terraform does not support interpolated values within backends.",
+          "Interpolations found in the following configuration backend attributes:",
+          ...variables.map((v) => `\t${this.constructor.name}: "${v}"`),
+        ].join("\n")
+      );
```
ansgarm, line 122: I'd prefer to add an error annotation instead. This will automatically print the construct id alongside and it won't short circuit after the first error.
Here's how they look for example: https://github.com/hashicorp/terraform-cdk/blob/5627fe99b08e931c32594a2403e05c86d9cc6e91/test/typescript/synth-error-annotations/test.ts#L23

PR #3468, opened 2024-01-26
fix: corrected formating for java provider dependencies

DanielMSchmidt:


test/run-against-dist
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -euo pipefail
```
mutahhir, line 2: Do we need the debugging flag?

test/java/provider-list-command/test.ts
```diff
@@ -34,6 +34,8 @@ describe("provider list command", () => {
     test("with json output", async () => {
       const res = await driver.exec("cdktf", ["provider", "list", "--json"]);
 
+      console.log(res);
```
mutahhir, line 37: Is this for debugging only? and we'll try to remove it later?

mutahhir (approved):

PR #3471, opened 2024-01-27
Fix(cli): A bug when opening browser using CLI hangs the shell process forever

razine-bensari:
### Related issue

Fixes #3470 

### Description

When we request an API token from terraform cloud (during `cdktf login` or `cdktf init`...) we do a sys call to open the browser. This sys call opens a new tab using the default browser defined by the OS. We pass an option when making this syscall called `{... wait: true}`. Based on the docs, this option will wait/block the standard input until the user exits/quits the browser. Not just the tab, the browser process. This is not user friendly, nor required in my opinion. As a result, when the browser opens and we copy the newly generated token then proceed to paste it in the terminal, there is no standard input to paste it in, the process is blocked.

What I did was simply remove this wait=true option since there is no need to wait for the browser to quit/exit in order to paste our token.

### Checklist

- [ ] I have updated the PR title to match [CDKTF's style guide](https://github.com/hashicorp/terraform-cdk/blob/main/CONTRIBUTING.md#pull-requests-1)
- [ ] I have run the linter on my code locally
- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation if applicable
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works if applicable
- [ ] New and existing unit tests pass locally with my changes

PR #3480, opened 2024-02-05
fix(cli): opening browser using CLI hangs the shell process forever

DanielMSchmidt:
Ran prettier on top of #3471


ansgarm (approved):

ansgarm:
The new test-case seems to fail in CI: https://github.com/hashicorp/terraform-cdk/actions/runs/7782541799/job/21219202309?pr=3480#step:12:93

image

PR #3495, opened 2024-02-12
fix: Make `CDKTF_LOG_FILE_DIRECTORY` a common flag available on all commands

### LINKED ISSUE #3489: CDKTF_LOG_FILE_DIRECTORY: 'cdktf init' errors out with 'Unknown argument' when this environment variable is set
### Actual Behavior

* If **CDKTF_LOG_FILE_DIRECTORY** is set on the system, then the `cdktf init` command **errors** out with the following message
  * > Unknown argument: logFileDirectory
  
  [The environment variable is specified with a path of an existing folder (with sufficient read/write access)]

* However, if this environment variable is set ***after**  `cdktf init`* is executed, then:
  * All the **other commands** (get, synth, deploy, diff, destroy, etc.) support CDKTF_LOG_FILE_DIRECTORY & **successfully** log the content in the specified directory.

### Expected Behavior

* Is it intentional that *CDKTF_LOG_FILE_DIRECTORY* is **not** supported during **init** & it does not allow logging to a specific directory? **If so:**
  * Is it because logging is solely for ***debugging*** CDKTF applications, so we want to save the logs only **after** the application is created to debug it [**Reference:** [Debugging](https://developer.hashicorp.com/terraform/cdktf/test/debugging)]

  * Or is it because we want project-specific logs?
     * As CDKTF allows managing different projects, streaming logs from two different CDKTF projects to the same file is not a valid use case.

     * Therefore, the `cdktf init` does not stream logs and expects us to configure logging to separate directories for each application ***after initialization***.

### Steps to Reproduce

**Step 1:** export CDKTF_LOG_FILE_DIRECTORY=<path/to/logging/directory/>

**Step 2:** In an empty directory, run `cdktf init --template=typescript` command.
[This returns an 'Unknown argument' error]

### Versions

1. language: typescript
2. cdktf-cli: 0.17.3
3. npm: 8.19.4
4. node: 16.20.1
5. terraform: 1.4.5

### Providers

_No response_

### Gist

_No response_

### Possible Solutions

_No response_

### Workarounds

_No response_

### Anything Else?

_No response_

### References

_No response_

### Help Wanted

- [ ] I'm interested in contributing a fix myself

### Community Note

mutahhir:
### Related issue

Fixes #3489 

### Description

We were using `CDKTF_LOG_FILE_DIRECTORY` within our logger, but we never made it into a supported CLI flag. CDKTF configures [yargs](https://yargs.js.org/) in a way that all environment variables with the `CDKTF_` prefix are treated as CLI flags. For commands that had [`strict`](https://yargs.js.org/docs/#api-reference-strictenabledtrue) enabled, that means that setting any non-supported flag with the `CDKTF_` prefix would cause the command to fail. This PR now adds support for this flag and it should work across all commands.

### DIFF
--- a/packages/cdktf-cli/src/bin/cdktf.ts
+++ b/packages/cdktf-cli/src/bin/cdktf.ts
@@ -111,6 +111,10 @@ yargs
     required: false,
     desc: "Which log level should be written. Only supported via setting the env CDKTF_LOG_LEVEL",
   })
+  .option("log-file-directory", {
+    require: false,
+    desc: "The directory path where CDKTF should create `cdktf.log` and print logs at the `debug` level. If not set, CDKTF writes logs to standard out at the level specified in `CDKTF_LOG_LEVEL`.",
+  })
   .option("context-json", {
     required: false,
     hidden: true,
--- a/packages/cdktf-cli/src/bin/cdktf.ts
+++ b/packages/cdktf-cli/src/bin/cdktf.ts
@@ -113,7 +113,7 @@ yargs
   })
   .option("log-file-directory", {
     require: false,
-    desc: "The directory path where CDKTF should create `cdktf.log` and print logs at the `debug` level. If not set, CDKTF writes logs to standard out at the level specified in `CDKTF_LOG_LEVEL`.",
+    desc: "The directory path where CDKTF should create `cdktf.log` and print logs at the `debug` level. If not set, CDKTF writes logs to standard out at the level specified in `CDKTF_LOG_LEVEL`. Only supported via setting the env CDKTF_LOG_FILE_DIRECTORY",
   })
   .option("context-json", {
     required: false,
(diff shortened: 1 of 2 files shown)


DanielMSchmidt (approved):
That is certainly an interesting issue :D

packages/cdktf-cli/src/bin/cdktf.ts
```diff
@@ -111,6 +111,10 @@ yargs
     required: false,
     desc: "Which log level should be written. Only supported via setting the env CDKTF_LOG_LEVEL",
   })
+  .option("log-file-directory", {
```
ansgarm, line 114: As we don't pass forward this flag, we might need to add a disclaimer as in the flag above:
> Only supported via setting the env CDKTF_LOG_LEVEL

mutahhir: Good catch! Updated.

PR #3528, opened 2024-02-28
feat(examples): Add OVHcloud Golang example

scraly:
### Description
Add an example in Go / Golang to use the OVHcloud provider.

### Checklist


crw:
Hi @scraly, apologies for the late response. Adding these examples directly to this repository is unlikely to be accepted due to the maintenance burden of having to maintain numerous examples directly in this repository. Typically these examples would be handled on the vendor’s site (in this case, OVHcloud). Please see [this as a reference](https://docs.databricks.com/en/dev-tools/terraform/cdktf.html). It is also conceivable such an example could be placed in the provider documentation. Please see [this as a reference](https://registry.terraform.io/providers/aiven/aiven/latest/docs/guides/syslog-service-integration), in this case you would be using CDK syntax instead of standard HCL. You could also publish this as a post to the [HashiCorp Discuss forum](https://discuss.hashicorp.com/), which would emphasize discussion.

In any case, thanks very much for this submission! We appreciate your contributions to the CDK TF project!

PR #3561, opened 2024-03-19
fix: Sort stacks in generated manifest.json file

torjue:
### Description

The sorting of the stacks in the `metadata.json` file generated by `cdktf synth` seems to be random between runs. This makes writing snapshot tests a bit cumbersome.


mutahhir:
Thank you for your PR. We already used https://www.npmjs.com/package/json-stable-stringify elsewhere, so I've updated the `manifest.json` output to also use the library in #3584

PR #3584, opened 2024-04-08
chore: manifest.json should have stable json

mutahhir:
### Description

Ensure that the `manifest.json` file has stable output. Removes instances where test snapshots fail when no content has changed, but `JSON.stringify` shuffles object attribute order. Suggested by #3561

### Checklist

### DIFF
--- a/packages/cdktf/lib/manifest.ts
+++ b/packages/cdktf/lib/manifest.ts
@@ -4,6 +4,7 @@ import * as path from "path";
 import * as fs from "fs";
 import { TerraformStack } from "./terraform-stack";
 import { AnnotationMetadataEntryType } from "./annotations";
+import stringify = require("json-stable-stringify");
 
 export interface StackAnnotation {
   readonly constructPath: string;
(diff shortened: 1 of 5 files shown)


packages/cdktf/lib/manifest.ts
```diff
@@ -4,6 +4,7 @@ import * as path from "path";
 import * as fs from "fs";
 import { TerraformStack } from "./terraform-stack";
 import { AnnotationMetadataEntryType } from "./annotations";
+import stringify = require("json-stable-stringify");
```
DanielMSchmidt, line 7: Nit: does this need to be a require import or can it be a normal static one?

DanielMSchmidt (approved):

packages/cdktf/lib/manifest.ts
mutahhir: I think it has to be. I tried the other one first, but the LSP told me not to do that.

DanielMSchmidt: ok :) I didn't yet grasp the different import and export styles fully, but if TS is ok with it I won't object :)

PR #3617, opened 2024-05-08
chore(release): ensure a copy of the license is included with published packages

xiehan:
This PR does two things which are similar but not the same:

- It attempts to make us compliant with a new organizational requirement to include a copy of the license file with all of our releases. The actual packaging process is a little bit opaque to me so I'm not 100% positive this solution works across all languages but I was able to verify at the very least that it works as expected for JS.
- It also attempts to make us compliant with what I assume is a long-standing requirement to generate a NOTICES.txt file with all the third-party dependency licenses that we've never followed. AFAIK this is not currently being enforced and I don't know whether they will start enforcing it, but I was curious about whether there's tooling out there to automate this and found that it was easier to implement than I'd expected, so I figured why not push that change as well.


DanielMSchmidt (commented):
Do we need a CI job to make sure these new files are up to date before releasing? I don't think our release process handles file changes in between steps well

DanielMSchmidt (approved):

PR #3654, opened 2024-06-17
fix: explicitly listen on ipv4 address for verdaccio due to Github runner image regression

ansgarm:
Configure Verdaccio to explicitly listen on IPv4 as there's some problem in the latest Github Runner Image with `localhost` and IPv4 / IPv6.


ansgarm:
Alright, so it seems like a recent Github runner image update is the reason for this failure: https://github.com/actions/runner-images/issues/10061

DanielMSchmidt (approved):

PR #3683, opened 2024-07-29
fix(lib): render empty string as quotes

DanielMSchmidt:
Also added a [mise](https://mise.jdx.dev/) config for my development environment


.mise.toml
ansgarm, line 1: ❤️

ansgarm (approved):

PR #3772, opened 2024-11-17
fix(lib): correct the concat method in TokenizedStringFragments

suojae:
The `concat` method was not properly updating `this.fragments` because `Array.prototype.concat` returns a new array and does not modify the original array in place. Changed the implementation to use `this.fragments.push(...other.fragments);` to correctly merge the fragments into `this.fragments`.

### Description

I thought `concat` method in `TokenizedStringFragments` was intended to merge fragments from another instance into `this.fragments`. However, it was not functioning correctly because `Array.prototype.concat` returns a new array and does not modify the original array in place.

To fix this issue, I updated the `concat` method to use `this.fragments.push(...other.fragments);`. This modifies `this.fragments` in place and ensures that all fragments are correctly merged.

### Checklist

### DIFF
--- a/packages/cdktf/lib/tokens/string-fragments.ts
+++ b/packages/cdktf/lib/tokens/string-fragments.ts
@@ -87,7 +87,7 @@ export class TokenizedStringFragments {
   }
 
   public concat(other: TokenizedStringFragments): void {
-    this.fragments.concat(other.fragments);
+    this.fragments.push(...other.fragments);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


DanielMSchmidt:
Hey, thanks for your help! Could you add a test case that shows which user-facing issue this fix solved so that we don't reintroduce the problem by accident in the future?

suojae:
Hi, thank you for your feedback! I have written a test case for the concat method to verify the user-facing issue is resolved. Could you confirm where exactly I should add the test code? I assume it should go into `packages/cdktf/test/tokens/string-fragments.test.ts`, but please let me know if there's a more appropriate location.

Issue #3638, opened 2024-05-29
removed block: orphan resources managed by cdktf/terraform


fathom-parth:
### Description

HCL supports `removed` blocks which allow you to orphan a resource managed by tf easily:
https://developer.hashicorp.com/terraform/language/resources/syntax#removing-resources

If we had something similar in cdktf, it would allow importing resources across stacks much more easily than the method [the documentation](https://developer.hashicorp.com/terraform/cdktf/examples-and-guides/refactoring#moving-resources-from-one-stack-to-another) has you use by instead using something like `removed` and `importFrom` on the cdktf objects.

### Help Wanted

- [ ] I'm interested in contributing a fix myself

### Community Note

- Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
- Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
- If you are interested in working on this issue or have submitted a pull request, please leave a comment

PR #3849, opened 2025-04-13
feat(lib): add the `remove` function to `TerraformResource`

xphir:
Adds support within the `TerraformResource` for the [removed block](https://developer.hashicorp.com/terraform/language/resources/syntax#removing-resources.)

### Related issue

Fixes https://github.com/hashicorp/terraform-cdk/issues/3638

#### Key Changes
- Introduced a `_removed` private field on `TerraformResource` to track removed state.
- Updated `toTerraform`, `toHclTerraform`, and `toMetadata` to support the removed block when applicable.
- Added a new `remove(...)` method to define removed resources, including optional lifecycle and provisioner support.
- Introduced a new `RemovedBlockLocalExecProvisioner` interface, as JSII does not support the `Omit` type against `LocalExecProvisioner`, to make the API clearer and reduce developer confusion.
- Added compatibility checks to prevent calling `remove()` in combination with `importFrom` or any `moveXX()` methods, since these operations are mutually exclusive in Terraform’s configuration model.
- Added versions checks for the `remove` functionality (`1.7) and the `provisioner` block (`1.9`)
- Added HCL rendering support via `renderRemoved`

#### Rationale
- Keeping a dedicated `_removed` field aligns with the patterns used for import and move, making the codebase more consistent and easier to reason about.
- Using a distinct provisioner type for `removed` blocks improves the developer experience by avoiding misuse of the `when` property and surfacing clearer validation (as opposed to a value check and throw).
- Enforcing mutual exclusivity via runtime checks ensures predictable behaviour and avoids subtle bugs where `toTerraform()` would silently ignore conflicting operations.

### Checklist

### DIFF
--- a/packages/cdktf/lib/terraform-resource.ts
+++ b/packages/cdktf/lib/terraform-resource.ts
@@ -18,6 +18,7 @@ import { ITerraformIterator } from "./terraform-iterator";
 import { Precondition, Postcondition } from "./terraform-conditions";
 import { TerraformCount } from "./terraform-count";
 import {
+  RemovedBlockLocalExecProvisioner,
   SSHProvisionerConnection,
   WinrmProvisionerConnection,
 } from "./terraform-provisioner";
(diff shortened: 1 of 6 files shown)


packages/cdktf/lib/terraform-resource.ts
```diff
@@ -393,6 +446,12 @@ export class TerraformResource
               [this.terraformResourceType]: [this.friendlyUniqueId],
             }
           : undefined,
+      // TODO: its was unclear to me if how removed should interact with `toMetadata`
```
xphir, line 449: Looking to get some clarification here how `removed` should work with `toMetadata()`

PR #3888, opened 2025-05-28
feat!: update minimum required Node.js version to 20.19

mutahhir:
### Description

Since Nodejs 18 is no longer LTS, we're bumping the minimum version required for CDKTF to Node 20

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1802,7 +1802,7 @@ const bucket = new s3.S3Bucket(this, "bucket");
 const firstRuleStage = Fn.lookup(
   bucket.lifecycleRule("0").tags,
   "stage",
-  "no-stage"
+  "no-stage",
 );
 const firstRuleTags = bucket.lifecycleRule("0").tags;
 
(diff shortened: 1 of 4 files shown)


xiehan:
@DanielMSchmidt Among other things, there's no Node 22 runner for GitHub Actions yet, so anyone trying to use this in a GHA context (without Docker) is going to have issues if we require Node 22 as minimum.

I agree with using Node 20 here, but my personal take has always been to use the lowest value that's considered LTS which is `20.9.0` in this case, otherwise some folks are still going to run into compatibility issues.

mutahhir:
> I agree with using Node 20 here, but my personal take has always been to use the lowest value that's considered LTS which is `20.9.0` in this case, otherwise some folks are still going to run into compatibility issues.

I didn't realize that there were specific minor versions that were LTS. Where did you get this information? I just went for the latest minor version available, but that's useful to know

mutahhir:
I think I got it. That's the release they marked it as LTS on their blog.

xiehan:
@mutahhir It's also included in https://nodejs.org/download/release/index.json and what https://github.com/cdktf/cdktf-repository-manager/blob/main/scripts/check-node-versions.js is designed to pull out

xiehan (approved):

